PR #214, opened 2023-05-17
!wip refactor(queue-manager): move pending swap to rango-types

RanGojo:
We need to remove wallet-type RF-586 and then finalize and merge this pr.


RanGojo:
fixed in https://github.com/rango-exchange/rango-client/pull/468

PR #327, opened 2023-08-15
feat(widget-v2): redesign text-field

mikasackermn:


widget/ui/src/components/TextField/TextField.styles.ts
```diff
@@ -0,0 +1,141 @@
...
+    },
+    suffix: {
+      true: { marginRight: '$10' },
+    },
+  },
+  backgroundColor: 'transparent',
+  '-webkit-appearance': 'none',
+  margin: 0,
+  '&:disabled': {
+    cursor: 'not-allowed',
```
Ikari-Shinji-re, line 123: I think it will be better if we remove the hover style in this case.

widget/ui/src/components/TextField/TextField.types.ts
```diff
@@ -0,0 +1,15 @@
+export type PropTypes = {
+  label?: string;
+  prefix?: React.ReactNode;
+  suffix?: React.ReactNode;
+  size?: 'small' | 'large';
+  variant?: 'contained' | 'outlined' | 'ghost';
+  fullWidth?: boolean;
+  type?: 'dark' | 'light';
```
Ikari-Shinji-re, line 8: The type property overrides the input type attribute . Please rename or remove this property.

Ikari-Shinji-re (approved):

PR #334, opened 2023-08-20
feat(widget-v2): redesign radio and radioGroup

nikaaru:


examples/wallets-demo/src/App.tsx
```diff
@@ -29,15 +30,17 @@ export function App() {
       }
       setLoading(false);
     };
-    getAllBlockchains();
+    getAllBlockchains().catch((error) => {
```
yeager-eren, line 33: use `void getAllBlockchains()` to solve the eslint.

nikaaru: Done

PR #339, opened 2023-08-26
feat(widget-v2) refinement home page 

samobasquiat:


widget/embedded/src/pages/Home.tsx
```diff
@@ -161,53 +217,85 @@ export function Home() {
...
+                }
+              }}
+            />
+            <SwithFromAndToButton />
+          </FromContainer>
+          <SwapInput
+            label="To"
+            chain={{
+              displayName: toChain?.displayName || '',
+              image: toChain?.logo || '',
```
nikaaru, line 253: In my opinion, it is better to have a default chain and token.
Because first, when the page is loaded, the image source tag is empty and it is displayed without an image.

widget/embedded/src/pages/Home.tsx
```diff
@@ -161,53 +217,85 @@ export function Home() {
...
+              }}
+              disabled={loadingMetaStatus === 'failed'}
+              loading={loadingMetaStatus === 'loading'}
+              onSelectMaxBalance={() => {
+                if (tokenBalance !== '0') {
+                  setInputAmount(tokenBalanceReal.split(',').join(''));
+                }
+              }}
+            />
+            <SwithFromAndToButton />
```
nikaaru, line 247: SwitchFromAndToButton is correct

widget/embedded/src/pages/Home.tsx
```diff
@@ -47,6 +71,8 @@ const Footer = styled('div', {
 });
 
 const balancePercision = 8;
```
nikaaru, line 73: balancePrecision is correct.  and it's better to use uppercase.

widget/embedded/src/pages/Home.tsx
samobasquiat: It should be handled in the config, allowing users to set their preferred default chain and token. However, I addressed the situation like this: if the meta is loaded but no token is selected, an image placeholder appears. The handling of its loading status is being implemented in another [PR](https://github.com/rango-exchange/rango-client/pull/342).

nikaaru (approved):

PR #345, opened 2023-08-29
feat(wallets): make wallet provider blockchains prop optional

RanGojo:


wallets/react/src/provider.tsx
```diff
@@ -1,21 +1,23 @@
+import type { ProviderContext, ProviderProps } from './types';
+import type { WalletType } from '@rango-dev/wallets-shared';
+
+import { allSupportedBlockchains } from 'rango-chains';
```
yeager-eren, line 4: Dependency hasn't been added to `package.json`

wallets/react/src/provider.tsx
```diff
@@ -181,14 +192,13 @@ function Provider(props: ProviderProps) {
 
   useEffect(() => {
     const shouldTryAutoConnect =
-      props.allBlockChains &&
-      props.allBlockChains.length &&
-      props.autoConnect &&
-      !autoConnectInitiated.current;
+      // props.allBlockChains &&
```
yeager-eren, line 195: No commented code.

wallets/shared/src/helpers.ts
```diff
@@ -1,22 +1,59 @@
...
-  Networks,
   Connect,
-  Wallet,
+  EvmNetworksChainInfo,
   InstallObjects,
+  Network,
+  Wallet,
 } from './rango';
+import type {
+  StdBlockchainInfo,
```
yeager-eren, line 10: This comment should be written on `rango-types`, but as we talked I will put it here.

Is there any standard for what we need here? or it's only our convention and how we structure the code? 
if there is no actual standard, I suggest to don't use `Std` prefix since `std` can be misleading for people.
`BlockchainInfo` would works for what we do here.

yeager-eren (commented):
I've put my comments on your approach but I'm against to make the `wallets-react` dependent on `rango-chains`.
First, If we are adding `rango-chains` to `react`, it should be added to `core` since it can be used separately as well.
Second and the main one, our `wallets` already support for passing the list of blockchains so the library user should pass the data and use `rango-chains` if the user want. (For our use case it should be on `embedded`)

PR #347, opened 2023-08-30
feat(widget-v2): loading for bridges and exchanges

mikasackermn:


widget/embedded/src/components/LoadingLiquiditySourceList/LoadingLiquiditySourceList.styles.ts
```diff
@@ -0,0 +1,9 @@
+import { styled } from '@rango-dev/ui';
+
+export const List = styled('ul', {
+  padding: 0,
+  margin: 0,
+  listStyle: 'none',
+  height: 325,
```
nikaaru, line 7: The loading height is half of the container height

widget/embedded/src/components/LoadingLiquiditySourceList/LoadingLiquiditySourceList.styles.ts
```diff
@@ -0,0 +1,9 @@
+import { styled } from '@rango-dev/ui';
+
+export const List = styled('ul', {
```
nikaaru, line 3: It is better that the width of the loading is a little far from the scroll

nikaaru (approved):

PR #349, opened 2023-08-30
feat(wallets): add evm to phantom

nikaaru:


RanGojo (changes requested):
When user used a Solana only wallet in phantom, he faces with this error: 
`Unsupported Account.  App is trying to connect to an EVM network that this account doesn't support.`
And he could not proceed anymore. Please check if we could have a solution to support this case as well.

wallets/provider-phantom/src/helpers.ts
```diff
@@ -1,11 +1,46 @@
...
+
+  const { solana, ethereum } = phantom;
+
+  const instances: Provider = new Map();
+
+  if (ethereum && ethereum.isPhantom) {
+    instances.set(Networks.ETHEREUM, ethereum);
+  }
+
+  if (solana && phantom.isPhantom) {
```
RanGojo, line 22: Shouldn't it be `solana.isPhantom`?

PR #351, opened 2023-09-03
feat(queue-manager): add delete action to queue manager

mikasackermn:


samobasquiat (approved):

queue-manager/core/src/manager.ts
```diff
@@ -96,15 +102,16 @@ class Manager {
   private events: Events;
   private persistor: Persistor;
   private context: ManagerContext;
-  private isPaused: boolean = false;
```
yeager-eren, line 99: Why you removed the type? it should be typed like other properties (it only has default value.)

queue-manager/core/src/manager.ts
```diff
@@ -145,7 +155,12 @@ class Manager {
 
     this.context = options.context || {};
     this.persistor = new Persistor();
-    this.sync();
+
+    this.sync()
+      .then(() => console.log('Synced successfully'))
```
yeager-eren, line 160: Why we need to write something in console?

queue-manager/core/src/manager.ts
```diff
@@ -227,50 +242,59 @@ class Manager {
    * @returns An instance of `Queue` with wrapped event handlers from Manager.
    *
    */
+  // eslint-disable-next-line destructuring/in-methods-params
```
yeager-eren, line 245: you can fix this by not destructing in params.

queue-manager/core/src/manager.ts
```diff
@@ -558,14 +595,14 @@ class Manager {
    * Usually we call this method after a change detected in the state of manager.
    *
    */
-  private handleUpdate(queue_id: QueueID) {
+  private async handleUpdate(queue_id: QueueID) {
```
yeager-eren, line 598: Why you need to change the behaviour of this method on the whole file? the idea was triggering `persistor` to `persist` and update data like a background job. You changed the whole logic, it will wait for persistor to sync data with storage then runs the remaining stuff. This is not the purpose of this implementation.
Do you have any requirements to change the behaviour?

queue-manager/react/src/provider.tsx
```diff
@@ -95,6 +102,7 @@ function Provider(props: PropsWithChildren<PropTypes>) {
   }, [props.isPaused]);
 
   return (
+    // eslint-disable-next-line react/jsx-no-constructed-context-values
```
yeager-eren, line 105: Please solve the issue instead of disabling the rule.

queue-manager/core/src/manager.ts
mikasackermn: ESlint automatically removes this type (when saving the file).

queue-manager/core/src/manager.ts
```diff
@@ -227,13 +238,9 @@ class Manager {
...
    *
    */
-  private createQueue({
-    queue_id,
-    queue_name,
-  }: {
-    queue_id: QueueID;
-    queue_name: QueueName;
-  }) {
+  private createQueue(props: { queue_id: QueueID; queue_name: QueueName }) {
```
yeager-eren, line 241:
```suggestion
  private createQueue(info: { queue_id: QueueID; queue_name: QueueName }) {
```

queue-manager/core/src/manager.ts
```diff
@@ -460,6 +476,11 @@ class Manager {
     return this.queues;
   }
 
+  public async deleteQueue(queue_id: QueueID) {
```
yeager-eren, line 479:
```suggestion
  public deleteQueue(queue_id: QueueID) {
```
Please make a normal function

queue-manager/core/src/persistor.ts
```diff
@@ -22,7 +21,7 @@ class Persistor {
   db: Promise<IDBPDatabase<Database>>;
   constructor() {
     this.db = import('idb')
-      .then((idb) => {
+      .then(async (idb) => {
```
yeager-eren, line 24: Why `async`?

mikasackermn: This is one of the items that are automatically placed after saving the file

RanGojo:
@samobasquiat please double check.

samobasquiat (approved):

PR #352, opened 2023-09-03
feat(widget-v2): update confirm swap and confirm wallets components

Ikari-Shinji-re:


widget/embedded/src/pages/ConfirmSwapPage.tsx
```diff
@@ -1,265 +1,372 @@
...
+  const { customDestinationEnabled, multiWallets, supportedWallets, config } =
+    props;
   const navigate = useNavigate();
-  const { navigateBackFrom } = useNavigateBack();
+  const [showWallets, setShowWallets] = useState(false);
+  const [dbErrorMessage, setDbErrorMessage] = useState<string>('');
+  const [showSlippageWarning, setShowSlippageWarning] = useState(false);
   const bestRoute = useBestRouteStore.use.bestRoute();
+  const [walletsConfirmed, setWalletsConfirmed] = useState(false);
+  const inputAmount = useBestRouteStore.use.inputAmount();
```
yeager-eren, line 97: Can you explain why we are call `useBestRouteStore.use.whatever()` separately? 
Could we use this approach:
```
const store = useBestRouteStore();
```
and whenever we need a value, we will call it in-place:
```
store.inputAmount()
```

I mean we can call the store and keep the store/state in a variable then in the rest of our program accessing the value directly and remove these:
```
 const outputAmount = useBestRouteStore.use.outputAmount();
  const inputUsdValue = useBestRouteStore.use.inputUsdValue();
  const outputUsdValue = useBestRouteStore.use.outputUsdValue();
  const setInputAmount = useBestRouteStore.use.setInputAmount();
  const selectedWallets = useBestRouteStore.use.selectedWallets();
  const showWalletsOnInit = !useBestRouteStore.use.walletsConfirmed();
```

widget/embedded/src/hooks/useWalletList.ts
```diff
@@ -0,0 +1,97 @@
...
+  }, []);
+
+  useEffect(() => {
+    walletsRef.current = wallets;
+  }, [wallets]);
+
+  return {
+    wallets: sortedWallets,
+    error,
+    onClickWallet,
```
yeager-eren, line 95: I suggest `handleClickWallet` instead, because it's an handler actually.

widget/embedded/src/hooks/useWalletList.ts
```diff
@@ -0,0 +1,97 @@
...
+
+interface Params {
+  supportedWallets: WidgetConfig['wallets'];
+  multiWallets: boolean;
+  config?: WidgetConfig;
+  chain?: string;
+  onBeforeConnect?: (walletType: string) => void;
+  onConnect?: (walletType: string) => void;
+}
+export function useWalletList(params: Params) {
```
yeager-eren, line 23: Please add a JSDOC comment and explain the functionality of this hook.

widget/embedded/src/hooks/useWalletList.ts
```diff
@@ -0,0 +1,97 @@
...
+  const onClickWallet = async (type: WalletType) => {
+    const wallet = state(type);
+    try {
+      if (error) {
+        setError('');
+      }
+      if (wallet.connected) {
+        await disconnect(type);
+      } else {
+        if (
```
yeager-eren, line 57: Can you explain this `if`?

yeager-eren (commented):
Please also rebase with the `alpha`.

PR #353, opened 2023-09-03
feat(widget-v2): animation for modal

samobasquiat:


widget/ui/src/components/Modal/Modal.styles.ts
```diff
@@ -2,15 +2,15 @@ import { styled } from '../../theme';
...
+  top: '1000px',
   left: '0',
   width: '100%',
   height: '100%',
   backgroundColor: 'rgba(0,0,0,.1)',
   zIndex: 10,
   borderRadius: '$sm',
   display: 'flex',
-
+  transition: 'top .8s ease-in-out',
```
nikaaru, line 13: I tested it on the connect wallets page, I'm not sure, but I think it's better if the animation is on the ModalContainer.
When the animation is on BackDrop, it looks like a shadow moves when it is closed.

nikaaru (approved):

PR #355, opened 2023-09-04
fix(widget-v2): base component problems

samobasquiat:


widget/embedded/src/pages/HistoryPage.tsx
```diff
@@ -78,34 +78,27 @@ export function HistoryPage() {
...
-                groupBy={groupSwapsByDate}
-              />
-            ) : (
-              <Alert type="info" title="Swap" />
-            )}
-          </SwapList>
+          <SwapsGroupContainer>
+            <SwapsGroup
+              list={searchedFor ? filteredList : list}
+              isSearched={!!searchedFor}
```
yeager-eren, line 94: I suggest don't consider a separate prop for this. if `list` prop is empty, we should show the `No results found`.

widget/ui/src/containers/SwapInput/SwapInput.tsx
```diff
@@ -1,27 +1,35 @@
...
   Typography,
 } from '../../components';
+import { useTextOverflowDetector } from '../../hooks';
 
 import { Container, InputAmount } from './SwapInput.styles';
 import { TokenSection } from './TokenSection';
 
 export function SwapInput(props: SwapInputProps) {
+  const tooltipContentRef = useRef<HTMLSpanElement | null>(null); // Specify the type
+  const isTruncated = useTextOverflowDetector<BaseProps['price']['usdValue']>(
```
yeager-eren, line 20: I'm not okay with using a js solution. I think we can fix it using CSS only.

samobasquiat: I used it to show tooltips for long text that we want to cut them. I removed the tooltip.

widget/ui/src/components/SwapContainer/SwapContainer.tsx
```diff
@@ -24,13 +24,11 @@ const MainContainer = styled('div', {
     '&::-webkit-scrollbar': { width: 2, height: '$8' },
     '&::-webkit-scrollbar-thumb': {
       borderRadius: '$sm',
-      backgroundColor: '$neutral300',
-    },
-    '&::-webkit-scrollbar-thumb:hover': {
-      backgroundColor: '$neutral500',
+      backgroundColor: '$secondary300',
```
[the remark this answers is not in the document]
samobasquiat: I don't think that we should update the body scrollbar. I didn't catch your comment. let's talk.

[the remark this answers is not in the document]
samobasquiat: Fixed.

PR #361, opened 2023-09-10
feat(widget-v2): token/chain selector problems

mikasackermn:


RanGojo (changes requested):
Please check the following items: @mikasackermn 
- [ ] link for token explorer is not working
- [ ] I think you only need change the sort of blockchains (move something to the first) when an item is selected from show more.
- [ ] blockchain tag background color doesn't match with design.
- [ ] please switch position of cosmos and osmosis in top filter.
- [ ] blockchain icon should be rounded.
- [ ] separator line should not have border-radius
- [ ] Padding between scroll bar items is not handled yet in search token page.
- [ ] when token name is very long, it's not displayed correctly. e.g. search: realtoken,

widget/embedded/src/components/AppRoutes.tsx
```diff
@@ -35,8 +35,15 @@ export function AppRoutes(props: PropTypes) {
       element: (
         <SelectSwapItemsPage
           type="from"
-          supportedChains={config?.from?.blockchains}
+          supportedBlockchains={config?.from?.blockchains}
           supportedTokens={config?.from?.tokens}
+          pinnedTokens={[
```
RanGojo, line 40: Please handle reading it fro config.

widget/embedded/src/components/AppRoutes.tsx
```diff
@@ -45,26 +52,33 @@ export function AppRoutes(props: PropTypes) {
       element: (
         <SelectSwapItemsPage
           type="to"
-          supportedChains={config?.to?.blockchains}
+          supportedBlockchains={config?.to?.blockchains}
           supportedTokens={config?.to?.tokens}
+          pinnedTokens={[
```
RanGojo, line 57: also here.

RanGojo:
It which cases we show the price like this? $
Screenshot from 2023-09-12 15-27-54

RanGojo:
Please consider these changes:

- [ ] use rounded icons for tokens
- [ ] remove explorer url from cosmos tokens for now
- [ ] use both (text+icon) for tokens explorer url link
- [ ] please change style / animation when a token doesn't have name. (move it to middle)

widget/embedded/src/types/config.ts
```diff
@@ -137,4 +137,5 @@ export type WidgetConfig = {
   language?: Language;
   theme?: WidgetTheme;
   externalWallets?: boolean;
+  pinnedTokens?: Asset[];
```
RanGojo, line 140: Please update the doc with new properties.

widget/embedded/src/utils/common.ts
```diff
@@ -29,3 +30,51 @@ export function containsText(text: string, searchText: string): boolean {
   // eslint-disable-next-line @typescript-eslint/no-magic-numbers
   return text.toLowerCase().indexOf(searchText.toLowerCase()) > -1;
 }
+
+export const colorShade = (col: string, amt: number) => {
```
RanGojo, line 34: Could we change it somehow to have a better background colors for dark theme when blockchain color is near to black?

mikasackermn: I did not find a solution for this because the color that is coming from the server is almost black

RanGojo:
Please check the infinite scroll bar problem.

Recording 2023-09-19 at 10 15 55

RanGojo:
- [ ] token display name is not vertically aligned to the middle.

Screenshot from 2023-09-19 10-51-15

RanGojo:
- [ ] Loading skeleton looks better now, but we need a bit change to keep it aligned with non-loading state. speacially in loading filter above.
- [ ] also each item height is not matched before/after load. 

 
Recording 2023-09-19 at 10 54 08

RanGojo:
Token icon position shouldn't change  on hover.

Recording 2023-09-19 at 11 11 11

RanGojo:
It seems blockchain selection sometimes doesn't work as expected.

Recording 2023-09-19 at 11 25 08 (1)

RanGojo (approved):

PR #369, opened 2023-09-26
feat(widget-v2): skeleton for wallets page

samobasquiat:


RanGojo (approved):

widget/embedded/src/components/WalletModal/WalletModal.types.ts
```diff
@@ -3,9 +3,9 @@ import type { WalletState } from '@rango-dev/ui';
 export interface PropTypes {
   open: boolean;
   onClose: () => void;
-  image: string;
-  state: WalletState;
-  error?: boolean;
+  image?: string;
+  state?: WalletState;
```
yeager-eren, line 7: It should alway has a state, why did you make it optional?

widget/ui/src/components/Wallet/Wallet.types.ts
```diff
@@ -55,11 +55,11 @@ interface WalletProps {
   onClick: (type: WalletType) => void;
   selected?: boolean;
   description?: string;
+  loadingStatus?: 'failed' | 'success' | 'loading';
```
yeager-eren, line 58: I think `status` is a better name here and also we can use `connected` instead of `sucess`.

samobasquiat: It's unrelated to the wallet state. it's just for when Meta has not been loaded yet. I've modified both its name and type.

widget/embedded/src/components/WalletModal/WalletModal.types.ts
samobasquiat: fixed.

PR #376, opened 2023-10-01
feat(playground-v2): implement general option for playground

samobasquiat:


widget/embedded/src/Widget.tsx
```diff
@@ -21,6 +21,9 @@ const MainContainer = styled('div', {
   display: 'flex',
   justifyContent: 'center',
   alignItems: 'center',
+  '*': {
```
yeager-eren, line 24: It works but it affects all the elements (even if for those that don't need) so it causes some performance issues.

Couldn't we apply only to `html` and `body`? 
Or if we want to reset elements and only set to inherit fonts to read from `html` or `body` ?
There are some solutions like https://meyerweb.com/eric/tools/css/reset/.

widget/embedded/src/hooks/useTheme.ts
```diff
@@ -1,87 +1,86 @@
...
+import {
+  Default_Primary_Radius,
+  Default_SECONDARY_Radius,
+} from '../utils/configs';
 
 export function useTheme({
   // colors,
   fontFamily = 'Roboto',
-  // borderRadius = 8,
+  borderRadius = Default_Primary_Radius,
```
yeager-eren, line 16: Please use all uppercase for naming.

yeager-eren: also for Default_SECONDARY_Radius,

widget/playground/src/components/Collapse/Collapse.tsx
```diff
@@ -0,0 +1,36 @@
+import type { PropTypes } from './Collapse.types';
```
yeager-eren, line 1: Couldn't we use Radix? Or don't we have anything in our design system?

samobasquiat:
- [ ] disable functional
- [ ] fix help icon
- [ ] general default open
- [ ] fix clickable trigger in general
- [ ] add hover on sidebar

widget/embedded/src/hooks/useTheme.ts
```diff
@@ -1,87 +1,86 @@
...
+  secondaryBorderRadius = Default_SECONDARY_Radius,
   mode = 'auto',
 }: WidgetTheme) {
   const theme = useSettingsStore.use.theme();
   const fetchMeta = useMetaStore.use.fetchMeta();
   const setTheme = useSettingsStore.use.setTheme();
-  // const light = themeColors?.light;
```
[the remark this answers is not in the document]
samobasquiat: I was not in the context of these comments but I removed them.

widget/embedded/src/hooks/useTheme.ts
samobasquiat: Fixed.

widget/playground/src/components/Collapse/Collapse.tsx
samobasquiat: It was supposed to be a simple one because of the deadline. however, I used Radix and transferred it to widget/UI.

widget/embedded/src/Widget.tsx
samobasquiat: In the past, we used to change the body of the whole page with add/remove class. I think it's not appropriate to change the body. I remove * and allow elements to inherit from their parents.

PR #380, opened 2023-10-04
feat(widget-v2): sync swap process with notifications

samobasquiat:


widget/embedded/src/components/SwapDetails/SwapDetails.tsx
```diff
@@ -90,12 +92,27 @@ export function SwapDetails(props: SwapDetailsProps) {
...
-    if (swap.status === 'success' || swap.status === 'failed') {
-      setShowCompletedModal(swap.status);
+    const existNotification = unreadNotifications.find(
+      (n) => n.requestId === swap.requestId
+    );
+    if (existNotification) {
+      const {
+        event: { type },
+      } = existNotification;
+      const isUnread =
```
RanGojo, line 110: Is there any reason that you check unread notification by this condition instead of reading from store?

widget/embedded/src/components/WidgetEvents.tsx
```diff
@@ -42,10 +42,16 @@ export function WidgetEvents() {
         toAccount && getWalletsDetails([toAccount]);
       }
       if (
-        event.type === StepEventType.TX_EXECUTION_BLOCKED &&
-        validBlockedStatuses.includes(event.status)
+        (event.type === StepEventType.TX_EXECUTION_BLOCKED &&
+          validBlockedStatuses.includes(event.status)) ||
+        event.type === StepEventType.FAILED
```
RanGojo, line 47: I didn't understand your intention about this line?
`event.type === StepEventType.FAILED
`

RanGojo (approved):

widget/embedded/src/components/SwapDetails/SwapDetails.tsx
samobasquiat: We don't need it. I removed it.

widget/embedded/src/components/WidgetEvents.tsx
samobasquiat: As I understand it, we don't trigger RouteEvent all the time, so I put it to make sure the notifier is called.

RanGojo: Please create another task to handle it.

PR #381, opened 2023-10-07
chore(widget-v2): refactor navigation

Ikari-Shinji-re:
TODO:

- [x] When user set a path other than `/*` our internal links should be relative or absolute? e.g. `/settings` or `settings`
- [x] In some paths (like `confirm-swap`), when user click on settings, back button should redirect the user to `confirm-swap` instead of home. (read from history, instead of `route`).

Potential solutions:

https://reactrouter.com/en/main/hooks/use-navigate#optionsrelative


RanGojo:
I've tested this PR again and it seems okay to me. But I think your css changes had some side effects, here is an example: (bullet style)

Screenshot from 2023-10-08 08-56-59

RanGojo (approved):

Ikari-Shinji-re:
To replicate the initial bug, modify the path for the widget component in the app package to a value other than ‘/*’, and then proceed with the swap flow until you reach the confirmSwapPage. Once you have confirmed the swap, the bug will manifest itself :
```ts
 <Routes>
        <Route path="/swap/*" element={<Widget config={configRef.current} />} />
      </Routes>
```

yeager-eren:
@Ikari-Shinji-re 
I tired to solve the basename issue using the URL spec. it's working properly.
I also wrote a more detailed documentation in Contributing.md.
Please check my last commit and let me know if there is any question.


URL Spec:
https://url.spec.whatwg.org/#urls

PR #385, opened 2023-10-08
fix(widget-v2): update playground export code ui

RanGojo:


samobasquiat:
I saw some UI problems:

- [ ]  The syntax highlighter, buttons, and input's radius should be fixed
- [ ]  Modals in Figma have fixed heights & widths and don't match
- [ ]  There should be a larger close icon.
- [ ] Vertically, the links are not centered
- [ ] In Figma, I saw an alert about a wrong API key

widget/ui/src/components/Modal/index.ts
```diff
@@ -1 +1,2 @@
 export { Modal } from './Modal';
+export { ModalHeader } from './Modal.styles';
```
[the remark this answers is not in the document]
RanGojo: Because modal title font size and close icon size are hard coded in Modal component, so i need to override that.

samobasquiat (approved):

widget/playground/src/components/ExportConfigModal/ExportConfigModal.tsx
```diff
@@ -0,0 +1,157 @@
...
+    initialConfig,
+    apiKey
+  );
+  const formatedConfig = formatConfig(filteredConfigForExport);
+
+  return (
+    <Modal
+      containerStyle={{
+        width: '1109px',
+        height: '865px',
```
yeager-eren, line 62: I suggest something like:

```
maxHeight: 865px;
height: 90%;
```
to make sure it works in smaller heights as well.

yeager-eren: Same thing for width:
maxWidth: '1109px',
width: 90%

widget/playground/src/helpers.ts
```diff
@@ -101,10 +101,12 @@ export function clearEmpties<T extends Record<string, any>>(obj: T): T {
 
 export function filterConfig(
   WidgetConfig: WidgetConfig,
-  initialConfig: WidgetConfig
+  initialConfig: WidgetConfig,
+  apiKey: string
```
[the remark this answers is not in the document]
RanGojo: Because the API key provided by dApp should not use in playground because the default API key cors setting on server side doesn't allow it.
So it is only for the export modal input and exported code not for using in playground (api calls).

widget/playground/src/components/ExportConfigModal/ExportConfigModal.tsx
```diff
@@ -0,0 +1,157 @@
...
+        padding: '$20',
+        fontFamily: 'Roboto',
```
yeager-eren, line 64: Shouldn't we read the font from `theme`?

widget/playground/src/components/ExportConfigModal/ExportConfigModal.tsx
```diff
@@ -0,0 +1,157 @@
...
+              <CloseIcon color="gray" size={22} />
+            </StyledIconButton>
+          </ModalFlex>
+        </ModalHeader>
+      }
+      open={open}
+      onClose={onClose}
+      title="Export Code"
+      anchor="center">
+      <>
```
yeager-eren, line 86: Unnecessary fragment

widget/playground/src/components/ExportConfigModal/ExportConfigModal.tsx
```diff
@@ -0,0 +1,157 @@
...
+      header={
+        <ModalHeader>
+          <Typography variant="headline" size="large">
+            Export Code
+          </Typography>
+          <ModalFlex>
+            <StyledIconButton
+              onClick={onClose}
+              variant="ghost"
+              style={{ width: '48px', height: '48px' }}>
```
yeager-eren, line 76: Why we need inline style here? `StyledIconButton` already has the samething.

widget/playground/src/components/ExportConfigModal/ExportConfigModal.tsx
```diff
@@ -0,0 +1,157 @@
...
+          </ModalFlex>
+        </ModalHeader>
+      }
+      open={open}
+      onClose={onClose}
+      title="Export Code"
+      anchor="center">
+      <>
+        <Divider size={32} />
+        <Flex>
```
yeager-eren, line 88: I suggest rename it to `Head`

widget/playground/src/components/ExportConfigModal/CodeBlock.tsx
```diff
@@ -0,0 +1,41 @@
...
+  const { language, theme, children } = props;
+
+  const [, handleCopy] = useCopyToClipboard(RESET_INTERVAL);
+
+  return (
+    <CodeBlockContainer>
+      <Tooltip content="Copy to clipboard" side="top">
+        <CopyCodeBlockButton
+          type="primary"
+          onClick={handleCopy.bind(null, children)}>
```
yeager-eren, line 28: Use arrow function please.

widget/playground/src/helpers.ts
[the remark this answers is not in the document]
yeager-eren: Good point. How about handle it in `widget/playground/src/App.tsx`? 
We have something like this:
```
const config = useConfigStore.use.config();
<Route path="/*" element={<Widget config={config} />} />
```
we can change it to:
```
<Route path="/*" element={<Widget config={{...config, apiKey: OUR_PLAYGROUND_API_KEY}} />} />
```

In this way it's more explicit what we are doing here and it won't be missed in future refactors.

widget/playground/src/components/ExportConfigModal/ExportConfigModal.tsx
yeager-eren: Good point. But what I know, we are going to apply the widget theme to our playground as well. if we are going to override some parts of that (e.g. font), I suggest to define the value inside `widget/playground/src/constants.ts` and use it here, it lets us to easily change them in future.

RanGojo: Yes, but it seems we didn't go this way in previous PRs.  But I moved it to constant as you suggested

PR #395, opened 2023-10-10
feat(widget-v2): support multi languages

nikaaru:


mikasackermn (approved):

widget/ui/src/components/I18nManager/I18nManager.tsx
```diff
@@ -16,7 +16,7 @@ import { messages as esMessages } from '../../../../../translations/es';
 import { messages as frMessages } from '../../../../../translations/fr';
 import { messages as jpMessages } from '../../../../../translations/jp';
 
-const DEFAULT_LANGUAGE = 'en';
+export const DEFAULT_LANGUAGE = 'en';
```
yeager-eren, line 19: I think moving this const to `widget/embedded/src/constants/languages.ts` is more appropirate place.

widget/ui/src/components/Flags/Flag.tsx
```diff
@@ -0,0 +1,25 @@
+import type { FlagPropTypes } from './Flags.types';
+import type { PropsWithChildren } from 'react';
+
+import React from 'react';
+
+export function Flag(props: PropsWithChildren<FlagPropTypes>) {
```
yeager-eren, line 6: I'm not sure making a separate component for `svg` here is necessary. I suggest to use `<svg ...>` right inside the component itself. 
I suggest this structure for this component:
```

/Flags
    index.ts -> only `export { default as English }`. you already did it.
    Flags.constants.ts -> `export const DEFAULT_SIZE = "1em"`
    Flags.types.ts -> keep your current file, using same interface for flags is a good idea .
    English.tsx -> equivalent to your `EnFlag.tsx` 
    ... -> Other languages

```

Note: Please use full language name for your files.

widget/embedded/src/store/settings.ts
```diff
@@ -7,14 +8,15 @@ import { removeDuplicateFrom } from '../utils/common';
 import { useMetaStore } from './meta';
 import createSelectors from './selectors';
 
-type Theme = 'auto' | 'dark' | 'light';
+export type Theme = 'auto' | 'dark' | 'light';
```
yeager-eren, line 11: I suggest rename the type to `ThemeMode` here.

widget/embedded/src/hooks/useLanguage.ts
```diff
@@ -0,0 +1,15 @@
+import { DEFAULT_LANGUAGE, type Language } from '@rango-dev/ui';
```
yeager-eren, line 1: You can extend the functionality of this hook to be single source of truth for our widget.
I suggest this interface for output of this hook:

```
interface UseLanguage{
  languages: LanguageItem[];
  defaultLanguage: string;
  activeLanguage: string;
  changeLanguage: (language: string) => void;  
}
```

Note 1: 
re-export your `LANGUAGES` (`widget/embedded/src/constants/languages.ts`) as `languages` from this hook. then update `widget/embedded/src/pages/LanguagePage.tsx` to use `languages` from this hook instead of importang from constants directly.

Note 2:
move your useEffect from this file to `widget/embedded/src/Widget.tsx` and use `changeLanguage`.

yeager-eren: The whole idea is this hook reading, updating and returning information about language.

widget/embedded/src/hooks/useLanguage.ts
```diff
@@ -0,0 +1,27 @@
...
+  languages: LanguageItem[];
+  defaultLanguage: Language;
+  activeLanguage: Language;
+  changeLanguage: (language: Language) => void;
+}
+
+export function useLanguage(): UseLanguage {
+  const language = useSettingsStore.use.language();
+  const setLanguage = useSettingsStore.use.setLanguage();
+  const languages = [...LANGUAGES];
```
yeager-eren, line 18:
```suggestion
  const languages = LANGUAGES;
```

widget/embedded/src/hooks/useLanguage.ts
```diff
@@ -0,0 +1,27 @@
...
+  const defaultLanguage = 'en';
```
yeager-eren, line 19: read from `widget/embedded/src/constants/languages.ts`.

widget/ui/src/components/I18nManager/I18nManager.tsx
nikaaru: Done

widget/embedded/src/hooks/useLanguage.ts
nikaaru: Done

widget/embedded/src/hooks/useLanguage.ts
nikaaru: Done

yeager-eren:
I've tried to fix the problem, there is two points that needs to be considered.
- When language has been changed, we should re-render the whole app to change texts.
- We shouldn't define i18n statically at the file scope because they won't be updated if language has changed. Please check my commit on `errorMessaages`.


Please check the commits if it solves the problem, we can go ahead and try to merge the PR. Before merge please check [this commit](https://github.com/rango-exchange/rango-client/pull/395/commits/d371f16159860285f949f5118a06b418f858f112) and make sure it won't break anything and also if it's not necessary we can revert the commit.

nikaaru:
> I've tried to fix the problem, there is two points that needs to be considered.
> 
> * When language has been changed, we should re-render the whole app to change texts.
> * We shouldn't define i18n statically at the file scope because they won't be updated if language has changed. Please check my commit on `errorMessaages`.
> 
> Please check the commits if it solves the problem, we can go ahead and try to merge the PR. Before merge please check [this commit](https://github.com/rango-exchange/rango-client/pull/395/commits/d371f16159860285f949f5118a06b418f858f112) and make sure it won't break anything and also if it's not necessary we can revert the commit.
> 
> **Note:** I only changed `errorMessages` as an example, please make sure there is no more i18n text inside our javascript that are defining at the file level. (if they are inside functions, there no need to change anything)

I checked all the ts files and fixed them

PR #399, opened 2023-10-14
fix(widget-v2): fix retry logic in failed swaps

Ikari-Shinji-re:
If a swap consists of multiple steps and fails at any one step, the retry mechanism should continue from the failed step and proceed with the remaining steps

### DIFF
--- a/widget/embedded/src/store/bestRoute.ts
+++ b/widget/embedded/src/store/bestRoute.ts
@@ -9,9 +9,11 @@ import { subscribeWithSelector } from 'zustand/middleware';
 
 import { ZERO } from '../constants/numbers';
 import { isPositiveNumber } from '../utils/numbers';
-import { getBestRouteToTokenUsdPrice } from '../utils/routing';
+import {
+  createRetryRoute,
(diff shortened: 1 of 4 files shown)


widget/embedded/src/store/bestRoute.ts
```diff
@@ -188,63 +193,75 @@ export const useBestRouteStore = createSelectors(
...
+        let fromToken: Asset | null = null;
+        let inputAmount = '';
         const firstStep = pendingSwap.steps[0];
         const lastStep = pendingSwap.steps[pendingSwap.steps.length - 1];
-        const fromBlockchain =
-          blockchains.find(
-            (blockchain) => blockchain.name === firstStep.fromBlockchain
-          ) || null;
-        const toBlockchain =
+        const lastSuccessfulStep = pendingSwap.steps
```
RanGojo, line 202: We have this logic in `getLastSuccessfulStep` function in rango-preset. We could probably use that here and avoid duplicate codes.

RanGojo (approved):

PR #403, opened 2023-10-16
chore(widget-v2): sync home page logic with confirm swap page

Ikari-Shinji-re:


widget/embedded/src/utils/routing.ts
```diff
@@ -1,5 +1,5 @@
 /* eslint-disable @typescript-eslint/no-magic-numbers */
```
yeager-eren, line 1: It maybe takes sometime but to be closer to what terms industry uses, we should consider to use `quote` instead of `route`. Please rename whatever you can to `quote`.

widget/embedded/src/components/Route/index.ts
```diff
@@ -0,0 +1 @@
+export { Route } from './Route';
```
yeager-eren, line 1: I think it's technically a `container` so we can move it to `/src/containers/Route`.

widget/embedded/src/components/Route/Route.tsx
```diff
@@ -0,0 +1,266 @@
...
+import { RouteWarningsAndErrors } from '../RouteWarningsAndErrors';
+
+import {
+  BestRouteContainer,
+  FooterAlert,
+  FooterStepAlarm,
+} from './Route.styles';
+
+export function Route(props: PropTypes) {
+  const { inputAmount } = useBestRouteStore();
```
yeager-eren, line 44: You've two `useBestRouteStore()`, please merge them into one.

widget/embedded/src/components/Route/Route.tsx
```diff
@@ -0,0 +1,266 @@
...
+  return (
+    <div>
+      {loading && (
+        <BestRouteContainer>
+          <BestRouteSkeleton type={type} expanded={expanded} />
+        </BestRouteContainer>
+      )}
+      {showBestRoute ? (
+        <BestRouteContainer>
+          <BestRoute
```
yeager-eren, line 204: I think we should improve the `BestRoute` props a bit here. Instead of passing the values as separate props (e.g. `totalFee`, `totalTime` and `steps`) which all of them are reading from `bestRoute` object, we can pass the whole object in a single prop (like `swap`/`route`) and do the calculation inside the `BestRoute` component itself (move `getBestRouteSteps` into the component).

widget/embedded/src/components/Route/Route.tsx
```diff
@@ -0,0 +1,266 @@
...
+              />
+            </FooterStepAlarm>
+          ) : undefined,
+      };
+    });
+  };
+
+  const showBestRoute = bestRoute && bestRoute.result && !loading;
+  return (
+    <div>
```
yeager-eren, line 196: Can you use `Fraagment` to keep the DOM cleaner?

widget/embedded/src/components/RouteWarningsAndErrors/RouteWarningsAndErrors.helpers.tsx
```diff
@@ -0,0 +1,78 @@
+import type {
```
yeager-eren, line 1: Helpers should be pure JS and not JSX. Please make the function into smaller pieces to make this happens if that works.

widget/embedded/src/utils/swap.ts
```diff
@@ -530,21 +561,12 @@ export function getWalletsForNewSwap(selectedWallets: Wallet[]) {
...
-  outputUsdValue: string | number | null
-) {
-  if (!inputUsdValue || !outputUsdValue) {
-    return null;
-  }
-  return new BigNumber(outputUsdValue)
-    .div(new BigNumber(inputUsdValue))
-    .minus(1)
-    .multipliedBy(100);
+export function getPercentageChange(input: BigNumber, output: BigNumber) {
```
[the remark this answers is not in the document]
yeager-eren: It can be BigNUmber or string in your recent change, I think it should be one of them, string.

widget/embedded/src/hooks/useConfirmSwap.ts
```diff
@@ -1,53 +1,49 @@
 /* eslint-disable @typescript-eslint/no-magic-numbers */
 import type {
```
yeager-eren, line 2: Maybe it's not in the scope of this task but could you use `usePrepareBlockchainList` structure for this hook as well? I mean move helpers to a separate file. then we can think about how we can write tests.

PR #404, opened 2023-10-16
hotfix(widget-v2): fix bugs in app store

Ikari-Shinji-re:


RanGojo:
- [ ] when setting default blockchain/token, the URL will be overrided after refresh. (but for the default amount, we have another behavior)

RanGojo (approved):

PR #415, opened 2023-10-23
fix(playground-v2): blockchains name & misspelling in playground

samobasquiat:


widget/ui/src/components/TextField/TextField.styles.ts
```diff
@@ -114,7 +114,7 @@ export const Input = styled('input', {
     color: '$neutral900',
   },
   '&:focus-within': {
-    outline: 'none',
+    outline: 'none !important',
```
yeager-eren, line 117: I'm not agree with using `!important`, we should solve it in a right way. the bug here is actually a side effect of `widget/embedded/src/Widget.tsx:32`. I'm not sure why we added that line. please check it with @Ikari-Shinji-re .

samobasquiat: fixed.

PR #416, opened 2023-10-23
feat(widget-v2): keep user history for selected chains

yeager-eren:
Remaining stuff:

- [x] List is flashy after select
- [x] Last selected item should be the first in history.


yeager-eren:
@Ikari-Shinji-re In my las commit these are fixed:
- Clicking on blockchains are flashy sometimes
- Main list (e.g. ETH) is hidden when preferred blockchains is more than list limit (e.g. 10)
- If only one item has left in list, we should show the `More` button.

Ikari-Shinji-re (approved):

Ikari-Shinji-re (approved):

PR #417, opened 2023-10-24
feat(playground-v2): test our previous theme preset for widget

mikasackermn:


mojtabast:
I tried to do some clean up and separating concerns in my last commit, changes include:

- Moving out fetching meta from `useTheme`
- Handling customized theme using a separate utility. default theme and customized theme should be more clear right now.
- A unique id will be generated for each customized theme to make sure we re applying correct theme tokens.

yeager-eren:
I tried to do some clean up and separating concerns in my last commit, changes include:

- Moving out fetching meta from `useTheme`
- Handling customized theme using a separate utility. default theme and customized theme should be more clear right now.
- A unique id will be generated for each customized theme to make sure we re applying correct theme tokens.

PR #424, opened 2023-11-01
feat: persist playground config

mikasackermn:


widget/playground/src/store/config.ts
```diff
@@ -228,8 +228,16 @@ export const useConfigStore = createSelectors(
...
-          externalWallets: state.config.externalWallets,
+          ...state,
+          config: {
+            ...Object.fromEntries(
+              Object.entries(state.config).filter(
+                ([key]) => !['externalWallets'].includes(key)
+              )
+            ),
+          },
         }),
```
Ikari-Shinji-re, line 241: It's possible to make it simpler :
```ts
 partialize: (state) => {
          const { externalWallets, ...config } = state.config;
          return { ...state, config };
        },
```

mikasackermn: fixed

Ikari-Shinji-re (approved):

RanGojo (approved):

PR #426, opened 2023-11-04
fix(widget-v2): initial state with props in app store and fix bug of passing liquidity sources via config

Ikari-Shinji-re:
- [ ] Initialize app store with widget config
- [ ] passing supported liquidity sources to quote request properly
- [ ] show liquidity sources based on widget config


RanGojo:
merged in 
https://github.com/rango-exchange/rango-client/pull/442

PR #427, opened 2023-11-04
fix(widget): fix safe accounts issue

RanGojo:
- [x] When safe provider is not connected, we shouldn't show it in list of selectable wallet for evm blockchains. (in confirm wallet step during the swap process in widget)
- [x] When safe provider is connected to an evm network, we should only populate account for that specific network not all evm chains.

### DIFF
--- a/wallets/core/src/wallet.ts
+++ b/wallets/core/src/wallet.ts
@@ -12,8 +12,13 @@ export type EventHandler = (
   event: Events,
   value: any,
   coreState: State,
-  supportedChains: BlockchainMeta[]
+  info: EventInfo
 ) => void;
+
+export type EventInfo = {
+  supportedBlockchains: BlockchainMeta[];
+  isContractWallet: boolean;
+};
 
 export interface State {
   connected: boolean;
@@ -34,13 +39,16 @@ class Wallet<InstanceType = any> {
   private actions: WalletActions;
   private state: State;
   private options: Options;
-  private meta: BlockchainMeta[];
+  private info: EventInfo;
 
   constructor(options: Options, actions: WalletActions) {
     this.actions = actions;
     this.options = options;
     this.provider = null;
-    this.meta = [];
+    this.info = {
+      supportedBlockchains: [],
+      isContractWallet: false,
+    };
     this.state = {
       connected: false,
       connecting: false,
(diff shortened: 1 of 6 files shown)


wallets/core/src/wallet.ts
```diff
@@ -30,17 +35,19 @@ export interface Options {
 }
 
 class Wallet<InstanceType = any> {
+  public provider: InstanceType | null;
   private actions: WalletActions;
   private state: State;
   private options: Options;
-  private meta: BlockchainMeta[];
-  public provider: InstanceType | null;
+  private supportedBlockchains: BlockchainMeta[];
```
yeager-eren, line 42: I suggest make the contextual info (like `supportedBlockchains` and `isContractWallet`) into one object will let us to do the deprecations (if needed) or add/remove features more easily. In addition to that, the code will be easier to maintain I believe.

So what I think is something like this:
`private info: {
supportedBlockchains: BlockchainMeta[];
isContractWallet: boolean;
}`

wallets/core/src/wallet.ts
```diff
@@ -308,8 +300,9 @@ class Wallet<InstanceType = any> {
     }
   }
 
-  setMeta(value: BlockchainMeta[]) {
-    this.meta = value;
+  setMeta(value: BlockchainMeta[], isContractWallet: boolean) {
```
yeager-eren, line 303: According to https://github.com/rango-exchange/rango-client/pull/427/files#r1382562861 , I suggest change this a little bit:
`setInfo(info: Info)`

yeager-eren: That would be great to let the user to pass partial values. something like `updateState(states: Partial<State>)` that we already have. It lets the user to only passing the needed values (`supportedBlockchains` or `isContractWallet`, and not both of them at the sametime).

wallets/core/src/wallet.ts
```diff
@@ -12,9 +12,14 @@ export type EventHandler = (
   event: Events,
   value: any,
   coreState: State,
-  supportedChains: BlockchainMeta[]
+  meta: EventMeta
```
yeager-eren, line 15: To be more consistent with https://github.com/rango-exchange/rango-client/pull/427/files#r1382562861 , If we rename it the meta to `info` is better.

`info: EventInfo`

RanGojo: My suggestion is to change `EventInfo` to something `WalletInfo` or something like that. because this model is also used in other sections (e.g. wallets.ts private field) which are not related to Events.

samobasquiat (approved):

PR #423, opened 2023-10-31
fix(wallet):  fix experimental chain issue on cosmos wallets

nikaaru:

PR #429, opened 2023-11-05
fix(wallet): add reject modal for experimental suggest chain

nikaaru:


RanGojo:
We need to merge https://github.com/rango-exchange/rango-client/pull/423 first

PR #430, opened 2023-11-06
fix(widget-v2): fix emitting failed event in swap execution

Ikari-Shinji-re:


nikaaru (approved):

samobasquiat:
I think we can remove `event.type === StepEventType.FAILED` in `widget/embedded/src/components/WidgetEvents.tsx`.

samobasquiat (approved):

PR #437, opened 2023-11-13
fix(widget-v2): fix HMR for widget and playground

Ikari-Shinji-re:


yeager-eren:
According to our offline conversation, I pushed a commit that I think can solve our problems. It runs ts check in a separate layer and we can fix the HMR itself using `source` field as well.

My commit is just an example that needs to be tested and polished.

Checklist:
- Rename the scripts inside `package.json` to more appropriate name.
- Install `npm-run-all` in workspace so we can use it without `npx`.
- I didn't check `npm-run-all` it uses `yarn` (because we are using it) or not, if not we should switch the package to something else.
- rename all `"module": "xxxx"` to `"source"`.
- Add a `ts-check` command to all packages like what I did for `wallets/react`.

PR #438, opened 2023-11-13
feat: add disabledFeatures to widget config

samobasquiat:


Ikari-Shinji-re:
If we disable a feature, the persisted values should not affect the application, and default values should be used instead . (like: disabledLiquiditySources, theme and language)

Ikari-Shinji-re:
I suggest moving the settings state to the app store as a slice and initializing it with widget config, similar to the data slice. Additionally, if the feature is disabled, reset the persisted field. Also, consider setting ‘auto’ mode as the default value for the theme if it is disabled.

widget/embedded/src/types/config.ts
```diff
@@ -21,6 +21,14 @@ export type WidgetColors = {
...
 };
 
+export enum DisabledFeatures {
+  Notifications = 'notifications',
+  Theme = 'theme',
+  Liquidity_Sources = 'liquiditySources',
+  Connect_Wallet = 'connectWallet',
+  Language = 'language',
+}
+
```
Ikari-Shinji-re, line 31: it is a common practice to use all caps for TypeScript enums.

widget/embedded/src/components/HeaderButtons/HomeButtons.tsx
```diff
@@ -26,6 +28,12 @@ export function HomeButtons(props: HomeButtonsPropTypes) {
     onClickNotifications,
   } = props;
 
+  const { config } = useAppStore();
+
+  const disabledNotifications = !!config?.disabledFeatures?.includes(
+    DisabledFeatures.Notifications
+  );
```
Ikari-Shinji-re, line 35: A helper function that verifies disabled features could be useful.
(please check other files)

widget/embedded/src/types/config.ts
samobasquiat: fixed.

widget/embedded/src/components/HeaderButtons/HomeButtons.tsx
samobasquiat: fixed.

Ikari-Shinji-re (approved):

PR #440, opened 2023-11-14
feat(widget-v2): support experimental features

samobasquiat:
# Summary
In this task, we were supposed to support experimental features. These Items are set in the dapp & storage and in the widget, we read apiUrl or experimental best route from config. 

For testing, we can manually set config values for experimentalFeatures & apiUrl to change the request body of best route & apiUrl in our fetches.

# Checklist:

### DIFF
--- a/widget/embedded/src/QueueManager.tsx
+++ b/widget/embedded/src/QueueManager.tsx
@@ -30,6 +30,7 @@ function QueueManager(props: PropsWithChildren<{ apiKey?: string }>) {
   const swapQueueDef = useMemo(() => {
     return makeQueueDefinition({
       API_KEY: props.apiKey || getConfig('API_KEY'),
+      BASE_URL: getConfig('BASE_URL'),
     });
   }, [props.apiKey]);
 
--- a/widget/embedded/src/constants/index.ts
+++ b/widget/embedded/src/constants/index.ts
@@ -1,1 +1,2 @@
 export const RANGO_PUBLIC_API_KEY = 'c6381a79-2817-4602-83bf-6a641a409e32';
+export const DEFAULT_BASE_URL = 'https://api.rango.exchange';
--- a/widget/embedded/src/services/httpService.ts
+++ b/widget/embedded/src/services/httpService.ts
@@ -8,6 +8,6 @@ export const httpService = () => {
   if (rango) {
     return rango;
   }
-  rango = new RangoClient(getConfig('API_KEY'));
+  rango = new RangoClient(getConfig('API_KEY'), getConfig('BASE_URL'));
   return rango;
 };
--- a/widget/embedded/src/utils/configs.ts
+++ b/widget/embedded/src/utils/configs.ts
@@ -2,6 +2,7 @@ import { RANGO_PUBLIC_API_KEY } from '../constants';
 
 export interface Configs {
   API_KEY: string;
+  BASE_URL?: string;
 }
 
 let configs: Configs = {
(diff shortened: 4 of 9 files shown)


widget/embedded/src/constants/index.ts
```diff
@@ -1 +1,5 @@
+import { getExperimentalBaseUrl } from '../utils/quote';
+
 export const RANGO_PUBLIC_API_KEY = 'c6381a79-2817-4602-83bf-6a641a409e32';
+export const DEFAULT_BASE_URL = 'https://api.rango.exchange';
+export const RANGO_BASE_URL = getExperimentalBaseUrl() || DEFAULT_BASE_URL;
```
mikasackermn, line 5: It seems that changing the base URL has no effect. please check.

widget/embedded/src/Widget.tsx
```diff
@@ -8,6 +8,7 @@ import React, { useContext, useEffect, useMemo, useState } from 'react';
 import { AppRouter } from './components/AppRouter';
 import { AppRoutes } from './components/AppRoutes';
 import { WidgetEvents } from './components/WidgetEvents';
+import { RANGO_BASE_URL } from './constants';
 import { globalFont } from './globalStyles';
 import { useLanguage } from './hooks/useLanguage';
 import { useTheme } from './hooks/useTheme';
```
mikasackermn, line 14: needReloadTheApp is one of the props related to ExperimentalFeatures located in the Dapp and if true, the app should be reloaded if the URL changes, and if false, no reload is required (not implemented in the widget).

widget/embedded/src/QueueManager.tsx
```diff
@@ -43,7 +44,7 @@ function QueueManager(props: PropsWithChildren) {
     })),
   };
 
-  const switchNetwork = (wallet: WalletType, network: Network) => {
+  const switchNetwork = async (wallet: WalletType, network: Network) => {
```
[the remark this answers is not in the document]
samobasquiat: It was added by Eslint. We have to.

widget/embedded/src/Widget.tsx
samobasquiat: yeah but I think It should be handled in dapp not here. In dapp we reload the page and that's why the widget is reloaded as well.

widget/embedded/src/constants/index.ts
samobasquiat: It was my mistake. fixed.

mikasackermn (approved):

yeager-eren:
I would suggest to consider it as a config and not a separate thing and implement it using local storage directly.

The final config can be like this:

```
config = {
   apiUrl?: string 
   experimental?: {
        routing?: boolean
   }
}
```

widget/embedded/src/utils/configs.ts
```diff
@@ -2,14 +2,15 @@ import { RANGO_PUBLIC_API_KEY } from '../constants';
```
yeager-eren, line 2: I think this file changes should be reverted.

widget/embedded/src/services/httpService.ts
```diff
@@ -8,6 +8,6 @@ export const httpService = () => {
   if (rango) {
     return rango;
   }
-  rango = new RangoClient(getConfig('API_KEY'));
+  rango = new RangoClient(getConfig('API_KEY'), getConfig('BASE_URL'));
```
yeager-eren, line 11: I think this one should be reverted as well.

widget/embedded/src/utils/configs.ts
yeager-eren: My bad. resolve it.

PR #449, opened 2023-11-20
feat: add balances value to useWidget

samobasquiat:
# Summary

In some steps we are getting the balance of the user, in this task we need to propagate the received data to dApp (like external wallets). It helps us to reduce network requests and make the widget a single source of truth for wallet balance.

Fixes # (issue)

these values are used in this PR https://github.com/rango-exchange/app/pull/12 and with linking it can be tested.

# Checklist:

### DIFF
--- a/widget/embedded/src/containers/WidgetInfo/WidgetInfo.tsx
+++ b/widget/embedded/src/containers/WidgetInfo/WidgetInfo.tsx
@@ -1,9 +1,11 @@
 import type { WidgetInfoContextInterface } from './WidgetInfo.types';
 
 import { useManager } from '@rango-dev/queue-manager-react';
-import React, { createContext, useContext, useMemo } from 'react';
+import React, { createContext, useContext } from 'react';
 
+import { useWalletsStore } from '../../store/wallets';
 import { getPendingSwaps } from '../../utils/queue';
+import { calculateWalletUsdValue } from '../../utils/wallets';
 
 export const WidgetInfoContext = createContext<
   WidgetInfoContextInterface | undefined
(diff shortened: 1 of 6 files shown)


widget/embedded/src/containers/WidgetInfo/WidgetInfo.tsx
```diff
@@ -12,11 +14,26 @@ export const WidgetInfoContext = createContext<
 export function WidgetInfo(props: React.PropsWithChildren) {
   const { manager } = useManager();
   const swaps = getPendingSwaps(manager);
+  const walletsDetails = useWalletsStore.use.connectedWallets();
+  const isWalletLoading = useWalletsStore.use.loading();
+  const totalBalance = calculateWalletUsdValue(walletsDetails);
+  const refetchBalance = useWalletsStore.use.getWalletsDetails();
   const value = useMemo(() => {
```
yeager-eren, line 21: Why it would be inside an Memo, when it will be changed always?

widget/embedded/src/containers/WidgetInfo/WidgetInfo.tsx
```diff
@@ -12,11 +14,26 @@ export const WidgetInfoContext = createContext<
...
     return {
       swaps,
+      walletsDetails,
```
yeager-eren, line 24: I would suggest to expose all of them as a single property. e.g.:

```
swaps,
wallets: {
isLoading,
details,
totalBalance,
refetch
}

```

widget/embedded/src/containers/WidgetInfo/WidgetInfo.tsx
samobasquiat: It is an Eslint error. I temporarily disable the rule.

widget/embedded/src/containers/WidgetInfo/WidgetInfo.tsx
samobasquiat: changed.

PR #452, opened 2023-11-22
fix: fix widget configs in memory router

Ikari-Shinji-re:
# Summary

config values for tokens and blockchains doesn't affect the widget in memory router.

test widget with and without react router.
test widget with react router and search parameters for token and blockchain.

# Checklist:

- [x] widget config for blockchains and tokens should work in memory router.
- [x] search parameters should override the widget state if config values exists. (using react router)


yeager-eren:
useLayoutEffect isn't a good solution, we merge it as workarround for now, in next week we will fix it properly on UpdateUrl refactoring.

PR #455, opened 2023-11-26
chore: export types & logics from widget

samobasquiat:
# Summary

There are some dependencies to @rango-dev/... (helpers, types, ...) in dapp and should be removed, so we should export them from embedded.

Fixes # (issue)
There is another PR here to test: 
https://github.com/rango-exchange/app/pull/14

# Checklist:


widget/embedded/src/index.ts
```diff
@@ -84,9 +106,14 @@ export {
   useWallets,
   useWidgetEvents,
   customizedThemeTokens,
+  readAccountAddress,
   MainEvents,
+  Networks,
```
yeager-eren, line 111: I'm a bit concerned about exporting so many things from `embedded`. Some of them is Rango specific I believe and will be deprecated soon like `Networks` and `WalletTypes`.

For more explicitly separate them, I suggest to export them using a separate `export`.

```
export {
  useWallets,
  useWidgetEvents,
  customizedThemeTokens,
  ...
 }
 
 // Internal exports for Rango
 export {
 Networks,
 WalletTypes,
 ...
 }
```

 We can do same thing for `exprot type { ...}` as well.
 
 The final goal keeping these internal things separated somehow, only put them on main api (export) when we are sure they should be there.

wallets/react/src/types.ts
```diff
@@ -20,8 +21,6 @@ export type ConnectResult = {
   provider: any;
 };
 
-export type Providers = { [type in WalletType]?: any };
```
yeager-eren, line 23: It's part of the `wallets/react` API and should be kept here. Please revert this change.

samobasquiat: fixed.

PR #458, opened 2023-12-02
fix: fix bug of not publishing language files in npm

mikasackermn:
# Summary

Language files inside /translations/* should be published with embedded as well, if not, it will not work properly.
I tested with our own code, there was no problem, I just transferred I18NManager to the widget (because it is used in the widget).

Serve Playground and change the language

# Checklist:


samobasquiat:
I encountered an error within the `widget/embedded/src/components/I18nManager/I18nManager.tsx `file:
`File 'rango-client/translations/en.ts' is not under 'rootDir' . 'rootDir' is expected to contain all source files.ts`. I think this is because of the values of rootDir in tsconfig. for example, in `rango-dev/ui` we considered it in tsconfig.

widget/ui/tsconfig.json
```diff
@@ -7,7 +7,7 @@
     "lib": ["dom", "esnext"],
     // match output dir to input dir. e.g. dist/index instead of dist/src/index
     "baseUrl": ".",
-    "rootDirs": ["./src", "../../../translations"],
+    "rootDirs": ["./src"],
```
samobasquiat, line 10: Do we still need an array? I mean `rootDirs` insetad `rootDir`

samobasquiat (approved):

RanGojo:
Not needed anymore.

PR #461, opened 2023-12-03
feat: export meta from useWidget

samobasquiat:
# Summary

In the dapp we call meta as well & delaying in fetching meta caueses some issues. We decieded to get meta from useWidget hook.

Fixes # (issue)

there is a WIP pr in dapp side:
https://github.com/rango-exchange/app/pull/17
with linking it can be tested.

# Checklist:

### DIFF
--- a/widget/embedded/src/containers/Wallets/Wallets.tsx
+++ b/widget/embedded/src/containers/Wallets/Wallets.tsx
@@ -28,7 +28,7 @@ export const WidgetContext = createContext<WidgetContextInterface>({
 });
 
 function Main(props: PropsWithChildren<PropTypes>) {
-  const { updateConfig, updateSettings } = useAppStore();
+  const { updateConfig, updateSettings, fetch: fetchMeta } = useAppStore();
   const blockchains = useAppStore().blockchains();
   const tokens = useAppStore().tokens();
   const walletOptions: ProvidersOptions = {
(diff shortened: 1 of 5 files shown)


widget/embedded/src/containers/WidgetInfo/WidgetInfo.helpers.ts
```diff
@@ -11,9 +13,14 @@ import { getPendingSwaps } from '../../utils/queue';
 
 export class WidgetHistory {
   private manager: Manager | undefined;
+  private retrySwap: (pendingSwap: PendingSwap, meta: Meta) => void;
 
-  constructor(manager: Manager | undefined) {
+  constructor(
+    manager: Manager | undefined,
+    retrySwap: (pendingSwap: PendingSwap, meta: Meta) => void
```
yeager-eren, line 20: I think make the second parameters to a more general object would be better.
`actions: { retrySwap: .... }`. 
lets us to add/remove more easily in futute.

widget/embedded/src/containers/Wallets/Wallets.tsx
```diff
@@ -40,6 +41,10 @@ function Main(props: PropsWithChildren<PropTypes>) {
   const onConnectWalletHandler = useRef<OnConnectHandler>();
   useSyncStoresWithConfig();
 
+  useEffect(() => {
+    void fetchMeta().catch();
```
yeager-eren, line 45:
```suggestion
    void fetchMeta().catch(console/log);
```
I know you just moved the code, but completely ignoring the error will make frustrations in future. we should see the error in console at least.

widget/embedded/src/containers/WidgetInfo/WidgetInfo.helpers.ts
samobasquiat: fixed.

widget/embedded/src/containers/Wallets/Wallets.tsx
samobasquiat: fixed.

PR #467, opened 2023-12-09
fix: fix wallet button state in swap details page

Ikari-Shinji-re:
# Summary

On Swap page, when we are showing the button for connecting the required wallet, if you click on connect wallet twice (when the wallet is on connecting state), it will crash.
We should disable the button, when we are connecting the wallet.

Fixes # (issue)

pass wallet-core state to wallet button

On Swap page, when we are showing the button for connecting the required wallet, if you click on connect wallet twice (when the wallet is on connecting state), It should not prompt any errors.

### DIFF
--- a/widget/embedded/src/components/SwapDetailsModal/SwapDetailsModal.WalletState.tsx
+++ b/widget/embedded/src/components/SwapDetailsModal/SwapDetailsModal.WalletState.tsx
@@ -1,30 +1,36 @@
 import type { WalletStateContentProps } from './SwapDetailsModal.types';
 
-import { MessageBox, Wallet, WalletState } from '@rango-dev/ui';
+import { MessageBox, Wallet } from '@rango-dev/ui';
 import { useWallets } from '@rango-dev/wallets-react';
 import React from 'react';
 
 import { getContainer } from '../../utils/common';
+import { mapStatusToWalletState } from '../../utils/wallets';
 
 import { WalletContainer } from './SwapDetailsModal.styles';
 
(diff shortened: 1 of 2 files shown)


widget/embedded/src/components/SwapDetailsModal/SwapDetailsModal.WalletState.tsx
```diff
@@ -1,30 +1,33 @@
...
 export const WalletStateContent = (props: WalletStateContentProps) => {
   const { type, title, currentStepWallet, message, showWalletButton } = props;
-  const { connect, getWalletInfo, state: walletState } = useWallets();
+  const { connect, getWalletInfo, state } = useWallets();
   const walletType = currentStepWallet?.walletType;
-  const isConnected = walletType && walletState(walletType).connected;
-  const state = isConnected ? WalletState.CONNECTED : WalletState.DISCONNECTED;
+  const walletState = !!walletType && mapStatusToWalletState(state(walletType));
+  const walletInfo = !!walletType && getWalletInfo(walletType);
+  const shouldShowWallet = showWalletButton && walletState && walletInfo;
```
samobasquiat, line 18: I think `shouldShowWallet` should be a `boolean`, but it's currently inferred as `false | WalletInfo`. Maybe is better to remove !! from `walletType` and place them before `walletState` && `walletInfo`. What do you think?

samobasquiat (approved):

RanGojo (approved):

PR #474, opened 2023-12-18
chore: use virtualized list for swap history items

Ikari-Shinji-re:
# Summary

If the user has an extensive list of swaps, such as 100 items, there are performance concerns with rendering the list. As of now, react-window does not provide support for grouping. To address this limitation, we plan to incorporate the react-virtuoso library, which offers support for grouping, and utilize it in instances where we are already employing react-window.

Fixes # (issue)

- [x] The scroll functionality should be restricted to the swaps list, rather than allowing it for the entire page.

# Checklist:


widget/embedded/src/components/SwapsGroup/SwapsGroup.tsx
```diff
@@ -57,68 +102,68 @@ export function SwapsGroup(props: PropTypes) {
...
+    <GroupedVirtuoso
+      endReached={() => {
+        if (loadedItems.current < list.length) {
+          loadMore();
+        }
+      }}
+      groupCounts={currentGroupCounts}
+      groupContent={(index) => {
+        return (
+          <Group hasMargin={index !== 0}>
```
samobasquiat, line 114: Can we use a divider instead of setting margin?

widget/embedded/src/components/SwapsGroup/SwapsGroup.tsx
```diff
@@ -2,15 +2,60 @@ import type { PropTypes } from './SwapsGroup.types';
...
+    setCurrentGroupCounts(
+      calculateGroupsSoFar(groupCounts, loadedItems.current)
+    );
+  }, [isLoading]);
+
+  useEffect(() => {
+    if (!isLoading) {
+      loadMore();
+    }
+  }, [isLoading]);
```
samobasquiat, line 58: Shouldn't we add loadMore in dependency?

widget/embedded/src/components/SwapsGroup/SwapsGroup.tsx
```diff
@@ -2,15 +2,60 @@ import type { PropTypes } from './SwapsGroup.types';
...
+        count -= group;
+        index++;
+      } while (count > 0 && index <= totalGroups.length);
+      return groups;
+    },
+    []
+  );
+
+  const loadMore = useCallback(() => {
+    if (list.length - loadedItems.current > ITEMS_PER_PAGE) {
```
samobasquiat, line 42: Can we have a better approach such as Math.min or something else instead of these conditions?

widget/embedded/package.json
```diff
@@ -37,6 +37,7 @@
     "rango-types": "^0.1.57",
     "react-i18next": "^12.2.0",
     "react-router-dom": "^6.8.0",
+    "react-virtuoso": "^4.6.2",
```
samobasquiat, line 40: As I noticed we have lots of libs for handling virtualized lists including: `react-virtuoso`, `react-window`, `react-window-infinite-loader`, `react-virtualized-auto-sizer`. If the new lib is better (bundle size, performance, easy to use, etc) can we make our virtualized list component to use it?

yeager-eren:
I'm agree with @samobasquiat on we shouldn't use different solutions for samething.

There are two approach to this:

1. Use our current libs:  (react-window, react-window-infinite-loader, react-virtualized-auto-sizer)
Downside: We lose sticky header, but can implement grouping. (As Shinji said)
You can ask @RanGojo if we need sticky or not. if we don't, we can go with this.

2. Use `virtuoso`:
Downside: bundle size.

It seems it supports for tree shaking, so the final bundle will be smaller hopefully. 
If we go with this approach, we should remove the other 3 libs since they already have 35kb size and by adding  `virtuoso` it goes to ~80kb in total.

yeager-eren (changes requested):
I'm agree with @samobasquiat on we shouldn't use different solutions for samething.

There are two approach to this:

Use our current libs: (react-window, react-window-infinite-loader, react-virtualized-auto-sizer)
Downside: We lose sticky header, but can implement grouping. (As Shinji said)
You can ask @RanGojo if we need sticky or not. if we don't, we can go with this.

Use virtuoso:
Downside: bundle size.

It seems it supports for tree shaking, so the final bundle will be smaller hopefully.
If we go with this approach, we should remove the other 3 libs since they already have 35kb size and by adding virtuoso it goes to ~80kb in total.

widget/embedded/src/components/SwapsGroup/SwapsGroup.tsx
Ikari-Shinji-re: The divider component does not yield the desired outcome.

widget/embedded/src/components/SwapsGroup/SwapsGroup.tsx
Ikari-Shinji-re: No, the "loadMore" function is independent and is only instantiated once.

samobasquiat:
Please update the PR description.

samobasquiat:
Screenshot1
Screenshot2
Screenshot3
Screenshot4

samobasquiat (approved):

widget/embedded/src/components/SwapsGroup/SwapsGroup.styles.ts
```diff
@@ -2,17 +2,34 @@ import { css, darkTheme, styled } from '@rango-dev/ui';
...
     $$color: '$colors$neutral600',
     [`.${darkTheme} &`]: {
       $$color: '$colors$neutral700',
     },
     color: '$$color',
   },
+  variants: {
+    hasMargin: {
+      true: {
+        marginTop: '$15',
```
yeager-eren, line 29: I don't know where this `15` is coming from.

widget/embedded/src/components/SwapsGroup/SwapsGroup.tsx
```diff
@@ -11,11 +16,52 @@ import {
...
+const ITEMS_PER_PAGE = 10;
 
 export function SwapsGroup(props: PropTypes) {
   const { list, onSwapClick, groupBy, isLoading } = props;
-  const groups = groupBy ? groupBy(list) : [{ title: 'History', swaps: list }];
+  const [currentGroupCounts, setCurrentGroupCounts] = useState<number[]>([]);
+  const loadedItems = useRef(0);
+  const { swaps, groupCounts, groups } = groupBy(list);
+
+  const calculateGroupsSoFar = useCallback(
```
yeager-eren, line 35: I don't understand what is this. Could you please describe?

yeager-eren:
I found another bug, I'm not sure what I did is correct but for making my list longer I used this trick:

```
 <SwapsGroup
              list={[...filteredList, ...filteredList, ...filteredList]}
```

But what I see is something like this which is not my complete list and it's only a part of the list.

https://github.com/rango-exchange/rango-client/assets/120930547/11bd88bf-2d8a-4ef7-ae67-ee5fea372beb

widget/embedded/src/components/SwapsGroup/SwapsGroup.tsx
Ikari-Shinji-re: I accidentally included "isLoading" in the loadMore dependency array. It does not depend on isLoading.

widget/embedded/src/components/SwapsGroup/SwapsGroup.tsx
Ikari-Shinji-re: The calculateGroupsSoFar Slices the total groups into the groups which contain the items so far. For example, if you have [10, 10, 10, 10] groups in total, slicing them to 23 will result in [10, 10, 3] :
https://virtuoso.dev/grouped-with-load-on-demand/

widget/embedded/src/components/HistoryGroupedList/HistoryGroupedList.tsx
```diff
@@ -0,0 +1,188 @@
...
+        title: i18n.t('Today'),
+        swaps,
+      },
+      {
+        title: i18n.t('This month'),
+        swaps,
+      },
+    ];
+    return (
+      <>
```
yeager-eren, line 80: Do we need fragment here?

widget/embedded/src/components/HistoryGroupedList/HistoryGroupedList.tsx
```diff
@@ -0,0 +1,188 @@
...
+                  className={groupStyles()}>
+                  {group.title}
+                </Typography>
+              </Time>
+              <Divider size={4} />
+              <SwapList>
+                {group.swaps.map((_, index) => {
+                  const key = index + group.title;
+                  return (
+                    <React.Fragment key={key}>
```
yeager-eren, line 97: Wouldn't it better if you add key to SwapListItem directly and remove the fragment?

widget/embedded/src/components/HistoryGroupedList/HistoryGroupedList.tsx
```diff
@@ -0,0 +1,188 @@
...
+                      <SwapListItem isLoading={true} />
```
yeager-eren, line 98: The use case here is kind of confusing I believe. Wouldn't better to extract the `widget/ui/src/components/SwapListItem/SwapListItem.tsx:22` to a separate component like `SwapListItem.Loading.tsx` and use in SwapListItem and here in your HistoryGrouppedList as well?

Ikari-Shinji-re: This is about the changes we made earlier. We took two approaches: either making a separate skeleton or loading component for the specific component, or dealing with the loading state inside the component itself. I think we should stick to one of these methods for all our components.

yeager-eren (approved):
Thanks.

PR #477, opened 2023-12-20
fix: walletDetails request is fired twice

samobasquiat:
# Summary

This pull request resolves the issue where the 'accountChanged' event fires twice in wallets like Metamask when we change the accounts, causing duplicate server requests for walletDetails.

To reproduce (for example in https://widget.rango.exchange/) by switching MetaMask accounts, observe multiple server requests for balance and account information.
[Screencast](https://github.com/rango-exchange/rango-client/assets/123297941/ce9dc26f-cd9b-4b8d-82f9-b91d6582b611)

Fixes # (issue)

I added cleanup functions in both wallets/core and providers. I tested all wallets. For those without "off" events, I returned `void` during cleanup. The testing procedure ensured that changing the account triggers a single server request.

# Checklist:

### DIFF
--- a/wallets/core/src/types.ts
+++ b/wallets/core/src/types.ts
@@ -66,6 +66,8 @@ export type Disconnect = (options: {
   destroyInstance: () => void;
 }) => Promise<void>;
 
+type CleanupSubscribe = () => void;
+
 export type Subscribe = (options: {
   instance: any;
   state: WalletState;
@@ -74,7 +76,7 @@ export type Subscribe = (options: {
   updateAccounts: (accounts: string[], chainId?: string) => void;
   connect: (network?: Network) => void;
   disconnect: () => void;
-}) => void;
+}) => CleanupSubscribe | void;
 
 export type SwitchNetwork = (options: {
   instance: any;
--- a/wallets/provider-station/src/index.ts
+++ b/wallets/provider-station/src/index.ts
@@ -14,6 +14,8 @@ import signer from './signer';
 
 const WALLET = WalletTypes.STATION;
 const STATION_WALLET_ID = 'station';
+const INTERVAL_TIMEOUT = 3_000;
+const MAX_TRY_COUNT = 3;
 
 export const config = {
   type: WALLET,
@@ -30,11 +32,11 @@ async function waitInterval(instance: any) {
       } else {
         count++;
       }
-      if (count > 3) {
+      if (count > MAX_TRY_COUNT) {
         resolve(state);
         clearInterval(interval);
       }
-    }, 3000);
+    }, INTERVAL_TIMEOUT);
   });
 }
 
--- a/wallets/provider-xdefi/src/index.ts
+++ b/wallets/provider-xdefi/src/index.ts
@@ -59,8 +59,7 @@ export const subscribe: Subscribe = ({
   updateChainId,
   connect,
 }) => {
-  const eth = chooseInstance(instance, meta, Networks.ETHEREUM);
-  eth?.on('chainChanged', (chainId: string) => {
+  const handleChainChanged = (chainId: string) => {
     const network = getBlockChainNameFromId(chainId, meta) || Networks.Unknown;
     /*
      *TODO:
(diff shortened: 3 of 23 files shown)


wallets/core/src/types.ts
```diff
@@ -74,7 +74,7 @@ export type Subscribe = (options: {
   updateAccounts: (accounts: string[], chainId?: string) => void;
   connect: (network?: Network) => void;
   disconnect: () => void;
-}) => void;
+}) => () => void;
```
yeager-eren, line 77: if a `subscribe` function doesn't implement a cleanup, it will return `undefined` by default in JS. So I think the type here will be something like this:
`
type CleanupSubscribe = () => void;

// ... 
=> CleanupSubscribe | undefined
`

yeager-eren:
Screenshot
(it should send `address` as well)

It seems it's initiating a bad request on `eager` mode. 
Reproduce: You can connect your metamask, then reload the page.

yeager-eren:
Please also add a reproduce case to your PR description, because I can not reproduce the bug always, it seems it's happening randomly for me.

wallets/core/src/types.ts
samobasquiat: fixed.

samobasquiat:
> Screenshot (it should send `address` as well)
> It seems it's initiating a bad request on `eager` mode. Reproduce: You can connect your metamask, then reload the page.

fixed.

samobasquiat:
> Please also add a reproduce case to your PR description, because I can not reproduce the bug always, it seems it's happening randomly for me.

added.

wallets/provider-station/src/index.ts
```diff
@@ -75,6 +77,9 @@ export const subscribe: Subscribe = ({
       }
     },
   });
+  return () => {
```
yeager-eren, line 80: Do we need these? if nothing returns, it means undefined which a defined behavior in our cleanup mechanism.

yeager-eren: There are some other places same as here, this comment apply on them as well.

wallets/provider-xdefi/src/index.ts
```diff
@@ -77,7 +76,11 @@ export const subscribe: Subscribe = ({
      */
     updateChainId(chainId);
     connect(network);
-  });
+  };
+  const eth = chooseInstance(instance, meta, Networks.ETHEREUM);
+  eth?.on('chainChanged', handleChainChanged);
+
+  return () => eth?.off('chainChanged', handleChainChanged);
```
yeager-eren, line 83:
```suggestion
  return () => { eth?.off('chainChanged', handleChainChanged); }
```

I don't know what `off` returns but for making sure it won't break anything in future, let's stick to the `void` which is defined as return type for cleanup.

widget/embedded/src/containers/Wallets/Wallets.tsx
```diff
@@ -99,6 +107,16 @@ function Main(props: PropsWithChildren<PropTypes>) {
       props.onUpdateState(type, event, value, state, meta);
     }
   };
+  const reloadWallet = useCallback(
+    () => connectWallet(accounts, tokens),
+    [accounts]
```
yeager-eren, line 112: You need to add `tokens` to callback dependency as well.

yeager-eren: nit: I think using `useCallback` here doesn't add any value, because all the codes inside the function are references so you won't save anything here.

yeager-eren (changes requested):

wallets/provider-station/src/index.ts
samobasquiat: fixed.

widget/embedded/src/containers/Wallets/Wallets.tsx
samobasquiat: fixed.

wallets/provider-xdefi/src/index.ts
samobasquiat: fixed.

yeager-eren (approved):

PR #484, opened 2023-12-26
feat: add transition for tabs

samobasquiat:
# Summary

Implemented a smooth transition for tab-switching for both the LeftNavigation Tabs indicator and the background of the themes tab. 

Switching tabs in LeftNavigation and themes should have a smooth transition.

# Checklist:

### DIFF
--- a/widget/playground/src/components/ExportConfigModal/ExportConfigModal.styles.tsx
+++ b/widget/playground/src/components/ExportConfigModal/ExportConfigModal.styles.tsx
@@ -19,6 +19,7 @@ export const ButtonsContainer = styled('div', {
   justifyItems: 'center',
   alignItems: 'center',
   backgroundColor: '$neutral300',
+  position: 'relative',
 });
 
 export const ModalFlex = styled('div', {
@@ -47,15 +48,27 @@ export const StyledButton = styled(Button, {
     },
     type: {
       secondary: {
+        zIndex: 10,
+        transition: 'color 0.8s linear',
         color: '$background',
-        backgroundColor: '$secondary500',
+        backgroundColor: 'transparent',
         '&:hover, &:focus': {
           color: '$background',
-          backgroundColor: '$secondary500',
+          backgroundColor: 'transparent',
         },
       },
     },
   },
+});
+
+export const BackdropTab = styled('div', {
+  width: '362px',
+  height: '$40',
+  backgroundColor: '$secondary500',
+  position: 'absolute',
+  borderRadius: '$xm',
+  inset: 0,
+  transition: 'transform 0.5s ease-in-out',
 });
 
 export const LinkContainer = styled('div', {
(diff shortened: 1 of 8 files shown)


RanGojo:
Please also add same animation when switching between tabs of export modal, it seems it was missing from task description.

RanGojo (commented):
Good job, please also add same animation when switching between tabs of export modal, it seems it was missing from task description.

widget/playground/src/components/ExportConfigModal/ExportConfigModal.styles.tsx
```diff
@@ -47,17 +48,28 @@ export const StyledButton = styled(Button, {
...
 });
 
+export const BackdropTab = styled('div', {
+  width: '362px',
+  height: '$40',
+  backgroundColor: '$secondary500',
+  position: 'absolute',
+  borderRadius: '$xm',
+  inset: 0,
+  transition: 'transform 0.5s cubic-bezier(0, 0, 0.86, 1.2)',
```
nikaaru, line 70: NIT: Considering that it is out of the container, I think a number smaller than one is better for the fourth parameter.

nikaaru (approved):

widget/playground/src/containers/StyleLayout/StyleLayout.styles.ts
```diff
@@ -34,17 +34,43 @@ export const Tabs = styled('div', {
...
+      true: {
+        $$color: '$colors$background',
+        [`.${darkTheme} &`]: {
+          color: '$colors$foreground',
+        },
+        color: '$$color',
+      },
+      false: {
+        '&:hover': {
+          backgroundColor: '$info100',
```
nikaaru, line 55: please double-check this hover in dark theme.

RanGojo (approved):

widget/playground/src/containers/StyleLayout/StyleLayout.styles.ts
samobasquiat: fixed.
(Note: I think in Figma, we do not have enough information about the dark playground theme in different sections.)

widget/playground/src/components/ExportConfigModal/ExportConfigModal.styles.tsx
samobasquiat: changed.

samobasquiat:
> Good job, please also add same animation when switching between tabs of export modal, it seems it was missing from task description.

added.
@RanGojo

RanGojo:
NIT: when I'm switching between tabs, before background color change, text color changes to white. It's not very important but I was wondering if we could fix it with some small changes in animation timings. @samobasquiat

RanGojo (approved):

samobasquiat:
> NIT: when I'm switching between tabs, before background color change, text color changes to white. It's not very important but I was wondering if we could fix it with some small changes in animation timings. @samobasquiat

fixed. @RanGojo

RanGojo (approved):

PR #486, opened 2023-12-27
fix: reset the quote when the source token matches the destination token

Ikari-Shinji-re:
# Summary

On the home page, if we have a quote and set the source and destination tokens to be the same, it disrupts the state, and the user can navigate to the "confirm-swap" page. Additionally, on the home page, when hovering over the "notifications" button and clicking on it, as long as the notification popover is open, the notification tooltip is visible.

Fixes # (issue)

Resetting the quote state when tokens are identical.
Removing the popover component from within the notification tooltip.

# Checklist:

### DIFF
--- a/widget/embedded/src/components/HeaderButtons/HomeButtons.tsx
+++ b/widget/embedded/src/components/HeaderButtons/HomeButtons.tsx
@@ -20,13 +20,7 @@ import { HeaderButton } from './HeaderButtons.styles';
 import { UnreadNotificationsBadge } from './UnreadNotificationsBadge';
 
 export function HomeButtons(props: HomeButtonsPropTypes) {
-  const {
-    layoutRef,
-    onClickRefresh,
-    onClickHistory,
-    onClickSettings,
-    onClickNotifications,
-  } = props;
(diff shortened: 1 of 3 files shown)


widget/embedded/src/components/HeaderButtons/HomeButtons.tsx
```diff
@@ -50,25 +44,24 @@ export function HomeButtons(props: HomeButtonsPropTypes) {
...
-              size="small"
-              variant="ghost"
-              onClick={onClickNotifications}>
-              <NotificationsIcon size={18} color="black" />
-              <UnreadNotificationsBadge />
-            </HeaderButton>
-          </Popover>
-        </Tooltip>
+          content={<NotificationContent />}>
+          <div>
```
samobasquiat, line 53: I fixed it in another way.
what do you think?
https://github.com/rango-exchange/rango-client/pull/488/files

samobasquiat (changes requested):

samobasquiat (approved):

RanGojo (approved):

PR #498, opened 2024-01-01
fix: persist language in store

nikaaru:
# Summary

When navigating to a new route, the language restores to the default state (En).
this issue was fixed by persistent language in the store.

# Checklist:

### DIFF
--- a/widget/embedded/src/containers/App/App.tsx
+++ b/widget/embedded/src/containers/App/App.tsx
@@ -20,7 +20,7 @@ export function Main() {
 
   const { config } = useAppStore();
   const { activeTheme } = useTheme(config?.theme || {});
-  const { activeLanguage, changeLanguage } = useLanguage();
+  const { activeLanguage } = useLanguage();
   const [lastConnectedWalletWithNetwork, setLastConnectedWalletWithNetwork] =
     useState<string>('');
   const [disconnectedWallet, setDisconnectedWallet] = useState<WalletType>();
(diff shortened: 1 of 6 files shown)


widget/embedded/src/containers/App/App.tsx
```diff
@@ -30,9 +30,6 @@ export function Main() {
     void useNotificationStore.persist.rehydrate();
     widgetContext.onConnectWallet(setLastConnectedWalletWithNetwork);
   }, []);
-  useEffect(() => {
-    changeLanguage(config?.language);
```
mikasackermn, line 34: It is impossible to alter the playground's language. Please review this line of code as it relates to it.

mikasackermn (changes requested):

mikasackermn (approved):

widget/embedded/src/hooks/useLanguage.ts
```diff
@@ -9,18 +9,20 @@ interface UseLanguage {
...
 export function useLanguage(): UseLanguage {
-  const { setLanguage, language } = useAppStore();
+  const { setLanguage, language, config } = useAppStore();
   const languages = LANGUAGES;
-  const defaultLanguage = DEFAULT_LANGUAGE;
+  const defaultLanguage = config?.language || DEFAULT_LANGUAGE;
 
   return {
-    activeLanguage: language,
+    activeLanguage: language || defaultLanguage,
```
Ikari-Shinji-re, line 22: I believe that if the language in the configuration changes, the language of the widget should also change and take precedence over both the default and user-selected values. Currently, when a user selects a language in the widget, the widget configuration is unable to impact the language. I suggest aligning the behavior with the token selection behavior for a more desirable outcome.

Ikari-Shinji-re (approved):

PR #510, opened 2024-01-08
fix: change widget-iframe height based on it's content

Ikari-Shinji-re:
# Summary

The current version of the widget calculates its height based on the viewport. However, in the iframe version of the widget, we are obliged to consider a fixed value for the iframe that represents the maximum possible height of the widget. Consequently, when the widget is loaded and is placed on the home page, an empty space is created at the bottom of the widget, disrupting the user experience. In this draft, we address this issue by using messaging between the iframe source and the hosting website, setting the iframe height to the actual height occupied by the widget.


yeager-eren:
Closing this PR in favor of https://github.com/rango-exchange/rango-client/pull/546

PR #515, opened 2024-01-10
feat: export useNotification hook

samobasquiat:
# Summary

1. export `notifications`  from  `useWidget` for displaying notifications beyond the widget.
2. Included AddIcon in the icons set.
3. Improved UI of notification & showing all events of a swap in notifications.

# Checklist:


widget/embedded/src/hooks/useNotifications.ts
```diff
@@ -0,0 +1,20 @@
...
+import { useNotificationStore } from '../store/notification';
+import { useUiStore } from '../store/ui';
+
+export function useNotifications() {
+  const setSelectedSwap = useUiStore.use.setSelectedSwap();
+  const { getUnreadNotifications } = useNotificationStore();
+  const navigate = useNavigate();
+  const notifications: Notification[] = getUnreadNotifications();
+  const onSelect = (requestId: Notification['requestId']) => {
+    setSelectedSwap(requestId);
```
RyukTheCoder, line 15: Considering that we are providing access to this function for external dapps, it would be prudent to include a validation check for the 'requestId'.

RyukTheCoder (approved):

widget/embedded/src/hooks/useNotifications.ts
samobasquiat: fixed.

RyukTheCoder (approved):

Ikari-Shinji-re:
I believe it would be more effective to include a check to ensure that if we are already on the selected swap page, invoking the onSelect function should not result in pushing another route to the URL.

Ikari-Shinji-re:
In case a Dapp utilizes the useNotifications hook outside of its browser-router or if the memory-router is employed, it is necessary to address the associated error.

samobasquiat:
> I believe it would be more effective to include a check to ensure that if we are already on the selected swap page, invoking the onSelect function should not result in pushing another route to the history stack.

reverted the changes.

samobasquiat:
> In case a Dapp utilizes the useNotifications hook outside of its browser-router or if the memory-router is employed, it is necessary to address the associated error.

reverted the changes.

widget/embedded/src/containers/WidgetInfo/WidgetInfo.tsx
```diff
@@ -28,6 +29,8 @@ export function WidgetInfo(props: React.PropsWithChildren) {
   const swappers = useAppStore().swappers();
   const loadingStatus = useAppStore().fetchStatus;
   const resetLanguage = useLanguage().resetLanguage;
+  const getUnreadNotifications =
+    useNotificationStore().getUnreadNotifications();
```
Ikari-Shinji-re, line 33: I believe "unreadNotifications" or "notifications" would be a more fitting choice.

samobasquiat: fixed.

Ikari-Shinji-re (approved):

PR #517, opened 2024-01-13
fix: fix connecting wallet issue

nikaaru:
# Summary

When the Wallet state is **connecting** and leaves the page or component, we disconnect all connecting statuses in the clean-up component.

And if the modal connecting wallet is closed, we disconnect the wallet again.

Fixes # (issue)

This issue has been fixed in 3 different places:

-  WalletPage
- ConfirmWalletsModal
- SwapDetail

After clicking to connect the wallet, do not close the main window of the wallet,
so that it remains in the connecting mode.
Then you can click on the back of the browser or close the conecting modal.
After you return to the previous page, Wallet should be disconnected.

# Checklist:

### DIFF
--- a/widget/embedded/src/hooks/useWalletList.ts
+++ b/widget/embedded/src/hooks/useWalletList.ts
@@ -9,12 +9,13 @@ import {
   type WalletType,
   WalletTypes,
 } from '@rango-dev/wallets-shared';
-import { useEffect, useState } from 'react';
+import { useCallback, useEffect, useRef, useState } from 'react';
 
 import { useAppStore } from '../store/AppStore';
 import { useWalletsStore } from '../store/wallets';
 import { configWalletsToWalletName } from '../utils/providers';
 import {
+  hashWalletsState,
   isExperimentalChain,
   mapWalletTypesToWalletInfo,
   sortWalletsBasedOnConnectionState,
@@ -39,6 +40,7 @@ export function useWalletList(params: Params) {
   const { state, disconnect, getWalletInfo, connect } = useWallets();
   const { connectedWallets } = useWalletsStore();
   const blockchains = useAppStore().blockchains();
+  const componentWillUnmount = useRef(false);
 
   /** It can be what has been set by widget config or as a fallback we use all the supported wallets by our library */
   const listAvailableWalletTypes =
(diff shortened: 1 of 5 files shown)


widget/embedded/src/hooks/useWalletList.ts
```diff
@@ -97,9 +97,7 @@ export function useWalletList(params: Params) {
   };
 
   useEffect(() => {
-    return () => {
-      disconnectConnectingWallets();
-    };
+    disconnectConnectingWallets();
   }, []);
```
Ikari-Shinji-re, line 102: I don't think it's a good idea to tidy up the state right after the component mounts for the first time or manage cleanup in another component using this hook (like the wallet page and the confirm wallets modal). It just spreads the logic all over.
You could go for something like this instead :
```ts
//utils.ts
export function hashWalletsState(walletsInfo: WalletInfo[]): string {
    return walletsInfo.map((w) => w.state).join('-');
  }
 
//useWalletList.ts

const disconnectConnectingWallets = useCallback(() => {
    const connectingWallets =
      wallets?.filter((wallet) => wallet.state === WalletState.CONNECTING) ||
      [];
    for (const wallet of connectingWallets) {
      void disconnect(wallet.type);
    }
  }, [hashWalletsState(wallets)]);

  useEffect(() => {
    return () => {
      disconnectConnectingWallets();
    };
  }, [disconnectConnectingWallets]);
```

nikaaru: Fixed

Ikari-Shinji-re (approved):

widget/embedded/src/utils/wallets.ts
```diff
@@ -551,3 +551,7 @@ export const isFetchingBalance = (
   !!connectedWallets.find(
     (wallet) => wallet.chain === blockchain && wallet.loading
   );
+
+export function stringWalletsState(walletsInfo: ModalWalletInfo[]): string {
```
samobasquiat, line 555: Since TypeScript can infer the return type of this function, we don't need ": string".
Nit: renaming the function can be useful.

samobasquiat (approved):

widget/embedded/src/utils/wallets.ts
nikaaru: Fixed

PR #529, opened 2024-01-20
feat: add cosmos account and signer for math wallet

RyukTheCoder:
# Summary

Implemented Cosmos connect account and sign transaction for Math Wallet using CosmJs library.

- [x] Connected Math Wallet Cosmos account
- [x] Signed and broadcasted a transaction from Atom on Cosmos to Osmo on Osmosis

# Checklist:

### DIFF
--- a/wallets/provider-math-wallet/src/index.ts
+++ b/wallets/provider-math-wallet/src/index.ts
@@ -16,11 +16,7 @@ import {
   subscribeToEvm,
   WalletTypes,
 } from '@rango-dev/wallets-shared';
-import {
-  cosmosBlockchains,
-  evmBlockchains,
-  solanaBlockchain,
-} from 'rango-types';
+import { evmBlockchains, solanaBlockchain } from 'rango-types';
 
 import {
   getNonEvmAccounts,
@@ -87,7 +83,9 @@ export const getWalletInfo: (allBlockChains: BlockchainMeta[]) => WalletInfo = (
 ) => {
   const evms = evmBlockchains(allBlockChains);
   const solana = solanaBlockchain(allBlockChains);
-  const cosmos = cosmosBlockchains(allBlockChains);
+  const cosmos = allBlockChains.filter(
+    (blockchainMeta) => blockchainMeta.name === Networks.COSMOS
+  );
   return {
     name: 'Math Wallet',
     img: 'https://raw.githubusercontent.com/rango-exchange/assets/main/wallets/math/icon.svg',
--- a/wallets/provider-math-wallet/src/index.ts
+++ b/wallets/provider-math-wallet/src/index.ts
@@ -16,7 +16,11 @@ import {
   subscribeToEvm,
   WalletTypes,
 } from '@rango-dev/wallets-shared';
-import { evmBlockchains, solanaBlockchain } from 'rango-types';
+import {
+  cosmosBlockchains,
+  evmBlockchains,
+  solanaBlockchain,
+} from 'rango-types';
 
 import {
   getNonEvmAccounts,
@@ -83,6 +87,7 @@ export const getWalletInfo: (allBlockChains: BlockchainMeta[]) => WalletInfo = (
 ) => {
   const evms = evmBlockchains(allBlockChains);
   const solana = solanaBlockchain(allBlockChains);
+  const cosmos = cosmosBlockchains(allBlockChains);
   return {
     name: 'Math Wallet',
     img: 'https://raw.githubusercontent.com/rango-exchange/assets/main/wallets/math/icon.svg',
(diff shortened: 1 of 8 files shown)


.eslintrc.json
```diff
@@ -21,8 +21,8 @@
   "parserOptions": {
     "ecmaVersion": "latest",
     "sourceType": "module",
-    // "project": ["tsconfig.base.json"]
-    "project": true
+    "project": ["tsconfig.base.json"]
```
mikasackermn, line 24: Please change this file.

wallets/provider-math-wallet/src/index.ts
```diff
@@ -45,6 +51,11 @@ export const connect: Connect = async ({ instance, meta }) => {
   const nonEvmResults = await getNonEvmAccounts(instance);
   results = [...results, ...nonEvmResults];
 
+  if (cosmosInstance) {
```
mikasackermn, line 54: Kindly insert this code into the function called getNonEvmAccounts.

.eslintrc.json
RyukTheCoder: This change is required because I changed the "global-wallets-env.d.ts" file. As a result of the change in that file, it becomes part of the staged changes, leading to an encountered error during ESLint's linting process. Consequently, I had to modify this line in the ESLint config file to address the issue.

wallets/provider-math-wallet/src/index.ts
RyukTheCoder: Done

mikasackermn (approved):

wallets/shared/src/rango.ts
```diff
@@ -154,6 +154,8 @@ export const KEPLR_COMPATIBLE_WALLETS: string[] = [
   WalletTypes.LEAP_COSMOS,
 ];
 
+export const DEFAULT_COSMOS_RPC_URL = 'https://rpc.cosmos.directory/cosmoshub';
+
```
Ikari-Shinji-re, line 158: Please modify the RPC URL and relocate it within the Math wallet package.

wallets/provider-math-wallet/src/index.ts
```diff
@@ -83,6 +87,7 @@ export const getWalletInfo: (allBlockChains: BlockchainMeta[]) => WalletInfo = (
 ) => {
   const evms = evmBlockchains(allBlockChains);
   const solana = solanaBlockchain(allBlockChains);
+  const cosmos = cosmosBlockchains(allBlockChains);
```
Ikari-Shinji-re, line 90: If Math Wallet is designed specifically for the Cosmos blockchain, it's crucial to return only the meta for the Cosmos blockchain instead of including all blockchains with the transaction type as Cosmos. This is essential to avoid potential issues in other packages like widget-embedded.

Ikari-Shinji-re:
I've examined this route in the widget with the changes from this pull request, along with the current Dapp using the Keplr wallet :
"http://localhost:3002/?fromBlockchain=OSMOSIS&fromToken=OSMO&toBlockchain=COSMOS&toToken=ATOM&fromAmount=0.01"
In the Dapp, the transaction successfully completes in less than a minute. However, in the widget, even after a prolonged period, the swap status remains stuck in the check-status step. I suspect there might be an issue with the parameters we are passing to the Math Wallet.

wallets/provider-math-wallet/src/index.ts
RyukTheCoder: Fixed. Thanks

wallets/shared/src/rango.ts
RyukTheCoder: I moved it here because it is not something specific to MathWallet and can be used for any Cosmos transaction. If you still think I should move it let me know.

RyukTheCoder:
> I've examined this route in the widget with the changes from this pull request, along with the current Dapp using the Keplr wallet : /?fromBlockchain=OSMOSIS&fromToken=OSMO&toBlockchain=COSMOS&toToken=ATOM&fromAmount=0.01 In the Dapp, the transaction successfully completes in less than a minute. However, in the widget, even after a prolonged period, the swap status remains stuck in the check-status step. I suspect there might be an issue with the parameters we are passing to the Math Wallet.

Actually it seems to be a bug in the backend project. In the widget the transaction completes in less than a minute too. You can check it by the url you receive from the "check_status" api. The problem is that the backend service sometimes is not able to detect that the transaction is compete and returns "running" status.

RyukTheCoder:
> I've examined this route in the widget with the changes from this pull request, along with the current Dapp using the Keplr wallet : /?fromBlockchain=OSMOSIS&fromToken=OSMO&toBlockchain=COSMOS&toToken=ATOM&fromAmount=0.01 In the Dapp, the transaction successfully completes in less than a minute. However, in the widget, even after a prolonged period, the swap status remains stuck in the check-status step. I suspect there might be an issue with the parameters we are passing to the Math Wallet.

This bug is resolved

Ikari-Shinji-re (approved):

PR #541, opened 2024-01-23
feat: Adding title to config and export IDs for accessing blockchain image and swap input

RyukTheCoder:
# Summary

Applied Leshka friendly styles and configs to widget.

Fixes # 1208

Please refer to the PR at [https://github.com/rango-exchange/leshka-app/pull/3](https://github.com/rango-exchange/leshka-app/pull/3) for detailed testing steps and guidance.

# Checklist:

### DIFF
--- a/widget/embedded/src/index.ts
+++ b/widget/embedded/src/index.ts
@@ -55,6 +55,7 @@ import {
 import { Networks, WalletTypes } from '@rango-dev/wallets-shared';
 import { PendingSwapNetworkStatus } from 'rango-types';
 
+import { WIDGET_UI_ID as UI_ID } from './constants';
 import { WidgetWallets } from './containers/Wallets';
 import { Widget } from './containers/Widget';
 import { useWidget } from './containers/WidgetInfo';
@@ -105,6 +106,7 @@ export {
   StepEventType,
   StepExecutionEventStatus,
   StepExecutionBlockedEventStatus,
+  UI_ID,
 };
 
 // Internal type exports for Rango
--- a/widget/embedded/src/pages/Home.tsx
+++ b/widget/embedded/src/pages/Home.tsx
@@ -62,7 +62,7 @@ export function Home() {
     setQuoteWarningsConfirmed,
   } = useQuoteStore();
 
-  const fetchMetaStatus = useAppStore().fetchStatus;
+  const { config, fetchStatus: fetchMetaStatus } = useAppStore();
 
   const { connectedWallets, getBalanceFor } = useWalletsStore();
   const { isActiveTab } = useUiStore();
@@ -152,7 +152,7 @@ export function Home() {
           navigate(navigationRoutes.wallets);
         },
         hasBackButton: false,
-        title: i18n.t('Swap'),
+        title: config.title || i18n.t('Swap'),
         suffix: (
           <HomeButtons
             onClickRefresh={
--- a/widget/embedded/src/store/slices/config.ts
+++ b/widget/embedded/src/store/slices/config.ts
@@ -3,6 +3,7 @@ import type { StateCreatorWithInitialData } from '../app';
 
 export const DEFAULT_CONFIG: WidgetConfig = {
   apiKey: '',
+  title: undefined,
   multiWallets: true,
   enableNewLiquiditySources: true,
   customDestination: true,
--- a/widget/ui/src/constants/index.ts
+++ b/widget/ui/src/constants/index.ts
@@ -1,0 +1,6 @@
+export const UI_ID = {
+  SWAP_FROM_INPUT_CONTAINER_ID: 'rango-swap-from-input-container',
+  SWAP_TO_INPUT_CONTAINER_ID: 'rango-swap-to-input-container',
+  SWAP_FROM_CHAIN_IMAGE_ID: 'rango-swap-from-chain-image',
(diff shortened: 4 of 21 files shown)


widget/embedded/src/pages/Home.tsx
```diff
@@ -152,7 +152,7 @@ export function Home() {
           navigate(navigationRoutes.wallets);
         },
         hasBackButton: false,
-        title: i18n.t('Swap'),
+        title: typeof config.title === 'string' ? config.title : i18n.t('Swap'),
```
yeager-eren, line 155:
```suggestion
        title: config.title || i18n.t('Swap'),
```

widget/embedded/src/store/slices/config.ts
```diff
@@ -1,8 +1,11 @@
 import type { WidgetConfig } from '../../types';
 import type { StateCreatorWithInitialData } from '../app';
 
+import { i18n } from '@lingui/core';
+
 export const DEFAULT_CONFIG: WidgetConfig = {
   apiKey: '',
+  title: i18n.t('Swap'),
```
yeager-eren, line 8: If it has default value, why the type is optional and in `widget/embedded/src/pages/Home.tsx` we've added a fallback? 
Please use `undefined` here or if you want to keep this value here, update the type to be only string and remove the condition you've added to `widget/embedded/src/pages/Home.tsx`

widget/ui/src/containers/SwapInput/TokenSection.types.ts
```diff
@@ -3,6 +3,7 @@ export type TokenSectionProps = {
   tokenImage: string;
   tokenSymbol: string;
   chain: string;
+  mode?: 'To' | 'From';
```
yeager-eren, line 6: Please change it to `chianImageId?: string`

yeager-eren (changes requested):
Our goal is giving the ability to override some specific places externally by provide some HTML ids. So we need to make sure we are maintaining the IDs in proper way and they wouldn't be break mistakenly during refcctors. So I suggest to keep the IDs in a separate files and improt IDs from there.

You can use `widget/embedded/src/constants/index.ts` and create a new object like this:

```js
const UI_ID = {
    FROM_CHAIN_IMAGE: "rango-swap-from-chain-image",
    ...
}
```

Note 1: We have `RANGO_SWAP_BOX_ID` as well. Please move that value inside your newly created object.

Note 2: Export this object from `embedded` then update your corresponding PR to use this object as well.

widget/embedded/src/pages/Home.tsx
RyukTheCoder: fixed

widget/embedded/src/store/slices/config.ts
RyukTheCoder: I changed it to `""`

widget/ui/src/containers/SwapInput/TokenSection.types.ts
RyukTheCoder: Done

widget/embedded/src/store/slices/config.ts
yeager-eren: So why the type for `title` has `undefined`? When it happens?

widget/ui/src/constants/index.ts
```diff
@@ -0,0 +1,6 @@
+export const UI_ID = {
+  RANGO_SWAP_FROM_INPUT_CONTAINER_ID: 'rango-swap-from-input-container',
```
yeager-eren, line 2: You can remove `RANGO_` prefix from these values.

widget/embedded/src/index.ts
```diff
@@ -126,3 +127,5 @@ export {
   PendingSwapNetworkStatus,
   EventSeverity,
 };
+// Internal constant exports for Rango
```
yeager-eren, line 130: It's not internal, it can be used by anyone.

widget/embedded/src/store/slices/config.ts
RyukTheCoder: Because it is possible for the user to not send any title as props and if I remove the undefined, the user will encounter a typescript error indicating that the title can not be undefined.

widget/ui/src/constants/index.ts
RyukTheCoder: Done

widget/embedded/src/index.ts
RyukTheCoder: Changed to "Constant exports for Rango"

widget/embedded/src/store/slices/config.ts
yeager-eren: Got it, so using `undefiend` for initial value is more appropriate.

widget/embedded/src/index.ts
yeager-eren: You don't need to export in a separate place. Just add it next to `StepExecutionBlockedEventStatus`

widget/embedded/src/store/slices/config.ts
RyukTheCoder: Changed

widget/embedded/src/index.ts
RyukTheCoder: Done

yeager-eren (approved):
Thanks. looks good to me.

PR #544, opened 2024-01-28
feat: add production release workflow

nikaaru:
# Summary

A new workflow named **Production Release**  has been added, which takes the latest translation changes on the **next** branch and then automatically performs the production release.

Testing scenarios:

Making a pull request on next then merge the PR, After merging PR into next, You can run yarn **production release** workflow manually, it should pull crowdin change on **next** branch and then it will automatically published.

# Checklist:

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,13 +1,9 @@
-name: Publish
+name: Release
 on:
-  push:
-    branches:
-      - 'main'
-      - 'next'
+  workflow_dispatch:
 
 jobs:
   localization:
-    if: ${{ github.ref == 'refs/heads/next' }}
     uses: ./.github/workflows/crowdin.yml
     secrets:
       CROWDIN_PROJECT_ID: ${{ secrets.CROWDIN_PROJECT_ID }}
@@ -16,8 +12,7 @@ jobs:
     with:
       create_pull_request: true
 
-  publish:
-    if: ${{ always() }}
+  release:
     runs-on: ubuntu-latest
     needs:
       - localization
@@ -26,15 +21,14 @@ jobs:
         with:
           fetch-depth: 0
           token: ${{ secrets.PAT }}
-      - run: git pull
 
       - name: Prepare
         uses: ./.github/actions/prepare
         with:
           NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
 
-      - name: Build, Version & Publish packages
-        run: yarn run publish
+      - name: Release Production
+        run: yarn run release-prod
         env:
           REF: ${{ github.ref }}
           GH_TOKEN: ${{ github.token }}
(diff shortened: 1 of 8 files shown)


samobasquiat:
LGTM, but a few questions:
1- Regarding the file extensions, we currently have both .yml and .yaml files. Wouldn't it be better to standardize on just one of them?
2- I noticed that in the new workflow, as well as potentially others, we're utilizing the `prepare` (=> "install packages") actions. Additionally, these actions are used in the `crowdin` workflow. However, in the `release` and `publish` workflows, we're using both of these actions again. Is it ok to run the `prepare` action twice?
3- Is it possible to limit the dropdown options for branches in the new workflow?
4- In cases where we already have a Post-released Pull Request, and a new release commit is made on that PR, the "conventional commit" workflow is triggered. However, it fails with the following error: `a pull request for branch "main" into branch "next" already exists`. Could we resolve this issue?

nikaaru:
> LGTM, but a few questions: 1- Regarding the file extensions, we currently have both .yml and .yaml files. Wouldn't it be better to standardize on just one of them? 2- I noticed that in the new workflow, as well as potentially others, we're utilizing the `prepare` (=> "install packages") action. Additionally, this action is used in the `crowdin` workflow. However, in the `release` and `publish` workflows, we're using both of these actions again. Is it ok to run the `prepare` action twice? 3- Is it possible to limit the dropdown options for branches in the new workflow? 4- In cases where we already have a Post-released Pull Request, and a new release commit is made on that PR, the "conventional commit" workflow is triggered. However, it fails with the following error: `a pull request for branch "main" into branch "next" already exists`. Could we resolve this issue? 5- When an error occurs in the localization workflow, the release workflow is skipped. Is it ok?

1- I think the .yml format is correct. I used the same format for the release workflow, if other places are needed, we can modify it later.

2- The installed files are cached and if there are no changes, it is read from the cache.

3- I check this point.

4- This issue already existed, but it was decided to change this item in the PR related to **hotfix**, with the new changes, this item will also be fixed.

5- This scenario was supposed to work the same way. That is, if the first job has an error, do not start the second job at all.

samobasquiat (approved):

.github/workflows/release.yml
```diff
@@ -0,0 +1,37 @@
...
+  workflow_dispatch:
+    inputs:
+      branch:
+        description: 'to trigger the workflow manually'
+        required: false
+        type: string
+
+jobs:
+  localization:
+    uses: ./.github/workflows/crowdin.yml
```
yeager-eren, line 12: Please pass the `next` explicitly here.

nikaaru: Fixed

yeager-eren:
We are going to release today, I will merge this PR, but it needs some tuning on hotfix flow as we talked.

PR #546, opened 2024-01-29
feat: add dynamicHeight to iframe

yeager-eren:
# Summary

We've a new feature for iframes to be able to get the swap box/widget height. There are some solutions to solve this:

1. Adding a new config to detect iframe
2. Using message passing to set some inital values (widget url is need to send messages between origins)
3. Using query param to set the initial value

The first one will pollute our public API for widget, so this is last thing I wanted. I tried the second approach but it has some issues regarding we couldn't detect when dom is ready in embedded page in iframe. I chose the last one, I added a `clientUrl` to be set from our iframe script, it will be used to send message between origins.

I also introduced a new standard way for our message passing. Another change is adding a feature management to our iframe script so user can activate them only if they need. 

Fixes RF-1192
Draft PR: https://github.com/rango-exchange/rango-client/pull/510 

You first need to `cd widget/iframe && yarn build ` then copy the `index.js` inside `/dist`.

There is another PR here: https://github.com/rango-exchange/widget-examples/pull/3

You can copy the `index.js` into `/iframe` in `widget-example` then script `src` path to point to your local js file instead of production build.

# Checklist:

### DIFF
--- a/widget/embedded/src/hooks/useIframe/useIframe.ts
+++ b/widget/embedded/src/hooks/useIframe/useIframe.ts
@@ -1,0 +1,55 @@
+import type { Messages } from './useIframe.types';
+
+import { useRef } from 'react';
+
+import { useAppStore } from '../../store/AppStore';
+
+import { isAppLoadedIntoIframe } from './useIframe.helpers';
+
+interface UseIframe {
+  send: (message: Messages) => void;
+  connectHeightObserver: (element: HTMLElement) => void;
+  disconnectHeightObserver: () => void;
+}
+
+function useIframe(): UseIframe {
+  const heightObserver = useRef<ResizeObserver | null>(null);
+  const { iframe } = useAppStore();
+  const isMessagePassingAvailable = isAppLoadedIntoIframe() && iframe.clientUrl;
+
+  const send = (message: Messages) => {
+    if (isMessagePassingAvailable) {
+      window.top?.postMessage(message, iframe.clientUrl!);
+    }
+  };
+
+  const connectHeightObserver = (element: HTMLElement) => {
+    heightObserver.current = new ResizeObserver((entries) => {
+      for (const entry of entries) {
+        send({
+          type: 'widget_height',
+          data: {
+            height: entry.contentRect.height,
+          },
+        });
+      }
+    });
+
+    heightObserver.current.observe(element);
+  };
+
+  const disconnectHeightObserver = () => {
+    if (heightObserver.current) {
+      heightObserver.current.disconnect();
+      heightObserver.current = null;
+    }
+  };
+
+  return {
+    send,
+    connectHeightObserver,
+    disconnectHeightObserver,
+  };
+}
+
+export { useIframe };
--- a/widget/embedded/src/hooks/useSyncUrlAndStore/useSyncUrlAndStore.ts
+++ b/widget/embedded/src/hooks/useSyncUrlAndStore/useSyncUrlAndStore.ts
@@ -33,6 +33,7 @@ export function useSyncUrlAndStore() {
   const blockchains = useAppStore().blockchains();
   const tokens = useAppStore().tokens();
   const isInRouterContext = useInRouterContext();
+  const { updateIframe } = useAppStore();
 
   const getUrlSearchParams = () => {
     const fromAmount = searchParams.get(SearchParams.FROM_AMOUNT);
@@ -41,6 +42,7 @@ export function useSyncUrlAndStore() {
     const toBlockchain = searchParams.get(SearchParams.TO_BLOCKCHAIN);
     const toToken = searchParams.get(SearchParams.TO_TOKEN);
     const autoConnect = searchParams.get(SearchParams.AUTO_CONNECT);
+    const clientUrl = searchParams.get(SearchParams.CLIENT_URL);
 
     return {
       fromAmount,
@@ -49,6 +51,7 @@ export function useSyncUrlAndStore() {
       toBlockchain,
       toToken,
       autoConnect,
+      clientUrl,
     };
   };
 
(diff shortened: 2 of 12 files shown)


widget/embedded/src/hooks/useIframe/useIframe.ts
```diff
@@ -0,0 +1,55 @@
...
+import { isAppLoadedIntoIframe } from './useIframe.helpers';
+
+interface UseIframe {
+  send: (message: Messages) => void;
+  connectHeightObserver: (element: HTMLElement) => void;
+  disconnectHeightObserver: () => void;
+}
+
+function useIframe(): UseIframe {
+  const heightObserver = useRef<any>(null);
```
Ikari-Shinji-re, line 16: you can remove any type and use "ResizeObserver | null"

widget/embedded/src/hooks/useIframe/useIframe.ts
```diff
@@ -0,0 +1,55 @@
+import type { Messages } from '@lingui/core';
```
Ikari-Shinji-re, line 1: You've brought in the incorrect type declaration for Messages.

widget/embedded/src/hooks/useSyncUrlAndStore/useSyncUrlAndStore.ts
```diff
@@ -73,6 +76,7 @@ export function useSyncUrlAndStore() {
         [SearchParams.TO_TOKEN]: tokenToSearchParam(toToken),
         [SearchParams.FROM_AMOUNT]: inputAmount,
         [SearchParams.AUTO_CONNECT]: autoConnect ?? undefined,
+        [SearchParams.CLIENT_URL]: clientUrl ?? undefined,
```
Ikari-Shinji-re, line 79: Do we always need to include the clientUrl in the URL?

widget/iframe/src/styles.ts
```diff
@@ -1,7 +1,19 @@
...
+  border: none;
+`;
+
+export function applyDefaultStyles(element: HTMLIFrameElement) {
+  element.style.cssText = commonStyles;
+  element.style.height = '100vh';
+  element.style.maxHeight = '700px';
+}
+export function applyDynamicHeightStyles(element: HTMLIFrameElement) {
+  element.style.backgroundColor = 'transparent';
```
Ikari-Shinji-re, line 17: Transparent background color is already present in common styles; you can exclude it from here.

widget/iframe/src/utils.ts
```diff
@@ -0,0 +1,13 @@
+// URL.canParse hasn't good support yet, so using a polyfill if it's not available.
+export function canParse(url: string) {
+  if ('canParse' in window.URL) {
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore
```
Ikari-Shinji-re, line 5: You can delete these comments.

widget/iframe/src/main.ts
```diff
@@ -50,25 +135,58 @@ export class RangoWidget {
...
+   */
+  private listenToMessages() {
+    window.addEventListener('message', (e) => {
+      if (this.features.dynamicHeight && e.data.type === 'widget_height') {
+        this.onWidgetHeightChange(e.data);
+      }
+    });
+  }
+
+  private onWidgetHeightChange(message: any) {
```
Ikari-Shinji-re, line 170: Is it possible to eliminate any type from this section?

Ikari-Shinji-re:
If the user selects the dynamic height option, a layout shift occurs after the widget is loaded, and I believe this negatively impacts the user experience. Is there anything we can do to address this issue?

Ikari-Shinji-re:
When dynamic height is chosen, navigating from the home page to other pages with a fixed height reveals a jump or layout shift in the width of the widget. To enhance visibility, consider adding a high-contrast background color to the iframe element.

widget/iframe/src/main.ts
yeager-eren: For avoiding duplicating types, I used any here. it should be export from `embeded`, but needs some extra works to export a new types of internal things. 
I tried to make it more clearer and write about next steps on top of if.

widget/iframe/src/utils.ts
yeager-eren: It will go through this error:
```
Property 'canParse' does not exist on type '{ new (url: string | URL, base?: string | URL | undefined): URL; prototype: URL; createObjectURL(obj: Blob | MediaSource): string; revokeObjectURL(url: string): void; }'.ts(2339)
```

widget/iframe/src/styles.ts
yeager-eren: fixed

widget/embedded/src/hooks/useSyncUrlAndStore/useSyncUrlAndStore.ts
yeager-eren: When user activate `dynamicHeight` it needs to setup a communication channel using message passing which has need origin.
Do I'm missing any point here?

widget/embedded/src/hooks/useIframe/useIframe.ts
yeager-eren: I'm tired of this. I tried first and it doesn't work, then I tried it right now it's working. sigh...
fixed.

widget/embedded/src/hooks/useIframe/useIframe.ts
yeager-eren: NIce catch. fixed.

Ikari-Shinji-re:
> > When dynamic height is chosen, navigating from the home page to other pages with a fixed height reveals a jump or layout shift in the width of the widget. To enhance visibility, consider adding a high-contrast background color to the iframe element.
> 
> Thanks, Nice catch! Reason was using `cssText`, it's reset the styles and then apply new styles which caused this issue. I changed the way we apply styles. It should be better now.

I gave it another test, and the issue is still present.

yeager-eren:
As we talked offline, we decided to keep query param and also there is no straight forward way to avoid that layout shift. So it's ready for a final review.

Ikari-Shinji-re (approved):

PR #559, opened 2024-02-04
fix: remove auto scroll behavior from collapsible components

Ikari-Shinji-re:
# Summary

In this pull request, we eliminated the auto-scroll feature present in collapsible components such as the quote and fee modal, which used to trigger the scrolling of the entire app utilizing the widget. Our plan is to implement a more robust solution for this behavior in the future.

# Checklist:


yeager-eren (approved):
Thanks.

PR #562, opened 2024-02-05
fix: apply playground improvements

RyukTheCoder:
# Summary

- [x] title for language setting
- [x] menu active state
- [x] reset for custom colors doesn't seem to work
- [x] disable text for max height / max width
- [x] Zoom on playground, make buttons out of screen.

Tested this by running playground on Chrome browser and checking each of listed items.

# Checklist:

### DIFF
--- a/widget/playground/src/components/SideNavigation/SideNavigation.styles.ts
+++ b/widget/playground/src/components/SideNavigation/SideNavigation.styles.ts
@@ -48,14 +48,12 @@ export const Tab = styled(Flex, {
 });
 
 export const Indicator = styled('div', {
-  width: '7px',
-  height: '64px',
-  borderRadius: '0 20px 20px 0',
+  width: '2px',
   backgroundColor: '$secondary500',
   position: 'absolute',
   transition: 'transform 0.4s ease-in-out',
   left: 0,
-  top: 20,
+  top: 0,
 });
 
 export const TabsContainer = styled(Flex, {
--- a/widget/playground/src/constants/presets.ts
+++ b/widget/playground/src/constants/presets.ts
@@ -4,6 +4,8 @@ import type {
   WidgetColorsKeys,
 } from '@rango-dev/widget-embedded';
 
+const DEFAULT_PRESET_ID = 1;
+
 export const TABS: { id: Mode; title: string }[] = [
   {
     id: 'light',
@@ -279,20 +281,8 @@ export const WIDGET_COLORS: { key: WidgetColorsKeys; label: string }[] = [
   },
 ];
 export const DEFAULT_COLORS = {
-  dark: {
-    primary: undefined,
-    secondary: undefined,
-    neutral: undefined,
-    background: undefined,
-    foreground: undefined,
-    info: undefined,
-  },
-  light: {
-    primary: undefined,
-    secondary: undefined,
-    neutral: undefined,
-    background: undefined,
-    foreground: undefined,
-    info: undefined,
-  },
+  dark: PRESETS.find((preset) => preset.id === DEFAULT_PRESET_ID)
+    ?.dark as WidgetColors,
+  light: PRESETS.find((preset) => preset.id === DEFAULT_PRESET_ID)
+    ?.light as WidgetColors,
 };
--- a/widget/playground/src/containers/configContainer/ConfigContainer.styles.ts
+++ b/widget/playground/src/containers/configContainer/ConfigContainer.styles.ts
@@ -35,11 +35,10 @@ export const LeftSide = styled('div', {
 });
 
 export const Main = styled('div', {
-  justifyContent: 'center',
-  alignItems: 'center',
   flexDirection: 'column',
   display: 'flex',
   width: '100%',
+  overflowY: 'clip',
 });
 
 export const HeaderContainer = styled('div', {
@@ -62,9 +61,10 @@ export const ResetButton = styled(StyledButton, {
 });
 
 export const Content = styled('div', {
-  height: '100%',
+  flex: '1 1 0%',
+  overflow: 'auto',
   display: 'flex',
-  justifyContent: 'center',
+  justifyContent: 'safe center',
   alignItems: 'center',
   flexDirection: 'column',
 });
--- a/widget/playground/src/containers/configContainer/ConfigContainer.tsx
+++ b/widget/playground/src/containers/configContainer/ConfigContainer.tsx
@@ -5,6 +5,7 @@ import {
   Divider,
   HeightIcon,
   LogoWithTextIcon,
+  NotSelectableTypography,
   Switch,
   Typography,
   WidthIcon,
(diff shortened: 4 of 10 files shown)


widget/playground/src/containers/configContainer/ConfigContainer.styles.ts
```diff
@@ -62,9 +61,11 @@ export const ResetButton = styled(StyledButton, {
...
-  height: '100%',
+  width: '100%',
+  flex: '1 1 0%',
+  overflow: 'auto',
   display: 'flex',
-  justifyContent: 'center',
+  justifyContent: 'safe center',
   alignItems: 'center',
   flexDirection: 'column',
 });
```
samobasquiat, line 71: these changes affected boundary guide.

samobasquiat (changes requested):

samobasquiat:
The indicator height is too large. I think it's not good. What do you think?

widget/playground/src/containers/configContainer/ConfigContainer.styles.ts
RyukTheCoder: fixed

samobasquiat (approved):

widget/playground/src/components/SideNavigation/SideNavigation.styles.ts
```diff
@@ -32,6 +32,7 @@ export const Tab = styled(Flex, {
   padding: '$15 0',
   cursor: 'pointer',
   width: '100%',
+  height: '105px',
```
yeager-eren, line 35: it's defined twice. one here and another one on `widget/playground/src/components/SideNavigation/SideNavigation.Tabs.tsx` and `TAB_HEIGHT`.

Make them using same value.

widget/playground/src/constants/presets.ts
```diff
@@ -279,20 +279,6 @@ export const WIDGET_COLORS: { key: WidgetColorsKeys; label: string }[] = [
...
-  },
-  light: {
-    primary: undefined,
-    secondary: undefined,
-    neutral: undefined,
-    background: undefined,
-    foreground: undefined,
-    info: undefined,
-  },
+  dark: PRESETS.find((preset) => preset.id === 1)?.dark as WidgetColors,
```
yeager-eren, line 282: it can be `undefined` (based on `find(...)?`) but you have added a type assertion which eliminated the `undefined`.

yeager-eren: Please also don't use magic values which means we prefer `const DEFAULT_PRESET_ID = 1;` over using a magic number directly.

widget/playground/src/containers/configContainer/ConfigContainer.styles.ts
```diff
@@ -62,9 +61,10 @@ export const ResetButton = styled(StyledButton, {
...
 
 export const Content = styled('div', {
-  height: '100%',
+  flex: '1 1 0%',
+  overflow: 'auto',
   display: 'flex',
-  justifyContent: 'center',
+  justifyContent: 'safe center',
```
yeager-eren, line 67: Is this a thing in CSS?

widget/ui/src/components/Typography/Typography.styles.tsx
```diff
@@ -33,6 +33,11 @@ export const TypographyContainer = styled('span', {
         whiteSpace: 'nowrap',
       },
     },
+    selectNone: {
```
yeager-eren, line 36: Please don't every css property a prop. and also make it boolean will cause to add `selectAll`, `selectText` and other possible values in future.

yeager-eren: You can cover you use case in playground by doing something like this.

```
const NotSelectableTypograph = styled(Typograph, { ... })
```

yeager-eren (changes requested):

RyukTheCoder:
> The indicator height is too large. I think it's not good. What do you think?

widget/playground/src/components/SideNavigation/SideNavigation.styles.ts
RyukTheCoder: Thanks, Fixed.

widget/playground/src/constants/presets.ts
RyukTheCoder: Thanks, I changed my code to use the "DEFAULT_PRESET_ID" that you suggested. About the other comment, the preset with id equal to 1 is hard coded so it is not possible for the result of the find method to be 'undefined' so I don't get the problem here.

widget/playground/src/containers/configContainer/ConfigContainer.styles.ts
RyukTheCoder: Yes, it is commonly used to address issues with a scrollable flex container. Let's consider a scenario where you have a div element with 'display' set to 'flex', 'justifyContent' set to 'center', and 'overflow' set to 'auto'. If this div element contains a child element with a height greater than the parent, the div element becomes scrollable. However the upper section of the child element may be hidden. To overcome this problem, you can utilize the 'safe center' property.

widget/ui/src/components/Typography/Typography.styles.tsx
RyukTheCoder: Added "NotSelectableTypograph" as a component to UI repository.

widget/playground/src/containers/configContainer/ConfigContainer.styles.ts
yeager-eren: Thanks. I didn't know it exists.
It seems it doesn't has satisfying browser support yet, but I tested it in all major browsers and everything was fine.

widget/playground/src/utils/colors.ts
```diff
@@ -10,15 +12,58 @@ export const getMainColor = (
   const isDarkTab = tab === 'dark';
   const isLightTab = tab === 'light';
   const isAutoTab = tab === 'auto';
+
   if (!!theme?.colors) {
-    if (isAutoTab && !theme.singleTheme) {
-      return (theme?.colors[mode || 'light'] || {})[key];
+    if (isAutoTab) {
```
yeager-eren, line 17: As we talked in tg, you can revert this change and only disable `Custom Color` when the selected preset isn't in the current tab.

RyukTheCoder: Done

widget/playground/src/containers/configContainer/ConfigContainer.tsx
```diff
@@ -67,29 +68,37 @@ export function ConfigContainer(props: PropsWithChildren) {
...
-                <Typography variant="label" size="medium" color="neutral600">
-                  Max Width: 390 px
-                </Typography>
-              </BoundarySize>
-            )}
-            <BoundaryGuide
-              style={{ borderWidth: showBoundaryGuide ? '1px' : 0 }}>
-              {props.children}
-            </BoundaryGuide>
+            <div>
```
yeager-eren, line 71: Why this `div` needed?

RyukTheCoder: To address the issue depicted in the image, a container needs to be created to properly align the texts related to the boundary guide. The container should have a width that fits the dimensions of the boundary guide to ensure correct placement.

image

yeager-eren: It's kinda weird, but if it's need, please put a comment on top of it and explain the reason because it can be removed in future easily since removing it will cause problem on a specific situation.

RyukTheCoder: Done

yeager-eren (approved):
Thanks. I'll merge it.

PR #565, opened 2024-02-05
fix: fix some swap messages in widget-embedded

Ikari-Shinji-re:
# Summary

Some displayed messages during the swap process are incorrect.

Fixes # (issue)

- [x] The "change network" message is displayed inaccurately for non-EVM blockchains: 

- [x] Consider displaying a distinct message if the required wallet is not connected, and present a separate message if the connected wallet is associated with a different account :

Initiate a swap on the Osmosis network, proceed to the sign transaction pop-up, and without signing the transaction, refresh the page. After the page reloads, attempt to connect the necessary wallet on the swap-details page.

# Checklist:

### DIFF
--- a/widget/embedded/src/containers/Wallets/Wallets.tsx
+++ b/widget/embedded/src/containers/Wallets/Wallets.tsx
@@ -3,15 +3,14 @@ import type {
   PropTypes,
   WidgetContextInterface,
 } from './Wallets.types';
-import type { Wallet } from '../../types/wallets';
 import type { ProvidersOptions } from '../../utils/providers';
 import type { EventHandler } from '@rango-dev/wallets-react';
-import type { Network } from '@rango-dev/wallets-shared';
 import type { PropsWithChildren } from 'react';
 
 import { Events, Provider } from '@rango-dev/wallets-react';
+import { type Network } from '@rango-dev/wallets-shared';
 import { isEvmBlockchain } from 'rango-sdk';
-import React, { createContext, useEffect, useRef, useState } from 'react';
+import React, { createContext, useEffect, useRef } from 'react';
 
 import { useWalletProviders } from '../../hooks/useWalletProviders';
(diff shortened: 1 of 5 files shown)


widget/embedded/src/containers/Wallets/Wallets.tsx
```diff
@@ -67,7 +65,9 @@ function Main(props: PropsWithChildren<PropTypes>) {
           supportedChainNames,
           meta.isContractWallet
         );
-        setAccounts(data);
+        if (data.length) {
```
samobasquiat, line 68: The issue has been fixed, but the twice (or multiple times) calling wallet-details API when switching accounts is back.

samobasquiat (changes requested):

samobasquiat (approved):

RanGojo:
I see this message now for connecting wallet:
`Please connect to keplr by using top right button on page.`
please change it to
`Please connect to keplr wallet.`

widget/embedded/src/containers/Wallets/Wallets.tsx
```diff
@@ -102,12 +102,6 @@ function Main(props: PropsWithChildren<PropTypes>) {
   };
   const isActiveTab = useUiStore.use.isActiveTab();
 
-  useEffect(() => {
-    if (accounts.length) {
-      connectWallet(accounts, tokens);
-    }
-  }, [accounts, tokens]);
-
```
RanGojo, line 110: I'm not sure about this change? What was the initial reason for this code?

RanGojo (changes requested):

widget/embedded/src/containers/Wallets/Wallets.tsx
samobasquiat: It was my change in one of my commits to ensure the switch accounts API call happens once. With the cleanup listener names fixed, we can revert it.

RanGojo (approved):

PR #570, opened 2024-02-05
fix: fix widget ui minor bugs

RyukTheCoder:
# Summary

- [x] fix  input placeholder in Firefox
- [x] some buttons colors are not correct
- [x] fix max button not working when the balance is 0
- [ ] fix animation for modal close -> Will be addressed in a separate PR.

Fixes # 1230

Tested this by running widget on Chrome and Firefox browsers and checking each of listed items.

# Checklist:

### DIFF
--- a/widget/embedded/src/components/ConfirmWalletsModal/ConfirmWalletsModal.tsx
+++ b/widget/embedded/src/components/ConfirmWalletsModal/ConfirmWalletsModal.tsx
@@ -84,17 +84,22 @@ export function ConfirmWalletsModal(props: PropTypes) {
   const isWalletRequiredFor = (blockchain: string) =>
     !!selectedQuote?.swaps.find((swap) => swap.from.blockchain === blockchain);
 
-  const [selectableWallets, setSelectableWallets] = useState<ConnectedWallet[]>(
+  const getInitialSelectableWallets = () =>
     connectedWallets.filter((connectedWallet) => {
       return (
         connectedWallet.selected &&
         requiredWallets.includes(connectedWallet.chain)
       );
-    })
+    });
+
+  const [selectableWallets, setSelectableWallets] = useState<ConnectedWallet[]>(
+    getInitialSelectableWallets()
   );
   const lastStepToBlockchainMeta = blockchains.find(
     (chain) => chain.name === lastStepToBlockchain?.name
   );
+
+  const isInsufficientBalanceModalOpen = balanceWarnings.length > 0;
 
   const isSelected = (walletType: string, chain: string) =>
     !!selectableWallets.find(
@@ -291,7 +296,7 @@ export function ConfirmWalletsModal(props: PropTypes) {
       })}
       anchor="center">
       <WatermarkedModal
-        open={balanceWarnings.length > 0}
+        open={isInsufficientBalanceModalOpen}
         onClose={setBalanceWarnings.bind(null, [])}
         container={modalContainer}>
         <MessageBox
--- a/widget/embedded/src/pages/Home.tsx
+++ b/widget/embedded/src/pages/Home.tsx
@@ -102,7 +102,7 @@ export function Home() {
     formatBalance(fromTokenBalance)?.amount ?? '0';
 
   const tokenBalanceReal =
-    !!fromBlockchain && !!fromToken
+    !!fromBlockchain && !!fromToken && fromTokenBalance?.amount
       ? numberToString(fromTokenBalance?.amount, fromTokenBalance?.decimals)
       : '0';
 
@@ -203,9 +203,7 @@ export function Home() {
               loadingBalance={fetchingBalance}
               tooltipContainer={getContainer()}
               onSelectMaxBalance={() => {
-                if (fromTokenFormattedBalance !== '0') {
-                  setInputAmount(tokenBalanceReal.split(',').join(''));
-                }
+                setInputAmount(tokenBalanceReal.split(',').join(''));
               }}
             />
             <SwitchFromAndToButton />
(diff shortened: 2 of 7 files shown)


widget/embedded/src/pages/Home.tsx
```diff
@@ -203,9 +203,7 @@ export function Home() {
               loadingBalance={fetchingBalance}
               tooltipContainer={getContainer()}
               onSelectMaxBalance={() => {
-                if (fromTokenFormattedBalance !== '0') {
-                  setInputAmount(tokenBalanceReal.split(',').join(''));
-                }
+                setInputAmount(tokenBalanceReal.split(',').join(''));
```
yeager-eren, line 206: it has inconsistent behavior. 
1. Don't select anything (open: `http://localhost:3002/`) then click on Max it will put `0` for input. 
2. Select a token that you have some in your wallet then click on Max it will set the amount that you have. Then select a token that you don't have any, click on Max, it will set '' for input.

widget/embedded/src/pages/ConfirmSwapPage.tsx
```diff
@@ -342,15 +342,13 @@ export function ConfirmSwapPage() {
...
-      {showWallets && (
-        <ConfirmWalletsModal
-          open={showWallets}
-          onClose={() => setShowWallets(false)}
-          onCancel={cancelFetch}
-          loading={fetchingConfirmationQuote}
-          onCheckBalance={onConfirmSwap}
-        />
-      )}
+      <ConfirmWalletsModal
```
yeager-eren, line 345: I'm not sure this would work as expected. On some certain situation `showWallets` will be updated (For example changing router that needs a new wallet to be connect). By removing this condition, `ConfirmWalletsModal` will not be reset anymore. I'm not sure this is expected behavior. please get a confirmation from @Ikari-Shinji-re first.

yeager-eren (changes requested):

widget/embedded/src/pages/Home.tsx
RyukTheCoder: Answered by the order:
1. I don't get how that inconsistent behavior is related to the related issue or my changes. It is declared in the related issue that "max button doesnt work when balance is 0 (it should work anyway)". So I should make sure that the max button should work even if the balance is equal to zero. I think that the problem you are referring to is about if we should display the max button and the balance if the user did not select any token.
2. That happens because 'fromTokenBalance.balance' is 'undefined' in that situation and the value of 'tokenBalanceReal' will be equal to "". I fixed it by setting the input amount equal to "0" in that situation by changing that line of code to "setInputAmount(tokenBalanceReal.split(',').join('') || '0');"

widget/embedded/src/pages/ConfirmSwapPage.tsx
RyukTheCoder: The best idea I came up with was to reset the state in a 'useEffect' but I don't know if that is the best thing to do. Let's request @Ikari-Shinji-re for the second review and see what he thinks about that.

widget/embedded/src/pages/Home.tsx
yeager-eren: the origin of this bug is `tokenBalanceReal` because `numberToString` is returning `''`. You can fix that line by changing it to something like this:
```
  const tokenBalanceReal =
    !!fromBlockchain && !!fromToken && fromTokenBalance?.amount
      ? numberToString(fromTokenBalance?.amount, fromTokenBalance?.decimals)
      : '0';
```

widget/embedded/src/pages/ConfirmSwapPage.tsx
yeager-eren: I'm still not sure this is a good fix. talk with Shinji and come up with another solution please. you can make it a separate task if it takes some good amount of effort.

widget/embedded/src/pages/Home.tsx
RyukTheCoder: Done

widget/embedded/src/components/ConfirmWalletsModal/ConfirmWalletsModal.tsx
```diff
@@ -84,18 +85,26 @@ export function ConfirmWalletsModal(props: PropTypes) {
...
   );
   const lastStepToBlockchainMeta = blockchains.find(
     (chain) => chain.name === lastStepToBlockchain?.name
   );
 
+  const isInsufficientBalanceModalOpen = useMemo(
+    () => balanceWarnings.length > 0,
+    [balanceWarnings]
+  );
+
```
Ikari-Shinji-re, line 107: I think utilizing useMemo might not provide any benefits in this scenario as the computation is not resource-intensive. Furthermore, the issue arises with the balanceWarnings being an array; passing it to the dependency array doesn't work as expected due to its reference being updated after each render.

Ikari-Shinji-re (approved):

widget/embedded/src/components/ConfirmWalletsModal/ConfirmWalletsModal.tsx
```diff
@@ -112,6 +118,16 @@ export function ConfirmWalletsModal(props: PropTypes) {
     lastStepToBlockchainMeta &&
     !isValidAddress(lastStepToBlockchainMeta, destination);
 
+  const resetStates = () => {
```
yeager-eren, line 121: state shouldn't be reset manually. it breaks easily in future.
You can write an higher order function to get `open` and after finishing the exit scenario pass it to actual Modal. As an example when `open` set to `true`, you HOC will get the value and keeps it, then after finishing the exit animation, then you will pass the `open` value to Modal eventually.

RyukTheCoder: Added a component named "ConditionalMountComponent" to ui repository which is able to mount and unmount its children components with a delay.

yeager-eren:
I resolved my comments since it was related to number 4 of the list which is going to be handled in a separate PR.

PR #573, opened 2024-02-06
hotfix: fix tab manager error in Next.js

Ikari-Shinji-re:
# Summary

We have recently introduced a feature that allows for having a single active tab in the widget, but it has been causing issues with the Next.js build. This pull request addresses the issue by fixing the problem and ensuring that the document object is accessed and relevant event listeners are attached only after the widget has been mounted on the DOM.

# Checklist:


yeager-eren (approved):
thanks

PR #579, opened 2024-02-12
feat: add animation to toast

samobasquiat:
# Summary

A transition is added to the toast component. it changed to transform instead of opacity.

Fixes # (issue)

You can test it by storybook. Or linking this PR to alpha branch of dapp.

# Checklist:

### DIFF
--- a/widget/ui/src/components/Toast/Toast.helpers.ts
+++ b/widget/ui/src/components/Toast/Toast.helpers.ts
@@ -5,3 +5,8 @@ export function idGenerator() {
     return id;
   };
 }
+
+export const TOAST_UNMOUNT_DELAY = 500; // the delay before the toast is unmounted
+export const TOAST_TRANSITION_DURATION = 300; // the duration of the toast transition
+export const HOVER_TRANSITION_DURATION = 200; // the duration of the toast transition hover
+export const TOAST_HIDE_DELAY = 200; // the delay before the toast is invisible to shift remained toast smoothly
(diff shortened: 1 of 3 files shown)


widget/ui/src/components/Toast/Toast.helpers.ts
```diff
@@ -5,3 +5,6 @@ export function idGenerator() {
     return id;
   };
 }
+
+export const CLOSED_DELAY = 400;
+export const OPEN_DELAY = 10;
```
RyukTheCoder, line 10: Can you explain why we need "OPEN_DELAY"? I removed it and it worked properly. Also the other variable is named "CLOSED_DELAY " but it is used as transition duration so it is not just for closing the toast and it is used for opening it too. So I think you can improve the name of that variable.

RyukTheCoder (changes requested):

RyukTheCoder (approved):

RyukTheCoder (approved):

widget/ui/src/components/Toast/Toast.helpers.ts
```diff
@@ -5,3 +5,6 @@ export function idGenerator() {
     return id;
   };
 }
+
+export const TOAST_EXIT_DELAY = 400; // the delay before the toast is unmounted
+export const TOAST_TRANSITION_DURATION = 400; // the duration of the toast transition
```
RanGojo, line 10: I think timing doesn't match exactly with the provided video specially in terms of exist delay. Let me double check it with the design team.

RanGojo (changes requested):
Good job for the enter/exit animations. 
I think the hover animation is missed in this PR. (You could check the provided video)

widget/ui/src/components/Toast/Toast.helpers.ts
samobasquiat: I reduced the time.

RanGojo (approved):
@samobasquiat Nice job!

PR #583, opened 2024-02-14
feat: update crowdin workflow and add pretranslate step

nikaaru:
# Summary

crowdin was removed from the publish and check workflows.
From now on, we only have crowdin in the release process
So that for each release:

- First, it extracts the new strings
- Then it pushes new strings to crowdin
- Then the pre-translate script will be called
- Then it downloads the translations for different languages, and opens a PR and merges all new translations to the **Next** branch.
And finally, it continues the release process

To test, you can first take some PRs from **Next** Branch and change the string in some files and merge them into **Next** Branch.
Finally, call the release workflow manually,
In this case, the new strings that are on the next branch must be uploaded to crowdin, pre-translated and new translations must be downloaded for each language and merged in the **Next** branch. And the release process continues

# Checklist:

### DIFF
--- a/scripts/crowdin/pretranslate.mjs
+++ b/scripts/crowdin/pretranslate.mjs
@@ -1,0 +1,60 @@
+import { CrowdinError } from "../common/errors.mjs";
+import { fetchDataWithAuthorization } from "./utils.mjs";
+
+import { 
+  MAXIMUM_PRETRANSLATION_STATUS_CHECK,
+  PRETRANSLATE_API, 
+  REQUEST_INTERVAL_TIMEOUT, 
+  FILE_API,
+  MACHINE_TRANSLATE_API,
+  PROJECT_API,
+} from "./constants.mjs";
+
+
+export const getMachineTranslationEngineID = async () =>{
+  const responseData = await fetchDataWithAuthorization(MACHINE_TRANSLATE_API);
+  if(!responseData.data || !responseData.data.length){
+    throw new CrowdinError('No data received for machine translation');
+  }
+  return responseData.data[0].data.id;
+}
+
+export const getLanguageIds = async () => {
+  const responseData = await fetchDataWithAuthorization(PROJECT_API);
+  return responseData.data.targetLanguageIds;
+};
+
+export const getSourceFileId = async () => {
+  const responseData = await fetchDataWithAuthorization(FILE_API);
+  if (!responseData.data || !responseData.data.length) {
+    throw new CrowdinError('No data received for source file id');
+  }
+  return responseData.data[0].data.id;
+};
+
+export const sendPreTranslateRequest = async ({ sourceFileId, languageIds, preTranslationOption }) => {
+  const responseData = await fetchDataWithAuthorization(PRETRANSLATE_API, 'POST', {
+    ...preTranslationOption,
+    fileIds: [sourceFileId],
+    languageIds,
+  });
+  return responseData.data.identifier;
+};
+
+export const checkPreTranslateStatus = async (preTranslationId) => {
+  const maxAttempts = MAXIMUM_PRETRANSLATION_STATUS_CHECK;
+  let attempt = 0;
+  const delay = (ms) => new Promise((resolve) => setTimeout(resolve, ms));
+  while (attempt < maxAttempts) {
+    const responseData = await fetchDataWithAuthorization(`${PRETRANSLATE_API}/${preTranslationId}`);
+    const status = responseData.data.status;
+    if (status === 'finished') {
+      return status;
+    } else {
+      console.log(`Pre-translation status: ${status}. Retrying in 10 seconds...`);
+      await delay(REQUEST_INTERVAL_TIMEOUT);
+      attempt++;
+    }
+  }
+  throw new CrowdinError('Timeout: Pre-translation did not succeed within the specified time.');
+};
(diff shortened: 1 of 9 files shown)


scripts/crowdin/pretranslate.mjs
```diff
@@ -0,0 +1,159 @@
+const projectId = process.env.CROWDIN_PROJECT_ID;
```
yeager-eren, line 1: This script should be structured like other scripts which means the entry point for running the command will be `command.mjs` (you can keep your helpers here).

yeager-eren: In first lines of your main function in `command.mjs`, please make sure your checking envs and fails early if they are not exist.

scripts/crowdin/pretranslate.mjs
```diff
@@ -0,0 +1,159 @@
...
+const token =  process.env.CROWDIN_PERSONAL_TOKEN;
+
+const BASE_URL = `https://api.crowdin.com/api/v2/projects/${projectId}`;
+const preTranslateURL = `${BASE_URL}/pre-translations`;
+const REQUEST_INTERVAL_TIMEOUT = 10_000;
+const MAXIMUM_REQUEST_RETRIES = 12;
```
yeager-eren, line 7: Naming is a bit misleading. I suggest something like `MAXIMUM_PRETRANSLATATION_STATUS_CHECK` or `MAXIMUM_STATUS_CHECK` or `MAX_STATUS_CHECK`

scripts/crowdin/pretranslate.mjs
```diff
@@ -0,0 +1,159 @@
+const projectId = process.env.CROWDIN_PROJECT_ID;
+const token =  process.env.CROWDIN_PERSONAL_TOKEN;
+
+const BASE_URL = `https://api.crowdin.com/api/v2/projects/${projectId}`;
+const preTranslateURL = `${BASE_URL}/pre-translations`;
```
yeager-eren, line 5: Uppercase please. it's constant, not a normal variable.

yeager-eren: same comment applies to `token` and `projectId` as well.

scripts/crowdin/pretranslate.mjs
```diff
@@ -0,0 +1,159 @@
...
+const REQUEST_INTERVAL_TIMEOUT = 10_000;
+const MAXIMUM_REQUEST_RETRIES = 12;
+
+const requestData = {
+  method: 'mt',
+  engineId: 410374,
```
yeager-eren, line 11: Don't use hardcoded value, engineId seems different in each account. you wrote `getMachineTranslationEngineID` but you are not using it

scripts/crowdin/pretranslate.mjs
```diff
@@ -0,0 +1,159 @@
...
+  autoApproveOption: 'all',
+  duplicateTranslations: false,
+  skipApprovedTranslations: true,
+  translateUntranslatedOnly: true,
+  translateWithPerfectMatchOnly: false,
+};
+
+const getMachineTranslationEngineID = async () =>{
+  try {
+    const response = await fetch('https://api.crowdin.com/api/v2/mts', {
```
yeager-eren, line 21: Why some of api endpoinst are defined as const but here you've used hardcoded value directly?

scripts/crowdin/pretranslate.mjs
```diff
@@ -0,0 +1,159 @@
...
+      throw new Error(`Failed to get source file id`);
+    }
+
+    const responseData = await response.json();
+    const sourceFileId = responseData.data[0].data.id;
+
+    return sourceFileId;
+  } catch (error) {
+    console.error('Error:', error);
+    throw error;
```
yeager-eren, line 82: Use same approach as we did in `scripts/common/errors.mjs` for errors. You can add CrowdinError for instance and use it for your errors here.

yeager-eren: Same comment applies to other errors.

yeager-eren (changes requested):
Thanks. I did my review. I have one question as well. Why we didn't use official crowdin's npm packages? I didn't dig into it but seems there are some packages out there.

yeager-eren:
I also getting bad request on my testing, Do I'm missing something here?
https://github.com/yeager-eren/rango-client/actions/runs/7938518813

scripts/crowdin/pretranslate.mjs
nikaaru: Fixed

scripts/crowdin/pretranslate.mjs
nikaaru: Fixed

scripts/crowdin/pretranslate.mjs
nikaaru: Fixed

scripts/crowdin/pretranslate.mjs
nikaaru: Fixed

scripts/crowdin/pretranslate.mjs
nikaaru: Fixed

scripts/crowdin/pretranslate.mjs
nikaaru: FIxed

PR #601, opened 2024-02-20
feat: add logging

yeager-eren:
## How it works

This is an experimental version of solution for our logging

Here is the high level architecture:

arch

*core*: expose `error`, `warn`, ... can be used by any library or client to emit log records.
*subscriber*: All the log records will be captured by subscriber and will be passed to `layers` to process them.
*console* and *sentry*: Two separate layers to process the logs and one, print to console and the other one send them to sentry.

You can freely record logs everywhere you wanted. just import `core` and use the methods. if there is no subscriber to capture them, nothing happens.

## How to test

It currently happens only on rejecting transactions (logRpc function) from wallet. if you want to test more levels, take a look at this commit: https://github.com/rango-exchange/rango-client/pull/601/commits/e17844b98b3fc00d6085b2ed705d10218326e4f3

There is another PR on [our dApp repo](https://github.com/rango-exchange/app/pull/61), you can see how you can link and run some tests against this PR using dApp.

I also added a subscriber to widget, only on dev mode (localhost). It will be eliminated on production and there is no plan to add it to production yet. So you can run `yarn dev:widget` to test PR as well.

## Next steps

We can add more logging our libs, but one important thing is every packages should have their own error instances (like `RangoPresetError` in this PR), it enriches the log records.

## TODOs

- [x] Initial draft
- [x] Console layer
- [x] Max level filtering
- [ ] Set Max level from env or window
- [x] Sentry Layer
- [x] Customized instance of error for our libraries 
- [x] Caret version?


logging/subscriber/src/helpers.ts
```diff
@@ -0,0 +1,5 @@
+import type { Level } from '@rango-dev/logging-types';
+
+export function isEnabled(maxLevel: Level, level: Level) {
+  return level >= maxLevel;
```
nikaaru, line 4: NIT: I think the name **baseLevel** conveys the meaning better than **maxLevel**

logging/subscriber/src/subscriber.ts
```diff
@@ -0,0 +1,34 @@
...
+
+import { EventType, Level } from '@rango-dev/logging-types';
+
+import { isEnabled } from './helpers';
+
+interface Options {
+  maxLevel: Level;
+}
+
+const DEFAULT_LEVEL = Level.Trace;
```
nikaaru, line 11: It is better to change the **DEFAULT_LEVEL** to cover more events

logging/subscriber/src/subscriber.ts
```diff
@@ -0,0 +1,34 @@
...
+    if (!isEnabled(maxLevel, level)) {
+      // Skipping event.
+      return;
+    }
+
+    for (const layer of layers) {
+      try {
+        layer.handler(event.detail);
+      } catch {
+        break;
```
nikaaru, line 30: we can use **continue** instead of **break**, Because if a layer encounters an error
The other layer can continue to work.

logging/console/src/layer.ts
```diff
@@ -0,0 +1,44 @@
...
+  ];
+}
+
+export function layer(): Layer {
+  return {
+    handler(payload) {
+      const message = formatMessage(payload);
+
+      switch (payload.level) {
+        case 0:
```
nikaaru, line 21: It is better to explicitly use the level name instead of the number

logging/core/src/logging.ts
```diff
@@ -0,0 +1,44 @@
+import type { Data, EventPayload, Message } from '@rango-dev/logging-types';
+
+import { EventType, Level } from '@rango-dev/logging-types';
+
+import { isValidInstance } from './helpers';
+
+export function ev(level: Level, message: Message, data?: Data) {
```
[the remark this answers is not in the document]
yeager-eren: That would be useful when you are going to emit some logs by a list. Conisder this code:

```
const myLogs = [
{level: 'error', message: ..},
{level: 'warn', message: ..}
];

myLogs.forEach(log => ev(log.level, log.message))
```

logging/subscriber/src/subscriber.ts
yeager-eren: It's intentional. it shouldn't proceed.

logging/subscriber/src/helpers.ts
yeager-eren: Agree. Renamed to `baseLevel`.

logging/subscriber/src/subscriber.ts
yeager-eren: `event.level` always should have a valid structure and having `level`, so this fallback was redundant. I removed it and added a runtime check.

logging/console/src/layer.ts
yeager-eren: Thanks. I changed it to use Level enum.

nikaaru (approved):

logging/subscriber/src/subscriber.ts
```diff
@@ -0,0 +1,37 @@
...
+import { isEnabled, isValidEvent } from './helpers';
+
+interface Options {
+  baseLevel: Level;
+}
+
+export function init(layers: Layer[], options: Options) {
+  const { baseLevel } = options;
+
+  document.addEventListener(EventType, function handler(e) {
```
RanGojo, line 14: I think we need to clean up listeners.

logging/sentry/package.json
```diff
@@ -0,0 +1,32 @@
...
+  "files": [
+    "dist",
+    "src"
+  ],
+  "dependencies": {
+    "@rango-dev/logging-types": "^0.1.0"
+  },
+  "devDependencies": {
+    "@sentry/browser": "^7.102.1"
+  },
```
RanGojo, line 21: I think sentry should be a peer dependency or dependency for this package not a dev dependency.

queue-manager/rango-preset/package.json
```diff
@@ -35,7 +34,7 @@
     "@types/uuid": "^8.3.4"
   },
   "dependencies": {
-    "@sentry/browser": "^6.12.0",
+    "@rango-dev/logging-core": "0.1.0",
```
RanGojo, line 37: Is it intentional that we used fix version here?

queue-manager/rango-preset/src/errors.ts
```diff
@@ -0,0 +1,7 @@
+export class RangoPresetError extends Error {
+  name = 'RangoPresetError';
+
+  constructor(msg: string) {
+    super(msg);
+  }
+}
```
RanGojo, line 7: I didn't understand point of declaring this class? why just not using Error class?

RanGojo (approved):
LGTM.

logging/subscriber/src/subscriber.ts
yeager-eren: This should be called once, during the whole app lifetime. Giving that option to clean up the event will be misleading I believe. it can causes to miss events (before init and after clean up) if do something rather than this way and it will be unreliable.

logging/sentry/package.json
yeager-eren: We are only importing some types from the package which is only needed for development and will be omitted during build. 
It's not peer dependency as well, because the package don't need sentry to work. the instance will be passed when app is going to initialize the layer.

queue-manager/rango-preset/package.json
yeager-eren: Nice catch! missed this one.

queue-manager/rango-preset/src/errors.ts
yeager-eren: There are two reasons behind this:
1. This approach will helps us to easily check the error using `instanceof` in runtime. 
2. More readable error on sentry. I attached `before` and `after` of doing this.

Screenshot
Screenshot

PR #611, opened 2024-02-24
feat: add loading state for quote-expanded

samobasquiat:
# Summary

Added a loading state for expanded-quote

the expanded quote has a loading prop or you can use expanded-quote-skeleton directly.

# Checklist:


RyukTheCoder:
The height of RouteContainer is a little different when the value of loading is equal to true. I think it maybe cause small lags when switching the loading value.

image

image

widget/ui/src/components/QuoteExpanded/QuoteExpanded.Skeleton.tsx
```diff
@@ -0,0 +1,95 @@
...
+  RouteContainer,
+  Separator,
+  VerticalLine,
+} from './QuoteExpanded.styles';
+
+const ITEM_SKELETON_COUNT = 3;
+
+export function QuoteSkeletonExpanded() {
+  return (
+    <RouteContainer selected={false}>
```
RyukTheCoder, line 21: I think it is better to use the same RouteContainer and only change the content of it. This way the container will not rerender when switching the loading value. Also you can use other components like RouteHeader and Steps to prevent duplicate code and also keeping the size of elements consistent on loading switch.  And maybe also you can create component just for new utilizations like SkeletonHeader and use them in the main component.

widget/ui/src/components/QuoteExpanded/QuoteExpanded.types.ts
```diff
@@ -0,0 +1,21 @@
...
+      tags: RouteTag[];
+      time: QuoteCostProps['time'];
+      fee: QuoteCostProps['fee'];
+      percentageChange?: PriceImpactProps['percentageChange'];
+      warningLevel?: PriceImpactProps['warningLevel'];
+      outputPrice: SwapInputProps['price'];
+      tooltipContainer?: HTMLElement;
+      onClick?: () => void;
+      selected?: boolean;
+      loading?: boolean;
```
RyukTheCoder, line 19: shouldn't loading be false?

RyukTheCoder (changes requested):
Great work. I just added some minor comments to be resolved. @samobasquiat

widget/ui/src/components/QuoteExpanded/QuoteExpanded.tsx
```diff
@@ -0,0 +1,182 @@
...
+  SwapperContainer,
+  SwapperImage,
+  TagsContainer,
+  tokenLabelStyles,
+  VerticalLine,
+} from './QuoteExpanded.styles';
+
+const ITEM_SKELETON_COUNT = 3;
+
+export function QuoteExpanded(props: PropTypes) {
```
nikaaru, line 41: I think the name of the component can be better,
For example, something like FullExpandedQuoute

nikaaru (approved):

PR #612, opened 2024-02-24
fix: fix wallet connect bugs in switch network and namespaces

RanGojo:
# Summary

Currently we have problem in initial connection to some wallets as we forced some required methods in the session proposal request. We could just leave required namespace as empty and pass everything we need as the optional namespace. So, it should work even if the wallet doesn't support some specific chains/methods. 

When we submit a session proposal, we receive the final accepted namespace in session response. So we could rely on that afterward. 
For some exceptional wallets e.g. trust wallet and 1inch mobile wallet, they are returning wrong namespace methods in response of the session, so we could just ignore their authorised namespace methods. For example, trust wallet is returning `wallet_switchEthereumChain` in the response but it doesn't support it.

Also for the matter of switch network, as the wallet connect v2 has multi-chain support in a single session, we don't need to recreate the session or switch chain if the chain is already authorised in session response. But in case that wallet supports the switch network as an optional methods, we could use it and rely on that.

I've tested connecting and switch network in multiple popular/requested wallets including:

- [x] Trust Wallet -> test was successful (all optional chains are accepted, so no need to do anything for switch network. Also, wallet returns wrong response for the accepted namespace methods.)
- [x] Metamask -> test was successful (wallet only authorize the current selected chain in session response, but since it supports the optional method for switch network, everything goes well)
- [x] 1inch -> test was successful with the behaviour similar to the trust wallet. Also when passing optional solana or cosmos namespaces, it faces with problem with the initial connection.
- [x] Rainbow -> test was successful (it accepts some limited chains of the optional chains and works fine without switch network as it's not either supported or needed.)
- [x] Safepal -> test was successful (similar to the rainbow)
- [x] Coinbase -> test failed (It seems we need to implement the coin base sdk and it's not possible to initiate a connection to the coinbase through the wallet connect library)
- [x] Kucoin (Halo) -> test failed (I wasn't also able to connect to it in other dapps like 1inch, it gives an error "user rejected". seems to be a bug on their side.)
- [x] Gem -> test succeeded. It's similar to rainbow. but they have a bug in sending transaction to the correct chain; that I've reported it to them. (They have this bug in different dapps e.g. 1inch or ...)

All these wallets were tested only on android. It would be great if the reviewer also test some wallets like Trust wallet and Metamask on IOS as well.

P.S. I've attached the session response for different wallets to help you understand how it works in different wallets

[trust.json](https://github.com/rango-exchange/rango-client/files/14396338/trust.json)
[safepal.json](https://github.com/rango-exchange/rango-client/files/14396337/safepal.json)
[rainbow.json](https://github.com/rango-exchange/rango-client/files/14396336/rainbow.json)
[metamask.json](https://github.com/rango-exchange/rango-client/files/14396335/metamask.json)
[1inch.json](https://github.com/rango-exchange/rango-client/files/14396333/1inch.json)
[gem.json](https://github.com/rango-exchange/rango-client/files/14396334/gem.json)

# Checklist:

### DIFF
--- a/wallets/core/src/wallet.ts
+++ b/wallets/core/src/wallet.ts
@@ -120,6 +120,7 @@ class Wallet<InstanceType = any> {
           network: requestedNetwork,
           newInstance: this.tryGetInstance.bind(this),
           getState: this.getState.bind(this),
+          updateChainId: this.updateChainId.bind(this),
         });
 
         /*
@@ -194,6 +195,21 @@ class Wallet<InstanceType = any> {
       // Typescript can not detect we are filtering out null values:(
       nextAccounts = accounts.filter(Boolean);
       nextNetwork = requestedNetwork || this.options.config.defaultNetwork;
+      /*
+       * For providers like wallet connect with async switch network support,
+       * it's not a correct assumption that current network after connection request
+       * equals to the requested network. (for example in case of session reconnect.)
+       * In this case, we could assume that the first item in connect result equals to
+       * the current chain. Otherwise, the state of wallet-core will be wrong.
+       */
+      if (this.options.config.isAsyncSwitchNetwork) {
+        const nextChainId = connectResult[0].chainId || Networks.Unknown;
+        nextNetwork =
+          getBlockChainNameFromId(
+            nextChainId,
+            this.info.supportedBlockchains
+          ) || Networks.Unknown;
+      }
     } else {
       const chainId = connectResult.chainId || Networks.Unknown;
       const network =
--- a/wallets/provider-walletconnect-2/src/helpers.ts
+++ b/wallets/provider-walletconnect-2/src/helpers.ts
@@ -1,4 +1,3 @@
-import type { CosmosMeta } from './types';
 import type { WalletState } from '@rango-dev/wallets-shared';
 import type { ProposalTypes } from '@walletconnect/types';
 import type { BlockchainMeta } from 'rango-types';
@@ -10,18 +9,12 @@ import {
 } from '@rango-dev/wallets-shared';
 import { WalletConnectModal } from '@walletconnect/modal';
 import { AccountId, ChainId } from 'caip';
-import {
-  cosmosBlockchains,
-  evmBlockchains,
-  isEvmBlockchain,
-} from 'rango-types';
+import { evmBlockchains, isEvmBlockchain } from 'rango-types';
 
 import {
-  DEFAULT_COSMOS_METHODS,
   DEFAULT_ETHEREUM_EVENTS,
   DEFAULT_ETHEREUM_METHODS,
   DEFAULT_SOLANA_CHAIN_ID,
-  DEFAULT_SOLANA_METHODS,
   EthereumRPCMethods,
   NAMESPACES,
 } from './constants';
@@ -47,75 +40,21 @@ type FinalNamespaces = {
   [key in NAMESPACES]?: ProposalTypes.BaseRequiredNamespace;
 };
 
-export function generateRequiredNamespace(
-  meta: BlockchainMeta[],
-  network: string
-): FinalNamespaces | undefined {
-  const evm = evmBlockchains(meta);
-  const cosmos = cosmosBlockchains(meta);
-
-  const requiredEvmChain = evm.find((chain) => chain.name === network);
-  const requiredCosmosChain = cosmos.find((chain) => chain.name === network);
-  const requiredSolanaChain = network === Networks.SOLANA;
-
-  if (requiredEvmChain) {
-    return {
-      [NAMESPACES.ETHEREUM]: {
-        events: DEFAULT_ETHEREUM_EVENTS,
-        methods: DEFAULT_ETHEREUM_METHODS,
-        chains: [
-          new ChainId({
-            namespace: NAMESPACES.ETHEREUM,
-            reference: String(parseInt(requiredEvmChain.chainId)),
-          }).toString(),
-        ],
-      },
-    };
-  } else if (!!requiredCosmosChain) {
-    return {
-      [NAMESPACES.COSMOS]: {
-        events: [],
-        methods: DEFAULT_COSMOS_METHODS,
-        chains: [
-          new ChainId({
-            namespace: NAMESPACES.COSMOS,
-            reference: requiredCosmosChain.chainId!,
-          }).toString(),
-        ],
-      },
-    };
-  } else if (requiredSolanaChain) {
-    return {
-      [NAMESPACES.SOLANA]: {
-        events: [],
-        methods: DEFAULT_SOLANA_METHODS,
-        chains: [`solana:${DEFAULT_SOLANA_CHAIN_ID}`],
-      },
-    };
-  }
-
-  return undefined;
-}
-
+/*
+ * Some wallets like 1inch (android) has problem when we pass cosmos chains in optional namespace
+ * Also some wallets like keplr mobile doesn't work when we don't pass cosmos chains in required namespace
+ * It seems to be a bug in their current implementation.
+ */
 export function generateOptionalNamespace(
   meta: BlockchainMeta[]
 ): FinalNamespaces | undefined {
   const evm = evmBlockchains(meta);
-  const cosmos = cosmosBlockchains(meta);
   const evmChains = evm.map((chain) => {
     return new ChainId({
       namespace: NAMESPACES.ETHEREUM,
       reference: String(parseInt(chain.chainId)),
     }).toString();
   });
-  const cosmosChains = cosmos
-    .filter((chain): chain is CosmosMeta => !!chain.chainId)
-    .map((chain) => {
-      return new ChainId({
-        namespace: NAMESPACES.COSMOS,
-        reference: chain.chainId,
-      }).toString();
-    });
 
   const namespaces: FinalNamespaces = {
     [NAMESPACES.ETHEREUM]: {
@@ -123,21 +62,7 @@ export function generateOptionalNamespace(
       events: DEFAULT_ETHEREUM_EVENTS,
       chains: evmChains,
     },
-    [NAMESPACES.SOLANA]: {
-      methods: DEFAULT_SOLANA_METHODS,
-      events: [],
-      chains: [`solana:${DEFAULT_SOLANA_CHAIN_ID}`],
-    },
   };
-
-  if (cosmosChains.length) {
-    namespaces[NAMESPACES.COSMOS] = {
-      methods: DEFAULT_COSMOS_METHODS,
-      events: [],
-      chains: cosmosChains,
-    };
-  }
-
   return namespaces;
 }
 
@@ -281,4 +206,35 @@ export function getEvmAccount(
     },
     address,
   });
+}
+
+// It's enough to return only connected network for the EVM networks and ignore others
+export function filterEvmAccounts(
+  accounts: {
+    address: string;
+    chainId: string;
+  }[],
+  currentChainId?: string
+) {
+  let firstEvmAddress = false;
+  return accounts
+    .filter(({ address, chainId }) => {
+      const isEvm = isEvmAddress(address);
+      if (isEvm) {
+        if (currentChainId && chainId !== currentChainId) {
+          return false;
+        }
+        if (!currentChainId) {
+          if (firstEvmAddress) {
+            return false;
+          }
+          firstEvmAddress = true;
+        }
+      }
+      return true;
+    })
+    .map(({ address, chainId }) => ({
+      accounts: [address],
+      chainId: chainId,
+    }));
 }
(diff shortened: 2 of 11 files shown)


wallets/provider-walletconnect-2/src/helpers.ts
```diff
@@ -47,58 +47,19 @@ type FinalNamespaces = {
...
+/*
+ * Since we didn't enable non-evm chains yet,
+ * it's better to not pass them in session proppsal request.
+ * As some wallets like 1inch wallet face with problem with passing non-evm chains
+ * even in optional name spaces.
+ */
 export function generateOptionalNamespace(
-  meta: BlockchainMeta[]
+  meta: BlockchainMeta[],
+  onlyEvm = true
```
yeager-eren, line 62: I couldn't find anywhere that passing `onlyEvm` as `false`. If we don't have any logic for that, we can remove these parts from this function and only keeps EVM.

yeager-eren: By the way, I could connect 1inch without facing any problem. is there any specific scenario you've encountered?

wallets/provider-walletconnect-2/src/index.ts
```diff
@@ -106,23 +107,9 @@ export const connect: Connect = async ({ instance, network, meta }) => {
...
-  if (!isNew) {
-    return {
-      chainId: String(currentChainId),
-      accounts: accounts?.[0].accounts,
-    };
-  }
-  const newChainId = accounts?.[accounts.length - 1].chainId;
-  void persistCurrentChainId(instance, newChainId);
-  return accounts?.[accounts.length - 1];
+  let currentChainId = !isNew ? await getPersistedChainId(instance) : undefined;
```
yeager-eren, line 110: `isNew` is using here only, couldn't we call `await getPersistedChainId(instance)` anyway and fallback to `false` whenever something goes wrong?
If yes, we can remove whole `isNew` stuff to simplify the code.

wallets/provider-walletconnect-2/src/index.ts
```diff
@@ -176,26 +163,33 @@ export const switchNetwork: SwitchNetwork = async ({
...
-     * we need to handle it manually by deleting last session and create a new one
-     * with correct chain id.
-     */
-    const session = await trySwitchByCreatingNewSession(instance, {
-      network,
-      meta,
-    });
-    instance.session = session;
+  const evm = evmBlockchains(meta);
+  const chainId = evm.find((chain) => chain.name === network)?.chainId || '0x1';
```
yeager-eren, line 169: Why we are falling back to `0x1` when we couldn't match anything? it shouldn't be throwing an error?

wallets/provider-walletconnect-2/src/index.ts
```diff
@@ -176,26 +163,33 @@ export const switchNetwork: SwitchNetwork = async ({
...
+    !ignoreNamespaceMethods(instance)
+  ) {
+    const currentNetwork = getState?.().network || Networks.ETHEREUM;
+    await updateSessionAccounts(instance, network, currentNetwork, meta);
+    await switchOrAddEvmChain(meta, network, currentNetwork, instance);
+  } else if (authorizedChains.includes(chaindIdStr)) {
+    updateChainId(chainId);
     return;
+  } else {
+    throw Error(`Chain ${chainId} is not configured.`);
```
yeager-eren, line 191: you can use our new logging packages to report the error. `import {error} from "@rango-dev/logging-core"`

wallets/provider-walletconnect-2/src/session.ts
```diff
@@ -87,7 +86,10 @@ export async function createSession(
 
       onCloseModal = new Promise((_, reject) => {
         modal.subscribeModal((state) => {
-          // the modal was closed so reject the promise
+          /*
+           * the modal was closed so reject the promise
+           * TODO CHECK REJECT UNHANDLED ERROR
```
yeager-eren, line 91: What do you mean by this? Avoiding sending modal rejection to sentry?

wallets/provider-walletconnect-2/src/session.ts
```diff
@@ -153,9 +152,9 @@ export async function tryConnect(
   const pairing = tryGetPairing(client);
   if (pairing) {
     try {
-      session = await restoreSession(client, pairing);
+      session = await restoreSession(client, pairing.topic);
     } catch (e) {
-      await disconnectSessions(client);
+      await disconnectSessions(client).catch((error) => console.log(error));
```
yeager-eren, line 157: This can be problematic. why you are ignoring the error? it wouldn't reject `tryConnect` promise if fails, which means it continues to the next lines. Is there any reason?

wallets/provider-walletconnect-2/src/session.ts
```diff
@@ -294,25 +240,24 @@ export function getAccountsFromSession(
        * It will return the chain id itslef if it's not that specific ID.
        */
       const chain = solanaChainIdToNetworkName(chainId.reference);
+      const isEvm = isEvmAddress(address);
+      if (isEvm) {
```
yeager-eren, line 244: Let keep this logic out of this function. `getAccountsFromSession` should return all the accounts from session, if you need only one evm account, you can make a separate function and filter the result whenever you need. (not in this function)

yeager-eren (commented):
Thanks. I put some comments. I have one question which may not related to you PR. Is there any reason we are only showing WalletConnect for EVMs? you can check `getWalletInfo`.

wallets/core/src/wallet.ts
```diff
@@ -120,6 +120,7 @@ class Wallet<InstanceType = any> {
           network: requestedNetwork,
           newInstance: this.tryGetInstance.bind(this),
           getState: this.getState.bind(this),
+          updateChainId: this.updateChainId.bind(this),
```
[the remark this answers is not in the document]
RanGojo: It's not as you think. As we don't do anything on wallet side when chain is already authorized, no event will be called.

wallets/provider-walletconnect-2/src/helpers.ts
RanGojo: Yes. if you pass optional cosmos / solana chain in namespace, it will fail.

wallets/provider-walletconnect-2/src/index.ts
RanGojo: I'm not sure what you mean but we need to only use persisted chain id in case of reconnect to set correct state for our wallet core.

wallets/provider-walletconnect-2/src/session.ts
RanGojo: I will remove the TODO. I think we have a bug in embedded in handling connect errors. (ignore first 25 seconds of the video)


https://github.com/rango-exchange/rango-client/assets/99180934/6245663f-b05f-42a8-94c7-511c18d1603f

wallets/core/src/wallet.ts
```diff
@@ -194,6 +195,14 @@ class Wallet<InstanceType = any> {
       // Typescript can not detect we are filtering out null values:(
       nextAccounts = accounts.filter(Boolean);
       nextNetwork = requestedNetwork || this.options.config.defaultNetwork;
+      if (this.options.config.isAsyncSwitchNetwork) {
```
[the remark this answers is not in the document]
RanGojo: I've added some comments in code.

wallets/provider-walletconnect-2/src/session.ts
RanGojo: Yes, you are right. Adding this catch in disconnect was enough.

RanGojo:
> Thanks. I put some comments. I have one question which may not related to you PR. Is there any reason we are only showing WalletConnect for EVMs? you can check `getWalletInfo`.

It should be logically possible and it works fine for trust wallet in terms of getting cosmos accounts. (I didn't check sign transaction again) But for keplr mobile, I faced an issue that they require to pass cosmos chains as required namespace which doesn't make sense. We could investigate more in another PR.

wallets/provider-walletconnect-2/src/helpers.ts
yeager-eren: Thanks. much better. Could you please add this comment on top of `generateOptionalNamespace` as well? to ensure we don't add it in future or if we did, we should test android.

wallets/provider-walletconnect-2/src/index.ts
yeager-eren: Thanks. Could you also send it to our `logging`? I mean use `logError` as well.

wallets/provider-walletconnect-2/src/index.ts
yeager-eren: Please use `new` before `Error`.

wallets/provider-walletconnect-2/src/session.ts
yeager-eren: So please make sure we created an issue on our backlog.

wallets/core/src/wallet.ts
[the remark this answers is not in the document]
yeager-eren: My bad.  I thought you didn't use `updateChainId` inside `switchNetwork`, I checked it again and i was missing your usage. Sorry.

wallets/provider-walletconnect-2/src/index.ts
yeager-eren: I mean if failing `getPersistedChainId` should reject the whole `connect` (it only gets chainId from storage), we can do something like this:

session.ts
```ts
export async function getPersistedChainId(instance: any) {
  try {
    return (await instance.client.core.storage.getItem(CHAIN_ID_STORAGE))
      ?.defaultChainId;
  } catch {
    return undefined;
  }
}
```

index.ts
```ts
let currentChainId = await getPersistedChainId(instance);
```

and then `isNew` logic codes since the variable is not used anymore.

yeager-eren (changes requested):

wallets/provider-walletconnect-2/src/index.ts
```diff
@@ -211,7 +209,7 @@ export const disconnect: Disconnect = async ({ instance }) => {
   const { client } = instance as WCInstance;
 
   if (client) {
-    void disconnectSessions(client);
+    void disconnectSessions(client).catch((error) => console.log(error));
```
yeager-eren, line 212: I guess you should remove `.catch` as well.

wallets/provider-walletconnect-2/src/index.ts
RanGojo: I think it's not correct because we might have some stale values for default chain id in storage (from previous connections) which shouldn't be used in case of a new connection.

wallets/provider-walletconnect-2/src/index.ts
RanGojo: I think it's not correct. Because the wallet is disconnected but we sometimes raise errors in disconnect because of try to disconnect all previous pairings session s in disconnect flow which some of them don't exist and it's not logical to raise error for them.

yeager-eren: That's fair. Could you please use `logError` instead?

wallets/provider-walletconnect-2/src/index.ts
yeager-eren: Here is how I see the flow:

`tryConnect` will try to restore the session, if it couldn't (stale) it will call `disconnectSessions` which remove old values.
if there is no session, it will create a new session and set `isNew = true`.

afterwards, we have this code:
```
  let currentChainId = !isNew ? await getPersistedChainId(instance) : undefined;
  currentChainId = currentChainId ? String(currentChainId) : undefined;
```
which means if a session restored, try get the chain from session. (it not possible to session be staled here, as explained before).
So we can safely read from storage and if the value is not there it will return `undefined`

wallets/provider-walletconnect-2/src/helpers.ts
```diff
@@ -282,3 +207,34 @@ export function getEvmAccount(
     address,
   });
 }
+
+// It's enough to return only connected network for the EVM networks and ignore others
+export function filterEvmAccounts(
```
yeager-eren, line 212: Maybe i'm wrong and I didn't test, but does this works too?

```
export function findFirstEvmAccount(accounts, currentChainId) {
  const account = accounts.find((account) => {
    const { address, chainId } = account;
    return isEvmAddress(address) && chainId == currentChainId;
  });

  return account;
}
```

wallets/provider-walletconnect-2/src/index.ts
RanGojo: Why we should report this to sentry? I didn't understand.

wallets/provider-walletconnect-2/src/helpers.ts
RanGojo: It's not the same logic. Firstly, I don't want to filter non-evm accounts but this code only returns evm account. Secondly, if the current chain id is undefined, I return the first evm account and exclude other evm accounts.
But I update it to make it more readable.

wallets/provider-walletconnect-2/src/index.ts
RanGojo: thanks for the explanation. I've changed the code now.

yeager-eren (approved):
LGTM. Thanks.

PR #621, opened 2024-03-03
fix: fix full expanded design issues

nikaaru:
# Summary

- update icon, logo, and text size 
- Fix the height of the layout header on all pages
- Prevent swaps for identical tokens
- Update the condition of showing expand and full-expand mode

You can test the expand and full expand modes by changing the **variant** in the **config**

# Checklist:


widget/embedded/src/components/HeaderButtons/RefreshButton.tsx
```diff
@@ -63,6 +63,7 @@ function RefreshButton({ onClick }: { onClick: (() => void) | undefined }) {
     <HeaderButton
       variant="ghost"
       size="small"
+      style={{ paddingTop: 0, paddingBottom: 0 }}
```
yeager-eren, line 66: That would be great if you could handle it in `HeaderButton` itself instead of writing inline styles.
I took a look at `HeaderButton` it seems it has `padding: 0` already, but for CSS specificity it doesn't apply. There are 2 solutions for fixing that inside the `HeaderButton` itslef:

1. Adding `!important` to `padding: 0`.
2. Use this trick: `  '&&': {
    padding: '0',
  },`

I'm more agree with the second approach, it's kinda weird but better than `important!` I believe.

yeager-eren (changes requested):

widget/embedded/src/components/HeaderButtons/RefreshButton.tsx
yeager-eren: as we talked, it can affect other places, so I resolve it for now.

yeager-eren (approved):
Thanks. LGTM.

Ikari-Shinji-re (approved):

PR #630, opened 2024-03-06
fix: fix queue manager context in widget embedded

Ikari-Shinji-re:
# Summary

Some parameters currently being passed to the queue-manager context, such as "disconnectedWallet," are incorrect. This leads to the queue-manager being unable to detect the disconnection of the required wallet. Additionally, the logic related to these states is scattered and challenging to follow.

# Checklist:


widget/embedded/src/hooks/useBootstrap/useBootstrap.ts
```diff
@@ -0,0 +1,62 @@
...
+  });
+
+  useEffect(() => {
+    tabManager.init();
+
+    if (!useNotificationStore.persist.hasHydrated()) {
+      void useNotificationStore.persist.rehydrate();
+    }
+
+    widgetContext.onConnectWallet(setLastConnectedWalletWithNetwork);
```
samobasquiat, line 48: Could you write a comment above these lines to clarify?

samobasquiat (approved):

widget/embedded/src/hooks/useBootstrap/useBootstrap.ts
```diff
@@ -0,0 +1,67 @@
...
+    if (!useNotificationStore.persist.hasHydrated()) {
+      void useNotificationStore.persist.rehydrate();
+    }
+
+    /**
+     * Add handlers to the widget context for connecting and disconnecting wallets.
+     * We pass these handlers to the wallet provider to keep track of the last connected and last disconnected wallets.
+     * The 'queue-manager' uses these details to spot any changes in the connection status of the required wallet during the swap process.
+     */
+    widgetContext.onConnectWallet(setLastConnectedWalletWithNetwork);
```
yeager-eren, line 53: I'm concerned about the current approach. We've placed `onConnectWallet` and `onDisconnectWallet` within `WidgetContext`, making them accessible everywhere. My concern is if someone else attempts to do the same (`widgetContext.onDisconnectWallet(  somethingElseFunction )`), it could lead to conflicts. I guess the last one defined and called would override the context so it can not be reliable and potentially unsafe .
If my understanding is accurate, How we can guard these handlers safely to not gone?

yeager-eren (changes requested):
Thanks. much cleaner now. 
I asked a question only and to make sure I underestand your changes well, the problem was we didn't have `onDisconnectWalletHandler` so we wouldn't update `disconnectedWallet` at all. This PR fixed this issue and other stuff was only refactoring, Am I right?

widget/embedded/src/hooks/useBootstrap/useBootstrap.ts
yeager-eren: as we talked offline, we decided to handle this somehow in future,

yeager-eren (approved):
Thanks.

PR #632, opened 2024-03-09
feat: add enabling centralized swappers to config

samobasquiat:
# Summary

Add enabling centralized swappers to config to pass to get meta & get route api params.

Fixes # (issue)

by passing `enableCentralizedSwappers` to config & checking the body of meta & qoute api call params

# Checklist:

### DIFF
--- a/widget/embedded/src/store/slices/data.ts
+++ b/widget/embedded/src/store/slices/data.ts
@@ -207,7 +207,10 @@ export const createDataSlice: StateCreator<
   // Actions
   fetch: async () => {
     try {
-      const response = await sdk().getAllMetadata();
+      const { enableCentralizedSwappers } = get().config;
+      const response = await sdk().getAllMetadata({
+        enableCentralizedSwappers,
+      });
 
       set({ fetchStatus: 'success' });
       const blockchains: BlockchainMeta[] = [];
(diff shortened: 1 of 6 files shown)


widget/embedded/src/types/config.ts
```diff
@@ -185,6 +185,8 @@ export type Features = Partial<
  *   If it is expanded, multiple routes will show up on the home page;
  *   If it is full-expanded, multiple routes will show up on the home page with full routes;
  *   if not, you will need to go to a different page to see the suggested routes.
+ * @property {boolean} enableCentralizedSwappers
+ * If you want to enable routing from the centralized protocols like Exodus, you could set this parameter to true.
```
RanGojo, line 189:
```suggestion
 * If you want to enable routing from the centralized protocols like XO Swap, you could set this parameter to true.
```

widget/embedded/src/store/slices/data.ts
```diff
@@ -207,7 +207,10 @@ export const createDataSlice: StateCreator<
   // Actions
   fetch: async () => {
     try {
-      const response = await sdk().getAllMetadata();
+      const { enableCentralizedSwappers } = get().config;
+      const response = await sdk().getAllMetadata({
+        enableCentralizedSwappers,
+      });
```
RanGojo, line 213: I see that we passed this parameter correctly to meta endpoint. ‌But I can't see 'XO Swap' in widget setting. Maybe we have a ui filter there on swapper type.

RanGojo (changes requested):

RanGojo (approved):

PR #633, opened 2024-03-09
feat: add more languages to widget

mikasackermn:
# Summary

add more languages to widget

Fixes # (issue)

- [x] A warning on top of language list to let user know we are using machine translations.
- [x] chinese (Mandarin)
- [x] Russian
- [x] German
- [x] Vietnamese
- [x] Swedish
- [x] Finnish
- [x] Dutch
- [x] Greek
- [x] Italian
- [x] Polish

Note: Not supported
Marathi, Telugu, Tamil, Nigerian Pidgin, Hindi, Bengali, Indonesian, Turkish, Vietnamese, Malay, Thai

# Checklist:

### DIFF
--- a/lingui.config.ts
+++ b/lingui.config.ts
@@ -1,6 +1,6 @@
 /** @type {import('@lingui/conf').LinguiConfig} */
 module.exports = {
-  locales: ['en', 'es', 'ja', 'fr', 'pt'],
+  locales: ['en', 'es', 'ja', 'fr', 'pt', 'zh', 'ru', 'de', 'uk', 'sv', 'fi', 'nl', 'el', 'it', 'pl'],
   sourceLocale: 'en',
   format: 'po',
   catalogs: [
--- a/translations/de.po
+++ b/translations/de.po
@@ -1,22 +1,17 @@
 msgid ""
 msgstr ""
-"POT-Creation-Date: 2023-11-06 17:24+0330\n"
+"POT-Creation-Date: 2024-03-10 11:57+0330\n"
 "MIME-Version: 1.0\n"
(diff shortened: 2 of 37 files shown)


lingui.config.ts
```diff
@@ -1,6 +1,6 @@
 /** @type {import('@lingui/conf').LinguiConfig} */
 module.exports = {
-  locales: ['en', 'es', 'ja', 'fr', 'pt'],
+  locales: ['en', 'es', 'ja', 'fr', 'pt', 'zh-CN', 'ru', 'de', 'uk', 'sv-SE', 'fi', 'nl', 'el', 'it', 'pl'],
```
nikaaru, line 3: I think we don't need the subset of languages right now
We can use the main set for now
I tested it on my repo and it was correct.

fix: 

zh-CN  ---> zh
sv-SE  ----> sv

Subsequently, the names of all components and files should be changed.

examples/wallets-demo/package.json
```diff
@@ -12,7 +12,7 @@
   },
   "dependencies": {
     "@rango-dev/provider-all": "^0.28.1-next.2",
-    "@rango-dev/ui": "^0.29.1-next.13",
+    "@rango-dev/ui": "^0.29.1-next.14",
```
nikaaru, line 15: As a rule, the package version should not be changed.
I think that if you pull the **next** branch, it will be fixed

mikasackermn: fixed

lingui.config.ts
mikasackermn: fixed

nikaaru (approved):
Thanks, LGTM

PR #635, opened 2024-03-10
fix: apply minor improvements to the widget

Ikari-Shinji-re:
# Summary

some minor improvements on the widget

Fixes # (issue)

- [x] The height of the quote skeleton doesn't align with the quote component, resulting in a minor layout shift 
image1
- [x] Display sorting strategy for quotes upon loading 
image2
- [x] In expanded mode, the swap input component's bottom corners should have rounded edges
image3
- [x] In expanded mode, the quote modals overflow due to the smaller height of the swap-box
image4
- [x] Remove the time restriction for the "try again" functionality

# Checklist:

### DIFF
--- a/widget/embedded/src/components/Quotes/Quotes.tsx
+++ b/widget/embedded/src/components/Quotes/Quotes.tsx
@@ -67,6 +67,18 @@ export function Quotes(props: PropTypes) {
 
   return (
     <>
+      {hasSort && (
+        <>
+          <StrategyContent>
+            <Typography size="xmedium" variant="title">
+              {i18n.t('Sort by')}
+            </Typography>
+            <SelectStrategy container={getContainer()} />
+          </StrategyContent>
+          <Divider size="10" />
+        </>
+      )}
+
       {loading &&
         Array.from({ length: ITEM_SKELETON_COUNT }, (_, index) => (
           <React.Fragment key={index}>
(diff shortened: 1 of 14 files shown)


widget/embedded/src/components/Quotes/Quotes.tsx
```diff
@@ -67,6 +67,18 @@ export function Quotes(props: PropTypes) {
 
   return (
     <>
+      {hasSort && (
```
samobasquiat, line 70: After this change in the loading state, we have a scroll
image

widget/embedded/src/containers/Inputs/Inputs.tsx
```diff
@@ -116,7 +120,9 @@ export function Inputs(props: PropTypes) {
         <SwitchFromAndToButton />
       </FromContainer>
       <SwapInput
-        sharpBottomStyle={!!selectedQuote || fetchingQuote}
+        sharpBottomStyle={
```
samobasquiat, line 123: It should also be in `expanded` mode.

samobasquiat:
I think this alert needs an info icon as well. 
image

samobasquiat (changes requested):

widget/embedded/src/components/Quotes/Quotes.tsx
Ikari-Shinji-re: The UI design has a scrollbar, and in my recent widget review, I proposed getting rid of it from all skeletons. Let's go through the pages and components to figure out the best way to handle this, but keep in mind that this decision should be separate from the current task.

samobasquiat: I prefer not to have a scroll in skeletons but let's allow to second reviewer to check it.

samobasquiat (approved):

RanGojo (approved):

PR #639, opened 2024-03-11
feat: attach config passed to widget to window

mikasackermn:
# Summary

attach `config` passed to widget to window

e.g.:
window.__rango.config

The config is available in the console.

# Checklist:

### DIFF
--- a/widget/embedded/global-env.d.ts
+++ b/widget/embedded/global-env.d.ts
@@ -1,0 +1,12 @@
+import type { WidgetConfig } from './src';
+
+export {};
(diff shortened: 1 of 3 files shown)


yeager-eren:
For debugging purposes we need to have a full representation of config, not only the part passed by dApp. It's possible to change default configs during releases for us, so we need to know what config user is using. I think that would be good to keep these two (dApp input and default configs) separated.

yeager-eren (approved):
thanks. lgtm.

PR #654, opened 2024-03-12
!WIP feat: add bitcoin connect and signer to phantom wallet

RyukTheCoder:
# Summary

Added connect Bitcoin to Phantom wallet. In order to implement bitcoin transaction Signer, it is needed to receive Unspent Transaction Outputs (UTXOs) associated with user accounts. Therefore, please note that this PR is currently marked as a draft, as further changes to the APIs are required to complete the implementation.

#Discussion

- [ ] We should consider what type of transaction we can support (SegWit, Non SegWit)
- [ ] Phantom should add support for memo in PSBTs
- [ ] To support Non SegWit UTXOs we should provide UTXO hex some way

### DIFF
--- a/wallets/core/src/types.ts
+++ b/wallets/core/src/types.ts
@@ -4,7 +4,11 @@ import type {
   WalletInfo,
   WalletType,
 } from '@rango-dev/wallets-shared';
-import type { BlockchainMeta, SignerFactory } from 'rango-types';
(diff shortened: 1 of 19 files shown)


RyukTheCoder:
Changes is done in another PR:
https://github.com/rango-exchange/rango-client/pull/972

PR #659, opened 2024-03-16
feat: implement hover state for full-expanded-route

samobasquiat:
# Summary

Added a hover state for full-expanded routes with showing fee & time, errors and internal-swaps for each step.

change the variant to full-expanded and hover on swapper icons.

# Checklist:


nikaaru:
image

The output amount is outside the hover container

nikaaru:
image

The swapper logo should be slightly above the hover container

nikaaru:
image

According to Figma, the background color of the tooltip should be this color

nikaaru:
image

According to Figma, the area where the tooltip should be triggered should be less

nikaaru:
image

The padding and box-shadow around the container during hover are wrong

nikaaru:
image

I think that for the case where the internal swap has more than one swapper, we should show it like this

nikaaru (approved):

mikasackermn (approved):

PR #662, opened 2024-03-18
fix: fix widget event hook

Ikari-Shinji-re:
# Summary

The `useWidgetEvents` hook doesn’t work when used outside of the widget.

Fixes # (issue)

We can opt to eliminate the `all` method from the `eventEmitter` and rely on the `off` method to detach event listeners instead.

You have the option to utilize the `useWidgetEvent` hook within the 'widget-app' package to subscribe to a particular event.

# Checklist:


widget/embedded/src/hooks/useBootstrap/useBootstrap.ts
```diff
@@ -69,4 +87,50 @@ export function useBootstrap() {
 
     return tabManager.destroy;
   }, []);
+
+  useEffect(() => {
```
samobasquiat, line 91: I'm not sure if deleting WidgetEvents and moving it to useBootstrap is a good idea.  The previous version was capsulated and did just one thing.

queue-manager/rango-preset/src/types.ts
```diff
@@ -278,7 +278,11 @@ export type RouteEvent =
   | RouteSucceededEvent
   | RouteFailedEvent;
 
+//  I think we should pick better names, but I don’t have any ideas at the moment.
```
samobasquiat, line 281: Please remove this comment.

samobasquiat (changes requested):

PR #666, opened 2024-03-25
feat: add ledger connect and signer

RyukTheCoder:
# Summary

Added Ledger connect and signer.

Fixes # (issue)

- [x] Connect Ledger device and open Ethereum app on device
- [x] Click to connect to Ledger on widget and allow to connect
- [x] Create an Ethereum transaction and sign using Ledger device

# Checklist:

### DIFF
--- a/wallets/provider-ledger/src/helpers.ts
+++ b/wallets/provider-ledger/src/helpers.ts
@@ -1,0 +1,71 @@
+import type Transport from '@ledgerhq/hw-transport';
+
+const ETHEREUM_CHAIN_ID = '0x1';
+
+export const ETH_BIP32_PATH = "44'/60'/0'/0/0";
+
+const ledgerErrorMessages: { [statusCode: number | string]: string } = {
+  21781: 'The device is locked',
+  25871: 'Related application is not ready on your device',
+  27013: 'Action denied by user',
+  INSUFFICIENT_FUNDS: 'Insufficient funds for transaction',
+};
+
+export function getLedgerInstance() {
+  /*
+   * Instances have a required property which is `chainId` and is using in swap execution.
+   * Here we are setting it as Ethereum always since we are supporting only eth for now.
+   */
+  return { chainId: ETHEREUM_CHAIN_ID };
+}
+
+export async function getLedgerAccounts(): Promise<{
+  accounts: string[];
+  chainId: string;
+}> {
+  try {
+    const transport = await transportConnect();
+
+    const eth = new (await import('@ledgerhq/hw-app-eth')).default(transport);
+
+    const accounts: string[] = [];
+
+    const result = await eth.getAddress(ETH_BIP32_PATH, false, true);
+    accounts.push(result.address);
+
+    return {
+      accounts: accounts,
+      chainId: ETHEREUM_CHAIN_ID,
+    };
+  } catch (error: any) {
+    throw getLedgerError(error);
+  } finally {
+    await transportDisconnect();
+  }
+}
+
+export function getLedgerError(error: any) {
+  const errorCode = error?.statusCode || error?.code; // ledger error || broadcast error
+
+  if (errorCode && !!ledgerErrorMessages[errorCode]) {
+    return new Error(ledgerErrorMessages[errorCode]);
+  }
+  return error;
+}
+
+let transportConnection: Transport | null = null;
+
+export async function transportConnect() {
+  transportConnection = await (
+    await import('@ledgerhq/hw-transport-webhid')
+  ).default.create();
+
+  return transportConnection;
+}
+
+export async function transportDisconnect() {
+  if (transportConnection) {
+    await transportConnection.close();
+    transportConnection = null;
+  }
+}
--- a/wallets/provider-ledger/src/index.ts
+++ b/wallets/provider-ledger/src/index.ts
@@ -1,0 +1,52 @@
+import type {
+  Connect,
+  Disconnect,
+  WalletInfo,
+} from '@rango-dev/wallets-shared';
+import type { BlockchainMeta, SignerFactory } from 'rango-types';
+
+import { Networks, WalletTypes } from '@rango-dev/wallets-shared';
+
+import {
+  getLedgerAccounts,
+  getLedgerInstance,
+  transportDisconnect,
+} from './helpers';
+import signer from './signer';
+
+const WALLET = WalletTypes.LEDGER;
+
+export const config = {
+  type: WALLET,
+};
+
+export const getInstance = getLedgerInstance;
+export const connect: Connect = async () => {
+  const ledgerAccounts = await getLedgerAccounts();
+
+  return ledgerAccounts;
+};
+
+export const disconnect: Disconnect = async () => {
+  void transportDisconnect();
+};
+
+export const getSigners: (provider: any) => SignerFactory = signer;
+
+export const getWalletInfo: (allBlockChains: BlockchainMeta[]) => WalletInfo = (
+  allBlockChains
+) => {
+  const ethereumBlockchain = allBlockChains.find(
+    (chain) => chain.name === Networks.ETHEREUM
+  );
+  return {
+    name: 'Ledger',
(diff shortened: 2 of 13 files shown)


RyukTheCoder:
Tested Transactions Request IDs:

aef6d4c7-41b9-49d0-885f-eb4d6db7e729

123f80cc-0e2b-4208-9526-e3ff118a5e5f

94a32e27-de98-4d69-8466-6e44ec6f6a0e

98fc30b8-baa3-434d-9f9f-fa77f39391a3

wallets/provider-ledger/src/helpers.ts
```diff
@@ -0,0 +1,54 @@
...
+
+const ledgerErrorMessages: { [statusCode: number | string]: string } = {
+  21781: 'The device is locked',
+  25871: 'Related application is not ready on your device',
+  27013: 'Action denied by user',
+  INSUFFICIENT_FUNDS: 'Insufficient funds for transaction',
+};
+
+export function getLedgerInstance() {
+  return { chainId: ETHEREUM_CHAIN_ID };
```
yeager-eren, line 16: Please add a comment to describe why it's like this here. (We've talked offline, writing it down for documenting purposes)

wallets/provider-ledger/src/helpers.ts
```diff
@@ -0,0 +1,54 @@
...
+
+    const result = await eth.getAddress(ETH_BIP32_PATH, false, true);
+    accounts.push(result.address);
+
+    return {
+      accounts: accounts,
+      chainId: ETHEREUM_CHAIN_ID,
+    };
+  } catch (error: any) {
+    throw new Error(getLedgerErrorMessage(error));
```
yeager-eren, line 39: I think you should change it to `throw getLedgerErrorMessage(error)`. 
And also change `getLedgerErrorMessage` function to something like this:
```js
if (errorStatusCode && !!ledgerErrorMessages[errorStatusCode]) {
    throw new Error(ledgerErrorMessages[errorStatusCode]);
  }
  throw error;
```

wallets/provider-ledger/src/index.ts
```diff
@@ -0,0 +1,48 @@
...
+
+export const getInstance = getLedgerInstance;
+export const connect: Connect = async () => {
+  const ledgerAccounts = await getLedgerAccounts();
+
+  return ledgerAccounts;
+};
+
+export const disconnect: Disconnect = async ({ instance }) => {
+  instance?.transport?.close?.();
```
yeager-eren, line 27: in your `getInstance` you are returning this: `return { chainId: ETHEREUM_CHAIN_ID };`
I guess it never has `transport` on it, if it's correct so this line, don't do anything. you can remove it.

wallets/provider-ledger/src/index.ts
```diff
@@ -0,0 +1,48 @@
...
+    (chain) => chain.name === Networks.ETHEREUM
+  );
+  return {
+    name: 'Ledger',
+    img: 'https://raw.githubusercontent.com/rango-exchange/assets/main/wallets/ledger/icon.svg',
+    installLink: {
+      DEFAULT:
+        'https://support.ledger.com/hc/en-us/articles/4404389606417-Download-and-install-Ledger-Live?docs=true',
+    },
+    color: '#dac7ae',
```
yeager-eren, line 45: Is it a correct color?

wallets/provider-ledger/src/signers/ethereumSigner/ledgerEthereumSigner.ts
yeager-eren, line 1: I suggest change `signers/ethereumSigner/ledgerEthereumSigner.ts` to `singers/ethereum.ts`

wallets/provider-ledger/src/signers/ethereumSigner/ledgerEthereumSigner.ts
```diff
@@ -0,0 +1,81 @@
...
+import AppEth, { ledgerService } from '@ledgerhq/hw-app-eth';
+import TransportU2F from '@ledgerhq/hw-transport-webhid';
+import { JsonRpcProvider, Transaction } from 'ethers';
+import { SignerError } from 'rango-types';
+
+import { ETH_BIP32_PATH, getLedgerErrorMessage } from '../../helpers';
+
+export const RPC_PROVIDER_URL = 'https://rpc.ankr.com/eth';
+
+export class LedgerEthereumSigner implements GenericSigner<EvmTransaction> {
```
yeager-eren, line 14:
```suggestion
export class EthereumSigner implements GenericSigner<EvmTransaction> {
```

wallets/provider-ledger/src/signers/ethereumSigner/ledgerEthereumSigner.ts
```diff
@@ -0,0 +1,81 @@
...
+        data: tx.data,
+        value: tx.value,
+        maxPriorityFeePerGas: tx.maxPriorityFeePerGas,
+        maxFeePerGas: tx.maxFeePerGas,
+      };
+
+      const unsignedTx =
+        Transaction.from(transaction).unsignedSerialized.substring(2); // Create unsigned transaction
+
+      const resolution = await ledgerService.resolveTransaction(
```
yeager-eren, line 46: Regarding this document: https://developers.ledger.com/docs/connectivity/ledgerJS/clear-signing

Should we pass second and third params as well? we are passing empty object.

This maybe helpful as well:
https://github.com/LedgerHQ/ledger-live/blob/develop/libs/ledgerjs/packages/hw-app-eth/tests/ledgerService.unit.test.ts

yeager-eren (changes requested):

wallets/provider-ledger/src/index.ts
RyukTheCoder: removed it.

wallets/provider-ledger/src/helpers.ts
RyukTheCoder: Improved it. Thanks

wallets/provider-ledger/src/signers/ethereumSigner/ledgerEthereumSigner.ts
RyukTheCoder: Those two objects are empty because all the properties of those objects are optional but they are not optional themselves. I checked those properties and didn't find anything that we need to pass. They include properties like nftExplorerBaseURL (Backend service responsible for signed NFT APDUS), pluginBaseURL, extraPlugins (extra plugins to add for the resolution), cryptoassetsBaseURL., domains (List of trusted names to clear sign)

wallets/provider-ledger/src/helpers.ts
```diff
@@ -0,0 +1,54 @@
...
+
+const ledgerErrorMessages: { [statusCode: number | string]: string } = {
+  21781: 'The device is locked',
+  25871: 'Related application is not ready on your device',
+  27013: 'Action denied by user',
+  INSUFFICIENT_FUNDS: 'Insufficient funds for transaction',
+};
+
+export function getLedgerInstance() {
+  // chainId put in instance because it is needed to be able to get chainId from provider
```
yeager-eren, line 13:
```suggestion
  // Instances have a required property which is `chainId` and is using in swap execution. Here we are setting it as Ethereum always since we are supporting only eth for now.  
```

wallets/provider-ledger/src/signers/ethereumSigner/ledgerEthereumSigner.ts
yeager-eren: How about `erc20` option? what it does?

RyukTheCoder: ResolutionConfig Allows to configure precisely what the service need to resolve. It seems we can set erc20:true if we need clear signing on ERC20 transaction. If we set it and it is not an ERC20 transaction, it should still work but will do a useless service resolution. In most of examples I see that every parameter is set to true like this: 

`const resolutionConfig: ResolutionConfig = {
  nft: true,
  erc20: true,
  externalPlugins: true,
`};``

PR #676, opened 2024-04-14
fix: address xdefi prioritization issue

samobasquiat:
# Summary

- [x] Fix `xdefi` override issue (even prioritization doesn't work)
- [x] Fix coinbase detection code
- [x] Fix EVM wallet conflicts
- [x] Handle initial state for the network in wallet-core for multi chain wallets

it can be tested by installing `xdefi`, `taho`, `coinbase` and `metamask` .

# Checklist:

### DIFF
--- a/wallets/core/src/wallet.ts
+++ b/wallets/core/src/wallet.ts
@@ -183,33 +183,35 @@ class Wallet<InstanceType = any> {
     let nextAccounts: string[] = [];
     let nextNetwork: Network | null | undefined = null;
     if (Array.isArray(connectResult)) {
+      let activeEvmNetwork: Network | null = null;
       const accounts = connectResult.flatMap((blockchain) => {
         const chainId = blockchain.chainId || Networks.Unknown;
         // Try to map chainId with a Network, if not found, we use chainId directly.
         const network =
           getBlockChainNameFromId(chainId, this.info.supportedBlockchains) ||
           Networks.Unknown;
+
+        /*
+         * When connecting to an evm instance, it will return address and wallet's active chain.
+         * On switch network we are comparing state's network and what passed as requestedNetwork.
+         * This code is for making sure we are setting correct active chain in state if it's evm.
+         */
+        if (!activeEvmNetwork && network !== Networks.Unknown) {
+          const blockchainMeta = this.info.supportedBlockchains.find(
+            (blockchain) => blockchain.name === network
+          );
+          if (blockchainMeta?.info?.infoType === 'EvmMetaInfo') {
+            activeEvmNetwork = network;
+          }
+        }
         // TODO: second parameter should be `string` when we decided to open source the package.
         return accountAddressesWithNetwork(blockchain.accounts, network);
       });
-      // Typescript can not detect we are filtering out null values:(
       nextAccounts = accounts.filter(Boolean);
-      nextNetwork = requestedNetwork || this.options.config.defaultNetwork;
-      /*
-       * For providers like wallet connect with async switch network support,
-       * it's not a correct assumption that current network after connection request
-       * equals to the requested network. (for example in case of session reconnect.)
-       * In this case, we could assume that the first item in connect result equals to
-       * the current chain. Otherwise, the state of wallet-core will be wrong.
-       */
-      if (this.options.config.isAsyncSwitchNetwork) {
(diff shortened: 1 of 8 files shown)


RanGojo (approved):
Good job. I've tested related wallets and seems they are all working now as expected!

wallets/provider-metamask/src/helpers.ts
```diff
@@ -1,17 +1,81 @@
...
-      return ethereum;
+  if (Array.isArray(ethereum?.providers)) {
+    let providers = ethereum.providers;
+    if (Array.isArray(ethereum.providers[0]?.providers)) {
+      providers = ethereum.providers[0]?.providers;
     }
+    return providers.find(getMetaMask) ?? null;
   }
-  return null;
+  return ethereum.isMetaMask ? ethereum : null;
```
RanGojo, line 80: This code throws error when there is no metamask extension enabled.
```
TypeError: Cannot read properties of undefined (reading 'isMetaMask')
```

```suggestion
  return ethereum?.isMetaMask ? ethereum : null;
```

wallets/provider-metamask/src/helpers.ts
```diff
@@ -1,17 +1,81 @@
...
-    // Getting intance from overrided structure from coinbase.
-    return getCoinbaseInstance('metamask');
-  } else {
-    if (!!ethereum && ethereum.isMetaMask) {
-      return ethereum;
+  if (Array.isArray(ethereum?.providers)) {
+    let providers = ethereum.providers;
+    if (Array.isArray(ethereum.providers[0]?.providers)) {
+      providers = ethereum.providers[0]?.providers;
     }
```
RanGojo, line 77: Please add a comment for this case (for example how could be reproduced) and also previous condition if possible

wallets/provider-metamask/src/helpers.ts
```diff
@@ -1,17 +1,81 @@
...
-  // Some wallets overriding the metamask. So we need to get it properly.
-  if (isCoinbaseWalletAvailable) {
-    // Getting intance from overrided structure from coinbase.
-    return getCoinbaseInstance('metamask');
-  } else {
-    if (!!ethereum && ethereum.isMetaMask) {
-      return ethereum;
+  if (Array.isArray(ethereum?.providers)) {
+    let providers = ethereum.providers;
+    if (Array.isArray(ethereum.providers[0]?.providers)) {
```
RanGojo, line 75:
```suggestion
    if (ethereum.providers.length > 0 && Array.isArray(ethereum.providers[0]?.providers) && ethereum.providers[0]?.providers.length > 0) {
```

wallets/provider-metamask/src/helpers.ts
```diff
@@ -1,17 +1,81 @@
...
+      providers = ethereum.providers[0]?.providers;
```
RanGojo, line 76:
```suggestion
      providers = providers[0]?.providers;
```

wallets/provider-metamask/src/helpers.ts
```diff
@@ -1,17 +1,81 @@
...
     }
+    return providers.find(getMetaMask) ?? null;
```
RanGojo, line 78: I'm not sure if we should return if it's null here. I think it's better to pass it to the next line.

wallets/core/src/wallet.ts
```diff
@@ -182,34 +182,32 @@ class Wallet<InstanceType = any> {
...
+    let activeEvmNetwork: Network | null = null;
     if (Array.isArray(connectResult)) {
       const accounts = connectResult.flatMap((blockchain) => {
         const chainId = blockchain.chainId || Networks.Unknown;
         // Try to map chainId with a Network, if not found, we use chainId directly.
         const network =
           getBlockChainNameFromId(chainId, this.info.supportedBlockchains) ||
           Networks.Unknown;
+
+        // check if this chain is an evm network
```
yeager-eren, line 194: It can be something like this:

```suggestion
        // When connecting to an evm instance, it will return address and wallet's active chain. On switch network we are comparing state's network and what passed as requestedNetwork. This code is for making sure we are setting active chain in state if it's evm.
```

yeager-eren (changes requested):

yeager-eren (approved):
lgtm

PR #684, opened 2024-04-21
feat: add hover style to all routes button

RyukTheCoder:
# Summary

Added hover style to "all routes" button.

Fixes #RF-1420

# Checklist:

### DIFF
--- a/widget/embedded/src/components/Quote/Quote.styles.ts
+++ b/widget/embedded/src/components/Quote/Quote.styles.ts
@@ -63,7 +63,25 @@ export const stepsDetailsStyles = css({
 });
 export const AllRoutesButton = styled(Button, {
   backgroundColor: 'transparent',
-  border: '1px solid $secondary',
+  border: '1px solid $secondary600',
+  [`.${darkTheme} &`]: {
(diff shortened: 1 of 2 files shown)


samobasquiat (approved):

samobasquiat:
its not related to this task but please handle it as well:
Screenshot
Screenshot

widget/embedded/src/components/Quote/Quote.styles.ts
```diff
@@ -64,6 +64,22 @@ export const stepsDetailsStyles = css({
 export const AllRoutesButton = styled(Button, {
   backgroundColor: 'transparent',
   border: '1px solid $secondary',
+  transition: 'background-color 0.3s ease',
+  paddingLeft: '$10 !important',
+  paddingRight: '$10 !important',
```
samobasquiat, line 69: NIT: I think using inner styles is better because we haven't used `!important `in our codes. However, we should find a better way to override styles later.

RyukTheCoder: Fixed

PR #685, opened 2024-04-21
fix: fix token selector infinite loader

Ikari-Shinji-re:
# Summary

Sometimes, the infinite loader in the token selector, which loads more pages when the list reaches its end, doesn't work.

This bug occurs when the widget fetches wallet balances and you scroll through a long list in the token selector. Or when you scroll through a long token list, causing it to load at least one more page, and switch between browser tabs, with the current tab becoming hidden, then returning to the same tab after.

# Checklist:


RanGojo (approved):
It works now. Thanks.

PR #688, opened 2024-04-22
feat: add auto retry for fetching confirm swap

samobasquiat:
# Summary

Added an auto-retry for the confirmation step if it fails; it will only be called once.
Fixes # (issue)

It can be tested by disabling the network during the confirmation step

# Checklist:


RanGojo (approved):

widget/embedded/src/hooks/useConfirmSwap/useConfirmSwap.ts
```diff
@@ -39,7 +39,7 @@ export function useConfirmSwap(): ConfirmSwap {
 
   useEffect(() => cancelFetch, []);
 
-  const fetch: ConfirmSwap['fetch'] = async (params: Params) => {
+  const fetch: ConfirmSwap['fetch'] = async (params) => {
```
RanGojo, line 42: I didn't understand the reason for this change.

samobasquiat: Typescript infers it automatically.

PR #694, opened 2024-04-27
fix: sync notifications with persisted swaps

Ikari-Shinji-re:
# Summary

Sometimes, browsers may delete `indexedDB` data, causing notifications and swaps to become out of sync. As a result, clicking on notifications could lead users to empty pages because the related swaps no longer exist. These notifications then remain unread indefinitely. This pull request tackles this problem by addressing the issue and removing unnecessary notifications from the user's local storage. Additionally, it introduces a button for clearing all notifications.
https://developer.mozilla.org/en-US/docs/Web/API/Storage_API/Storage_quotas_and_eviction_criteria

Initiate a swap.
Use Chrome `DevTools` to clear the `indexedDB` database for swaps.
After refreshing the page, no notifications should remain.

If a swap has been completed (whether successful or failed) and the user has read the associated notifications, those notifications should be deleted from local storage. This can be verified in Chrome `DevTools` by searching for `notifications` in the `localStorage` section.

# Checklist:

### DIFF
--- a/widget/embedded/src/hooks/useSyncNotifications/useSyncNotifications.ts
+++ b/widget/embedded/src/hooks/useSyncNotifications/useSyncNotifications.ts
@@ -1,0 +1,25 @@
+import { useManager } from '@rango-dev/queue-manager-react';
+import { useEffect } from 'react';
+
+import { useNotificationStore } from '../../store/notification';
+import { getPendingSwaps } from '../../utils/queue';
+
+export function useSyncNotifications() {
+  const { isSynced, syncNotifications } = useNotificationStore();
+  const { manager, state } = useManager();
+
+  useEffect(() => {
+    const shouldSyncNotifications =
+      useNotificationStore.persist.hasHydrated() &&
+      state.loadedFromPersistor &&
+      !isSynced;
+
+    if (shouldSyncNotifications) {
+      syncNotifications(getPendingSwaps(manager));
+    }
+  }, [
+    useNotificationStore.persist.hasHydrated(),
+    state.loadedFromPersistor,
+    isSynced,
+  ]);
+}
--- a/widget/embedded/src/store/notification.ts
+++ b/widget/embedded/src/store/notification.ts
@@ -1,26 +1,31 @@
 import type { Notification } from '../types/notification';
-import type {
-  Route,
-  RouteEvent,
-  StepEvent,
-} from '@rango-dev/queue-manager-rango-preset';
 
+import {
+  type PendingSwapWithQueueID,
+  type Route,
+  type RouteEvent,
+  type StepEvent,
+} from '@rango-dev/queue-manager-rango-preset';
 import { create } from 'zustand';
 import { persist, subscribeWithSelector } from 'zustand/middleware';
 
 import createSelectors from './selectors';
 
 export interface NotificationState {
+  isSynced: boolean;
   notifications: Notification[];
   setNotification: (event: RouteEvent | StepEvent, route: Route) => void;
-  setAsRead: (requestId: Notification['requestId']) => void;
-  getUnreadNotifications: () => Notification[];
+  removeNotification: (requestId: Notification['requestId']) => void;
+  getNotifications: () => Notification[];
+  syncNotifications: (swaps: PendingSwapWithQueueID[]) => void;
+  clearNotifications: () => void;
 }
 
 export const useNotificationStore = createSelectors(
   create<NotificationState>()(
     persist(
       subscribeWithSelector((set, get) => ({
+        isSynced: false,
         notifications: [],
         setNotification: (event, route) => {
           const fromStep = route.steps[0];
@@ -29,7 +34,6 @@ export const useNotificationStore = createSelectors(
           const notification: Notification = {
             event,
             creationTime: Date.now(),
-            read: false,
             requestId: route.requestId,
             route: {
               from: {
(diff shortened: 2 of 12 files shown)


widget/embedded/src/store/notification.ts
```diff
@@ -1,26 +1,33 @@
...
+} from '@rango-dev/queue-manager-rango-preset';
 import { create } from 'zustand';
 import { persist, subscribeWithSelector } from 'zustand/middleware';
 
+import { isSwapFinished } from '../utils/swap';
+
 import createSelectors from './selectors';
 
 export interface NotificationState {
+  isSynced: boolean;
```
nikaaru, line 17: can we remove isSynced?

widget/embedded/src/store/notification.ts
```diff
@@ -1,26 +1,33 @@
...
   notifications: Notification[];
   setNotification: (event: RouteEvent | StepEvent, route: Route) => void;
   setAsRead: (requestId: Notification['requestId']) => void;
```
nikaaru, line 20: I think it is better to change `setAsRead` to `removeNotification`.
and remove the read property from the Notification type.

widget/embedded/src/store/notification.ts
```diff
@@ -1,26 +1,33 @@
...
   getUnreadNotifications: () => Notification[];
```
nikaaru, line 21: Also we can change `getUnreadNotifications` to `getNotification`.

widget/embedded/src/store/notification.ts
```diff
@@ -54,23 +61,61 @@ export const useNotificationStore = createSelectors(
...
         setAsRead: (requestId) => {
-          set((state) => ({
-            notifications: state.notifications.map((notificationItem) =>
-              notificationItem.requestId === requestId
-                ? { ...notificationItem, read: true }
-                : notificationItem
-            ),
-          }));
+          set((state) => {
+            const notifications: Notification[] = [];
```
nikaaru, line 65: This function can be very simple to clear only the notification.
(remove `isSwapFinished` from this logic)

widget/embedded/src/store/notification.ts
```diff
@@ -54,23 +61,61 @@ export const useNotificationStore = createSelectors(
...
+          });
         },
         getUnreadNotifications: () => {
-          return get().notifications.filter(
-            (notificationItem) => !notificationItem.read
-          );
+          const { isSynced, notifications } = get();
+          let unreadNotifications: Notification[] = [];
+
+          if (isSynced) {
```
nikaaru, line 85: Can we remove `isSynced` from getting notification functionality?

nikaaru (approved):
Thanks,
LGTM

widget/embedded/src/store/notification.ts
```diff
@@ -116,6 +92,26 @@ export const useNotificationStore = createSelectors(
         name: 'notification',
         skipHydration: true,
         partialize: ({ isSynced, ...otherProps }) => otherProps,
+        version: 1,
+        migrate: (persistedState, version) => {
```
nikaaru, line 96: It was interesting
good job

widget/embedded/src/hooks/useSyncNotifications/useSyncNotifications.ts
```diff
@@ -0,0 +1,18 @@
...
+import { useNotificationStore } from '../../store/notification';
+import { getPendingSwaps } from '../../utils/queue';
+
+export function useSyncNotifications() {
+  const { syncNotifications } = useNotificationStore();
+  const { manager, state } = useManager();
+
+  const swaps = getPendingSwaps(manager);
+
+  const shouldSyncNotifications =
```
RyukTheCoder, line 12: Currently this code runs in every render and in result of that, 'syncNotifications' runs in every render which seems to be an expensive function. Is this the proper behavior?

widget/embedded/src/store/notification.ts
```diff
@@ -53,24 +57,61 @@ export const useNotificationStore = createSelectors(
...
+        clearNotifications: () => set({ notifications: [] }),
       })),
+
       {
         name: 'notification',
         skipHydration: true,
+        partialize: ({ isSynced, ...otherProps }) => otherProps,
+        version: 1,
+        migrate: (persistedState, version) => {
+          if (version === 0) {
```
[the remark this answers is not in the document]
Ikari-Shinji-re: Previously, we stored all notifications in the local storage, regardless of whether they were read or unread. Each notification had a property called "read". However, this approach is no longer necessary. To synchronize the user's persisted data in the local storage with the current version, we are considering implementing a migration process. It's important to retain this code for the time being. If we introduce further changes to the structure of this storage in the future, we may adjust this migration to ensure that user storage is updated to the latest version accordingly.

widget/embedded/src/hooks/useSyncNotifications/useSyncNotifications.ts
```diff
@@ -0,0 +1,18 @@
+import { useManager } from '@rango-dev/queue-manager-react';
+
+import { useNotificationStore } from '../../store/notification';
+import { getPendingSwaps } from '../../utils/queue';
+
+export function useSyncNotifications() {
```
RyukTheCoder, line 6:
```suggestion
export function useSyncNotifications() {
  const { isSynced, syncNotifications } = useNotificationStore();
  const { manager, state } = useManager();

  useEffect(() => {
    const shouldSyncNotifications =
      useNotificationStore.persist.hasHydrated() &&
      state.loadedFromPersistor &&
      !isSynced;

    if (shouldSyncNotifications) {
      syncNotifications(getPendingSwaps(manager));
    }
  }, [
    useNotificationStore.persist.hasHydrated(),
    state.loadedFromPersistor,
    isSynced,
  ]);
}
```

RyukTheCoder: Thanks. Please check this suggestion.

RyukTheCoder (changes requested):

RyukTheCoder (approved):
LGTM

PR #695, opened 2024-04-27
fix: improve widget performance

Ikari-Shinji-re:
# Summary

If a token is excluded from a blockchain that hosts numerous tokens, such as Ethereum, it leads to a decline in the widget's performance.

You can verify this using the playground. In either the 'from' or 'to' section, navigate to the Ethereum tokens subsection, exclude one token, and confirm the adjustments. Afterwards, attempt to interact with the widget by transitioning from the home page to the tokens page and searching for a token.

# Checklist:

### DIFF
--- a/widget/embedded/src/services/cacheService.ts
+++ b/widget/embedded/src/services/cacheService.ts
@@ -1,0 +1,35 @@
+import type { Token } from 'rango-sdk';
+
+interface CacheServiceInterface<V> {
+  get<K extends keyof V>(key: K): V[K] | undefined;
+  set<K extends keyof V>(key: K, value: V[K]): void;
+  remove<K extends keyof V>(key: K): void;
+  clear(): void;
+}
+
+class CacheService<T> implements CacheServiceInterface<T> {
+  #cache: Map<string, T[keyof T]> = new Map();
+
+  get<K extends keyof T>(key: K): T[K] | undefined {
+    return this.#cache.get(key as string) as T[K];
+  }
+
+  set<K extends keyof T>(key: K, value: T[K]): void {
+    this.#cache.set(key as string, value);
+  }
+
+  remove<K extends keyof T>(key: K): void {
+    this.#cache.delete(key as string);
+  }
+
+  clear(): void {
+    this.#cache.clear();
+  }
+}
+
+export type CachedEntries = {
+  supportedSourceTokens: Token[];
+  supportedDestinationTokens: Token[];
+};
+
+export const cacheService = new CacheService<CachedEntries>();
--- a/widget/embedded/src/store/slices/data.test.ts
+++ b/widget/embedded/src/store/slices/data.test.ts
@@ -3,6 +3,7 @@ import type { EvmBlockchainMeta, Token } from 'rango-sdk';
 
 import { assert, beforeEach, describe, expect, test } from 'vitest';
 
+import { cacheService } from '../../services/cacheService';
 import {
   createEvmBlockchain,
   createInitialAppStore,
@@ -17,6 +18,7 @@ let customTokens: [Token, Token, Token];
 let rangoBlockchain: EvmBlockchainMeta;
 
 beforeEach(() => {
+  cacheService.clear();
   rangoBlockchain = createEvmBlockchain();
 
   customTokens = [
@@ -52,6 +54,10 @@ beforeEach(() => {
   customTokens.forEach((token) => {
     const tokenHash = createTokenHash(token);
     initData._tokensMapByTokenHash.set(tokenHash, token);
+    if (!initData._tokensMapByBlockchainName[token.blockchain]) {
+      initData._tokensMapByBlockchainName[token.blockchain] = [];
+    }
+    initData._tokensMapByBlockchainName[token.blockchain].push(tokenHash);
   });
 
   const appStore = createAppStore();
@@ -349,5 +355,83 @@ describe('search in tokens', () => {
 
     expect(firstResult).toBe(djangoToken.symbol);
     expect(secondResult).toBe('RNG');
+  });
+});
+
+describe('supported tokens from config', () => {
+  test('Should ensure tokens include only tokens from the config', () => {
+    const rangoToken = customTokens[0];
+    const djangoToken = customTokens[1];
+
+    const configTokens = [rangoToken, djangoToken];
+
+    appStoreState = updateAppStoreConfig(appStoreState, {
+      from: {
+        tokens: configTokens,
+      },
+      to: { tokens: configTokens },
+    });
+
+    let sourceTokens = appStoreState.tokens({
+      type: 'source',
+    });
+
+    let destinationTokens = appStoreState.tokens({
+      type: 'destination',
+    });
+
+    expect(configTokens).toMatchObject(sourceTokens);
+    expect(configTokens).toMatchObject(destinationTokens);
+
+    appStoreState = updateAppStoreConfig(appStoreState, {
+      from: {
+        blockchains: [rangoBlockchain.name],
+        tokens: {
+          [rangoBlockchain.name]: {
+            tokens: configTokens,
+            isExcluded: false,
+          },
+        },
+      },
+      to: {
+        blockchains: [rangoBlockchain.name],
+        tokens: {
+          [rangoBlockchain.name]: {
+            tokens: configTokens,
+            isExcluded: false,
+          },
+        },
+      },
+    });
+
+    sourceTokens = appStoreState.tokens({
+      type: 'source',
+    });
+
+    destinationTokens = appStoreState.tokens({
+      type: 'destination',
+    });
+
+    expect(sourceTokens).toMatchObject(sourceTokens);
+    expect(configTokens).toMatchObject(destinationTokens);
+  });
+
+  test('Check tokens calculation is caching', () => {
+    const rangoToken = customTokens[0];
+    const djangoToken = customTokens[1];
+
+    appStoreState = updateAppStoreConfig(appStoreState, {
+      from: {
+        tokens: [rangoToken, djangoToken],
+      },
+    });
+
+    expect(cacheService.get('supportedSourceTokens')?.length ?? 0).toBe(0);
+
+    appStoreState.tokens({
+      type: 'source',
+    });
+
+    expect(cacheService.get('supportedSourceTokens')?.length ?? 0).toBe(2);
   });
 });
--- a/widget/embedded/src/store/slices/data.ts
+++ b/widget/embedded/src/store/slices/data.ts
@@ -1,16 +1,18 @@
 // We keep all the received data from server in this slice
 
 import type { ConfigSlice } from './config';
-import type { Balance, TokenHash } from '../../types';
+import type { CachedEntries } from '../../services/cacheService';
+import type { Balance, Blockchain, TokenHash } from '../../types';
 import type { Asset, BlockchainMeta, SwapperMeta, Token } from 'rango-sdk';
 import type { StateCreator } from 'zustand';
 
+import { cacheService } from '../../services/cacheService';
 import { httpService as sdk } from '../../services/httpService';
 import { compareWithSearchFor, containsText } from '../../utils/common';
-import { isTokenExcludedInConfig } from '../../utils/configs';
 import { createTokenHash, isTokenNative } from '../../utils/meta';
 import { sortLiquiditySourcesByGroupTitle } from '../../utils/settings';
 import { areTokensEqual, compareTokenBalance } from '../../utils/wallets';
+import { matchTokensFromConfigWithMeta } from '../utils';
 
 type BlockchainOptions = {
   type?: 'source' | 'destination';
@@ -30,6 +32,7 @@ export type FetchStatus = 'loading' | 'success' | 'failed';
 export interface DataSlice {
   _blockchainsMapByName: Map<string, BlockchainMeta>;
   _tokensMapByTokenHash: Map<TokenHash, Token>;
+  _tokensMapByBlockchainName: Record<Blockchain, TokenHash[]>;
   _popularTokens: Token[];
   _swappers: SwapperMeta[];
   fetchStatus: FetchStatus;
@@ -48,8 +51,9 @@ export const createDataSlice: StateCreator<
   DataSlice
 > = (set, get) => ({
   // State
-  _blockchainsMapByName: new Map<string, BlockchainMeta>(),
-  _tokensMapByTokenHash: new Map<TokenHash, Token>(),
+  _blockchainsMapByName: new Map(),
+  _tokensMapByTokenHash: new Map(),
+  _tokensMapByBlockchainName: {},
   _popularTokens: [],
   _swappers: [],
   fetchStatus: 'loading',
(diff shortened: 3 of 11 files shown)


nikaaru:
Can we have a  `ـsupportedToken` and get tokens from it wherever we want, instead of getting them from `_token` (all tokens)?
For example, if we only support 10 tokens, the token filter operation will be applied to all tokens.

nikaaru:
image
If we have two blockchains whose tokens are filtered,
and switch between them, the list of tokens is not updated

widget/embedded/src/store/slices/data.ts
```diff
@@ -96,12 +99,38 @@ export const createDataSlice: StateCreator<
...
 
+    const includedTokens = new Set();
+    const excludedTokens = new Set();
+
+    if (supportedTokensConfig) {
+      if (Array.isArray(supportedTokensConfig)) {
+        supportedTokensConfig.forEach((token) =>
+          includedTokens.add(createTokenHash(token))
+        );
+      } else if (!Array.isArray(supportedTokensConfig)) {
```
nikaaru, line 110: Do we need this condition?
`Array.isArray(supportedTokensConfig)` in else conditions is always `false`

widget/embedded/src/store/slices/data.ts
```diff
@@ -96,12 +99,38 @@ export const createDataSlice: StateCreator<
...
+        Object.values(supportedTokensConfig).forEach((value) => {
+          value.tokens.forEach((token) => {
+            if (token.blockchain === options.blockchain) {
```
nikaaru, line 113: Can we check blockchain equality before `forEach`?

widget/embedded/src/store/slices/data.ts
```diff
@@ -96,12 +99,38 @@ export const createDataSlice: StateCreator<
...
+      }
+    }
+
     const list = tokensFromState
       .filter((token) => {
-        if (
-          supportedTokensConfig &&
-          isTokenExcludedInConfig(token, supportedTokensConfig)
-        ) {
+        const tokenExcluded =
```
nikaaru, line 127: In the configuration, we selected only two blockchains and limited each to a few tokens.
When we select all blockchains in the widget, it will show all tokens instead of showing the selected tokens.

Ikari-Shinji-re:
> image If we have two blockchains whose tokens are filtered, and switch between them, the list of tokens is not updated

This bug exists in our production version. You can verify it with the same configuration. If two blockchains have the same number of tokens, switching between blockchains doesn't trigger a render. While the store's state is valid, the UI is incorrect. We can address this in a separate task, possibly by using useDeepCompareEffect.

nikaaru:
When we have selected all the tokens of a blockchain, we don't have any data in the include or exclude section in the config.
In this case, all tokens must be in the Include Tokens section.
Now, when we select all the tokens in the widget, it does not work properly

nikaaru:
Also, when we exclude a series of tokens in the config for a blockchain when we have selected all the tokens in the token selector component in the widget, the tokens of that blockchain will not be displayed.

widget/embedded/src/store/slices/data.ts
```diff
@@ -95,13 +98,47 @@ export const createDataSlice: StateCreator<
     const blockchains = get().blockchains({
       type: options.type,
     });
+    const includedTokens = new Set<string>();
+    const excludedTokens = new Set<string>();
+    if (supportedTokensConfig) {
+      const shouldIncludeOrExcludeTokens = (
```
nikaaru, line 104: I think `shouldIncludeOrExcludeTokens` function doesn't give a good idea of ​​what it's supposed to do.

widget/embedded/src/store/slices/data.ts
```diff
@@ -95,13 +98,47 @@ export const createDataSlice: StateCreator<
...
 
     const list = tokensFromState
       .filter((token) => {
-        if (
-          supportedTokensConfig &&
-          isTokenExcludedInConfig(token, supportedTokensConfig)
-        ) {
+        const tokenExcluded =
+          (includedTokens.size > 0 &&
+            !includedTokens.has(createTokenHash(token))) ||
```
nikaaru, line 137: I don't think the presence of the `or condition` here works properly

widget/embedded/src/store/slices/data.ts
```diff
@@ -29,6 +29,9 @@ export type FetchStatus = 'loading' | 'success' | 'failed';
 export interface DataSlice {
   _blockchainsMapByName: Map<string, BlockchainMeta>;
   _tokensMapByTokenHash: Map<TokenHash, Token>;
+  _tokensMapByBlockchainName: Record<string, Record<TokenHash, Token>>;
+  _supportedSourceTokens: Map<TokenHash, Token>;
```
nikaaru, line 33: We currently have this Map for the token,
I think it is better to use Array for supported tokens.

widget/embedded/src/store/slices/data.ts
```diff
@@ -346,4 +324,50 @@ export const createDataSlice: StateCreator<
...
+    if (Array.isArray(options.data.tokens)) {
+      options.data.tokens.forEach((token) => {
+        const tokenHash = createTokenHash(token);
+        const tokenFromMeta = tokens?.[token.blockchain]?.[tokenHash];
+        if (tokenFromMeta) {
+          nextState[tokenHash] = tokenFromMeta;
+        }
+      });
+    } else {
+      if (!options.data.blockchains) {
```
nikaaru, line 339: Maybe it's better to use `else if` here.

widget/embedded/src/store/slices/data.ts
```diff
@@ -346,4 +324,50 @@ export const createDataSlice: StateCreator<
...
+        nextState = Object.fromEntries(get()._tokensMapByTokenHash.entries());
+      }
+      options.data.blockchains?.forEach((blockchain) => {
+        const value = !Array.isArray(options.data.tokens)
+          ? options.data.tokens?.[blockchain]
+          : undefined;
+        if ((!value || value?.isExcluded) && tokens?.[blockchain]) {
```
nikaaru, line 346: Excluded tokens do not work properly.
It seems that this condition is not correct.

widget/embedded/src/store/slices/data.ts
Ikari-Shinji-re: it isn't working here

nikaaru: foreEach on `options.data.blockchains` worked only if `options.data.blockchains` existed.

widget/embedded/src/store/slices/data.ts
```diff
@@ -313,4 +323,49 @@ export const createDataSlice: StateCreator<
       throw error;
     }
   },
+  setSupportedTokens: (options) => {
+    const tokens = get()._tokensMapByBlockchainName;
```
nikaaru, line 327:
```suggestion
    const tokensMapByBlockchainName = get()._tokensMapByBlockchainName;
```

nikaaru (approved):
Thanks,
LGTM

yeager-eren (changes requested):
We've reviewed this on a meet, it needs some changes on store includes:
1. Remove `  _supportedSourceTokens: [], _supportedDestinationTokens: [],` from state.
2. Try to use `_tokensMapByTokenHash` whenever possible and if `_tokensMapByBlockchainName` needed, keep `TokenHash` only, not `Token` as well.
3. Use memoization for avoiding redundant calculation and directly use it in `.tokens()` which means `setSupportedTokens` can be removed as well. 
4. Write some tests to cover this changes.

widget/embedded/src/store/slices/data.test.ts
```diff
@@ -351,3 +356,89 @@ describe('search in tokens', () => {
     expect(secondResult).toBe('RNG');
   });
 });
+
+describe('supported tokens from config', () => {
```
yeager-eren, line 360: We need a test to cover cache is hitting or not. Something like this:

```js
test('Check tokesn calculation is caching', () => {
    const rangoToken = customTokens[0];
    const djangoToken = customTokens[1];

    appStoreState = updateAppStoreConfig(appStoreState, {
      from: {
        tokens: [rangoToken, djangoToken],
      },
    });

    expect(cacheService.get('supportedSourceTokens')?.length || 0).toBe(0);
    appStoreState.tokens({
      type: 'source',
    });

    expect(cacheService.get('supportedSourceTokens')?.length || 0).toBe(1);
  });
```

yeager-eren: The whole point of this is adding a cache layer. we should cover its behavior using tests. like previous comment. You can also check updaing, deleting and other cases if you have.

widget/embedded/src/store/slices/data.ts
```diff
@@ -1,17 +1,80 @@
...
 
+import { cacheService } from '../../services/cacheService';
 import { httpService as sdk } from '../../services/httpService';
 import { compareWithSearchFor, containsText } from '../../utils/common';
-import { isTokenExcludedInConfig } from '../../utils/configs';
 import { createTokenHash, isTokenNative } from '../../utils/meta';
 import { sortLiquiditySourcesByGroupTitle } from '../../utils/settings';
 import { areTokensEqual, compareTokenBalance } from '../../utils/wallets';
 
+export const calculateSupportedTokens = (params: {
```
yeager-eren, line 21: To be honest, this function is seems to be written in Japanese to me , Could you clear the assumptions by adding some comments on the flow?

yeager-eren: Also you can create a `store/utils/data.ts` to move this function to that file.

yeager-eren: Please add some unit tests to test and show how it should behave more clearly. You can create a `store/utils/data.test.ts` to put your unit tests.

widget/embedded/src/store/slices/data.ts
```diff
@@ -48,8 +114,11 @@ export const createDataSlice: StateCreator<
   DataSlice
 > = (set, get) => ({
   // State
-  _blockchainsMapByName: new Map<string, BlockchainMeta>(),
-  _tokensMapByTokenHash: new Map<TokenHash, Token>(),
+  _blockchainsMapByName: new Map(),
+  _tokensMapByTokenHash: new Map(),
+  _tokensMapByBlockchainName: {},
+  _supportedSourceTokens: [],
```
yeager-eren, line 120: Why do we need `_supportedSourceTokens` anymore? and also `_supportedDestinationTokens`.

yeager-eren (changes requested):
Thanks. i've put my thoughts on your PR.

widget/embedded/src/store/utils/data.ts
```diff
@@ -0,0 +1,93 @@
...
+    config.blockchains.forEach((blockchain) => {
+      const value = !Array.isArray(config.tokens)
+        ? config.tokens?.[blockchain]
+        : undefined;
+      /**
+       * If no tokens exist for a blockchain or if the list of tokens is excluded for a blockchain,
+       * we include all tokens for that blockchain from the meta response.
+       */
+      if (
+        (!value || value?.isExcluded) &&
```
yeager-eren, line 63: This `if` and next `forEach` have a coupled logic which makes it confusing. I suggest to separate them to clearly communicate what are we do here.

```js

if(!value && !!meta.tokensMapByBlockchainName?.[blockchain]){
// ...

return;   // <-----
}

if(value.isExcluded){
// Iterate over meta tokens and exclude `value.tokens` 
}
else{
// Iterate over value.tokens and add them to `result`
}

```

widget/embedded/src/store/utils/data.ts
```diff
@@ -0,0 +1,93 @@
...
+import type { Token } from 'rango-sdk';
+
+import { createTokenHash } from '../../utils/meta';
+
+/**
+ * This function retrieves tokens and blockchains from the widget configuration and combines them with token data from the meta response to determine which tokens should be displayed in the widget.
+ * To optimize performance, the function minimizes unnecessary iterations over the meta tokens.
+ * The result of this function should be cached and recalculated whenever the widget configuration changes.
+ */
+export function calculateSupportedTokens(params: {
```
yeager-eren, line 12:
```suggestion
export function matchTokensFromConfigWithMeta(params: {
```

yeager-eren (changes requested):

widget/embedded/src/store/utils/data.test.ts
```diff
@@ -0,0 +1,101 @@
...
+        [tokens[2].blockchain]: { tokens: [tokens[2]], isExcluded: false },
+      },
+    };
+    const result = calculateSupportedTokens(mockData);
+    expect(result).toEqual([tokens[0], tokens[1]]);
+  });
+
+  test('should include all tokens from a blockchain if that blockchain does not have any tokens in config.tokens', () => {
+    mockData.config = {
+      blockchains: [tokens[0].blockchain],
```
yeager-eren, line 84: the intention can be confusing while reading test when using `tokens[0].blockchain`. Please use hard coded values e.g. `blockchains: ["blockchainA"]`.
Note: Update all `mockData.config.blockchains` to hard coded values.

widget/embedded/src/store/utils/data.test.ts
```diff
@@ -0,0 +1,101 @@
...
+const mockData: Parameters<typeof calculateSupportedTokens>[0] = {
+  type: 'source',
+  meta: {
+    tokensMapByTokenHash: new Map(),
+    tokensMapByBlockchainName: {},
+  },
+  config: { tokens: [], blockchains: undefined },
+};
+
+tokens.forEach((token) => {
```
yeager-eren, line 26: you can make this part of code a function called `generateTokensMapByBlockchainName` or whatever.
signature:
```
generateTokensMapByBlockchainName(tokens: Token[]): Record<Blockchain, TokenHash[]>
```
It will be a pure function and you can run it to generate hashes in your tests:

```
test( ... , () => {
mockData.meta.tokensMapByBlockchainName = generateTokensMapByBlockchainName(...)
})
```

yeager-eren: I also really don't think making a global object and mutating it in every test is a good idea. Making the tests fragile and hard to read.

My suggestion is moving `mockData` to each tests and only keep `const tokens` globally. You should only make a util for `mockdata.meta` to be reuse in all your tests. In my previous comment I described what it can look like.

yeager-eren (changes requested):

widget/embedded/src/store/utils/data.ts
```diff
@@ -0,0 +1,93 @@
...
+
+      if (tokenFromMeta) {
+        result[tokenHash] = tokenFromMeta;
+      }
+    });
+    /**
+     * From this point onward,
+     * the function handles the other type of widget configuration.
+     */
+  } else if (!config.blockchains) {
```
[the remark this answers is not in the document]
Ikari-Shinji-re: `Object.values(meta.tokensMapByTokenHash)` returns an empty array because Map objects in JavaScript are not plain objects.

widget/embedded/src/services/cacheService.ts
```diff
@@ -0,0 +1,35 @@
...
+
+interface CacheServiceInterface<V> {
+  get<K extends keyof V>(key: K): V[K] | undefined;
+  set<K extends keyof V>(key: K, value: V[K]): void;
+  remove<K extends keyof V>(key: K): void;
+  clear(): void;
+}
+
+class CacheService<T> implements CacheServiceInterface<T> {
+  private cache: Map<string, T[keyof T]> = new Map();
```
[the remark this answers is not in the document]
Ikari-Shinji-re: Alright, we need to update the `tsconfig.json` to enable the use of private fields.

widget/embedded/src/store/utils/data.test.ts
```diff
@@ -5,97 +5,151 @@ import { describe, expect, test } from 'vitest';
...
+  { ...createToken(), blockchain: BLOCKCHAIN_A },
+  { ...createToken(), blockchain: BLOCKCHAIN_A },
+  { ...createToken(), blockchain: BLOCKCHAIN_B },
+  { ...createToken(), blockchain: BLOCKCHAIN_C },
 ];
 
-const mockData: Parameters<typeof calculateSupportedTokens>[0] = {
-  type: 'source',
-  meta: {
+type A = Parameters<typeof matchTokensFromConfigWithMeta>[0];
```
yeager-eren, line 21: Better naming please, e.g. `MatchTokensFromConfigWithMetaParam`. Or maybe a shorter one.

yeager-eren (approved):
Thanks. Great work. I just requested a minor change. you can squash and make it ready for merge.

widget/embedded/src/store/utils/data.ts
```diff
@@ -56,47 +98,32 @@ export function matchTokensFromConfigWithMeta(params: {
...
-        addTokens(meta.tokensMapByBlockchainName[blockchain]);
-        targetTokensForBlockchain.tokens.forEach((token) => {
-          const tokenHash = createTokenHash(token);
-          delete result[tokenHash];
-        });
-      } else {
-        //We retrieve the corresponding token from the meta and include it in the result.
-        addTokens(targetTokensForBlockchain.tokens);
-      }
+      handleConfigTokensObject(configTokens, Array.from(blockchains));
```
yeager-eren, line 116: This is really good to reuse the logic here, but because both `handleConfigTokensObject` and `addTokens` are implicitly update environment (`result` variable), it made the code less readable and harder to follow.

I suggest two things here: 
1. if you want to keep the function (`handleConfigTokensObject`), rename to a more descriptive name and also make the function pure by returning `result`  instead of mutating the parent scope. I mean `function handleConfigTokensObject(...): Record<TokenHash, Token>` then update the variable like this `result = handleConfigTokensObject(...)`.
2. Your code can be restructured to not need to make a separate function for reusing the code (`handleConfigTokensObject`). 
It can be something like this:

```js

let configBlockchains;
if (config.blockchains?.length) {
configBlockchains = config.blockchains;
}else{
configBlockchains = new Set(Object.keys(configTokens));

 Object.keys(meta.tokensMapByBlockchainName).forEach((blockchain) => {
        if (!blockchains.has(blockchain)) {
          addTokens(meta.tokensMapByBlockchainName[blockchain]);
        }
      });
}

// Rest of code that has been moved to `handleConfigTokensObject`

```

yeager-eren (changes requested):

yeager-eren (approved):
Thanks. Great work!

PR #696, opened 2024-04-27
fix: make sure to correctly pass validation parameters when setting up a new swap

Ikari-Shinji-re:
# Summary

We should fix passing validations parameters in create transaction step:
```validations: {balance: false, fee: false}```

if user proceed anyway => pass both `false`
otherwise => pass both `true`

# Checklist:


RanGojo (approved):
Good job.

PR #683, opened 2024-04-21
fix: address passing wallet-connect project id

samobasquiat:
# Summary

fixing wallet connect project id

Fixes # (issue)

# How did you test this change?

linking with Dapp and click on wallet connect in modal 

# Checklist:

- [x] I have performed a self-review of my code

PR #697, opened 2024-04-27
feat: improve wallet connect variables check

RyukTheCoder:
# Summary

Implemented throwing error if WalletConnect is not excluded and WC_PROJECT_ID have not passed to provider-all.

Fixes # 1464

Tested it by excluding and including wallet-connect in widget wallets list and also adding and removing WC_PROJECT_ID.

# Checklist:

### DIFF
--- a/wallets/provider-all/src/helpers.ts
+++ b/wallets/provider-all/src/helpers.ts
@@ -1,0 +1,14 @@
+import type { ProviderInterface } from '@rango-dev/wallets-react';
+import type { WalletType } from '@rango-dev/wallets-shared';
+
+import { WalletTypes } from '@rango-dev/wallets-shared';
+
+export const isWalletConnectExcluded = (
+  selectedProviders?: (WalletType | ProviderInterface)[]
+) =>
+  selectedProviders &&
+  !selectedProviders.find((provider) =>
+    typeof provider === 'string'
+      ? provider === WalletTypes.WALLET_CONNECT_2
+      : provider.getWalletInfo([]).name === 'WalletConnect'
+  );
--- a/wallets/provider-all/src/index.ts
+++ b/wallets/provider-all/src/index.ts
@@ -1,3 +1,7 @@
+import type { Environments as WalletConnectEnvironments } from '@rango-dev/provider-walletconnect-2';
+import type { ProviderInterface } from '@rango-dev/wallets-react';
+import type { WalletType } from '@rango-dev/wallets-shared';
+
 import * as argentx from '@rango-dev/provider-argentx';
 import * as bitget from '@rango-dev/provider-bitget';
 import * as braavos from '@rango-dev/provider-braavos';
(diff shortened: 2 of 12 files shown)


wallets/provider-all/src/helpers.ts
```diff
@@ -0,0 +1,12 @@
+import type { ProviderInterface } from '@rango-dev/wallets-react';
+import type { WalletType } from '@rango-dev/wallets-shared';
+
+import { WalletTypes } from '@rango-dev/wallets-shared';
+
+export const isWalletConnectExcluded = (
+  selectedProviders?: (WalletType | ProviderInterface)[]
+) =>
+  selectedProviders &&
+  !selectedProviders.find(
```
yeager-eren, line 10: You are only checking string mode of the union (`selectedProviders`).

I guess it should be something like this:
```ts
  !selectedProviders.find((provider) => {
    if (typeof provider === 'string') {
      return provider === WalletTypes.WALLET_CONNECT_2;
    }
    return provider.getWalletInfo([]).name;
  });
```

wallets/provider-all/src/index.ts
```diff
@@ -28,10 +32,26 @@ import * as trustwallet from '@rango-dev/provider-trustwallet';
...
+interface Environments {
+  [key: string]: Record<string, string>;
+  walletconnect2: WalletConnectEnvironments;
+}
 
-export const allProviders = (enviroments?: Enviroments) => {
-  walletconnect2.init(enviroments?.walletconnect2 || {});
+export const allProviders = (
+  environments?: Environments,
+  selectedProviders?: (WalletType | ProviderInterface)[]
```
yeager-eren, line 44: I think you don't need to make a separate param. you can add it to `environments`. you can rename `environments` to `options`. I'm okay with both approach.

yeager-eren (changes requested):
Thanks. I've added some comments on your code. Please ask @samobasquiat to check it also.

wallets/provider-all/src/helpers.ts
RyukTheCoder: Tnx, Fixed

yeager-eren (approved):
LGTM. please ask @samobasquiat to do a second review also.

samobasquiat:
Great job.
I think we should also consider making the `config` parameter in the `useWalletList` function required.

RyukTheCoder:
> Great job. I think we should also consider making the `config` parameter in the `useWalletList` function required. For example, recently, I fixed a bug related to this hook with the changes in https://github.com/rango-exchange/rango-client/pull/683/files which we missed due to the absence of type guarding.

I didn't get the point of `config` being required and getting it from `useAppStore` and passing it to  `useWalletList` everywhere we want to use it, so I moved getting config from `useAppStore` inside the `useWalletList` hook.

samobasquiat (approved):

PR #704, opened 2024-04-29
fix: replace vercel build command with a custom script

yeager-eren:
# Summary

Running `vercel build` causes to re-run `yarn install` and `yarn build` (on target), it's redundant and has some linking issues which broke our pipelines. In this PR, we replace the command with an equivalent script.  

Fixes # (issue)


scripts/deploy/utils.mjs
```diff
@@ -41,6 +58,10 @@ export async function deploySingleProjectToVercel(pkg) {
     { env }
   );
 
+  const currentDir = dirname(fileURLToPath(import.meta.url));
+  const configPath = join(currentDir, 'vercel.json');
+  await makeOutputFolderForBuildOutputApi(pkg.location, configPath);
+
   await execa(
     'vercel',
```
nikaaru, line 66: Do we still need Vercel Build?

PR #707, opened 2024-04-30
chore: remove vercel build command

yeager-eren:
# Summary

[In previous PR](https://github.com/rango-exchange/rango-client/pull/704), we missed to remove the `vercel build` command. This PR removes the command and also fix a bug on `config.json` for vercel deployment.


.github/workflows/deploy.yml
```diff
@@ -23,9 +23,12 @@ jobs:
           GH_TOKEN: ${{ github.token }}
           VERCEL_ORG_ID: ${{ secrets.VERCEL_ORG_ID }}
           VERCEL_TOKEN: ${{ secrets.VERCEL_TOKEN }}
-          VERCEL_PROJECT_WALLETS: ${{ secrets.VERCEL_PROJECT_WALLETS }}
```
nikaaru, line 26: In my opinion, if we don't need these packages
Check in the **_listPackagesToBeDeployed_** function
If the project ID does not exist, it will not be included in the output value

yeager-eren: Good catch! Thanks.
I fixed it in my last commit. All the clients can be deployed again as before.

nikaaru (approved):
Thanks.
LGTM

PR #710, opened 2024-05-05
fix: fix bugs in fetching quotes

Ikari-Shinji-re:
# Summary

When we move from the routes page to the settings page and then return to the routes page, the fetching of quotes doesn't activate. 
The debouncing of the quotes request should only occur when there's a change in the input amount.
If the parameters for the quote are altered in the widget configuration (such as in the playground), the widget shouldn't be left in a broken state. This issue currently occurs on the `confirm swap` page and `routes` page. To address this, we can reset the quote state and redirect the user to the home page.

You can disable certain swappers or tweak the slippage, ensuring they're properly applied to the quotes request. 
After setting all the quotes parameters. Navigate from the home page to other pages. The quotes request should be triggered without any debounce.

# Checklist:

### DIFF
--- a/widget/embedded/src/pages/Routes.tsx
+++ b/widget/embedded/src/pages/Routes.tsx
@@ -15,7 +15,6 @@ import { useQuoteStore } from '../store/quote';
 export function RoutesPage() {
   const navigate = useNavigate();
   const navigateBack = useNavigateBack();
-
   const {
     selectedQuote,
     refetchQuote,
@@ -41,6 +40,7 @@ export function RoutesPage() {
       header={{
         onWallet: () => {
           navigate(wallets_url);
+          updateQuotePartialState('refetchQuote', true);
         },
         onBack: () => {
           updateQuotePartialState('refetchQuote', false);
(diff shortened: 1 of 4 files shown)


widget/embedded/src/pages/Routes.tsx
```diff
@@ -41,6 +41,7 @@ export function RoutesPage() {
       header={{
         onWallet: () => {
           navigate(wallets_url);
+          updateQuotePartialState('refetchQuote', true);
```
samobasquiat, line 44: I didn't get it. Could you explain it?

samobasquiat: Also whenever I get back to the `Home` or `Routes` pages the `bestRoute` is called. Is it ok?

Ikari-Shinji-re: Whenever a user navigates to the home or routes page, the optimal approach is to trigger a request for the latest quotes, including updated parameters such as source and destination tokens, input amount, swappers, and slippage, to ensure the most current data is displayed. However, there are exceptions to this rule. If a user navigates between the home and routes pages, the request should not be initiated to prevent disrupting and confusing users who have already selected a quote.
A flag is stored in the quotes store to determine whether to initiate the request for quotes or not based on this conditions.

samobasquiat (approved):

samobasquiat (approved):

PR #712, opened 2024-05-06
fix: disable the confirm wallet button if all required wallets are not selected

Ikari-Shinji-re:
# Summary

If a user connects a wallet different from the required type for the swap process and attempts to confirm the wallet, no wallet is selected in the confirmation modal. However, if the user inputs a valid custom address, the confirm button becomes enabled, allowing the user to proceed with the swap. In this scenario, no selected wallet is transmitted to the server, potentially resulting in the failure of the transaction creation process: 

First, connect only the `Keplr` wallet. Then, attempt a swap on an EVM route. In the confirm wallet modal, enter a valid custom address.

# Checklist:

### DIFF
--- a/widget/embedded/src/utils/swap.ts
+++ b/widget/embedded/src/utils/swap.ts
@@ -40,7 +40,7 @@ import {
 import { getBlockchainShortNameFor } from './meta';
 import { numberToString } from './numbers';
 import { getPriceImpact, getRequiredBalanceOfWallet } from './quote';
-import { getRequiredChains } from './wallets';
+import { getQuoteWallets } from './wallets';
 
 export function getOutputRatio(
   inputUsdValue: BigNumber | null,
@@ -253,24 +253,6 @@ export function canComputePriceImpact(
     parseFloat(inputAmount || '0') !== 0 &&
     !!quote
   );
-}
-
-export function requiredWallets(quote: SelectedQuote | null) {
-  const wallets: string[] = [];
-
-  quote?.swaps.forEach((swap) => {
-    const currentStepFromBlockchain = swap.from.blockchain;
-    const currentStepToBlockchain = swap.to.blockchain;
-    let lastAddedWallet = wallets[wallets.length - 1];
-    if (currentStepFromBlockchain != lastAddedWallet) {
-      wallets.push(currentStepFromBlockchain);
-    }
-    lastAddedWallet = wallets[wallets.length - 1];
-    if (currentStepToBlockchain != lastAddedWallet) {
-      wallets.push(currentStepToBlockchain);
-    }
-  });
-  return wallets;
 }
 
 export const getUsdPrice = (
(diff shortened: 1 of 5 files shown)


widget/embedded/src/utils/swap.ts
```diff
@@ -753,25 +751,43 @@ export function shouldRetrySwap(pendingSwap: PendingSwap) {
...
   showCustomDestination: boolean,
   customDestination: string | null,
   quote: SelectedQuote | null,
   selectedWallets: { walletType: string; chain: string }[],
   lastStepToBlockchain?: BlockchainMeta
-) {
+): boolean {
+  if (!quote || !lastStepToBlockchain || fetching) {
+    return true;
+  }
```
RanGojo, line 764: You used `lastStepToBlockchain` to only check if a custom destination is valid or not. So it doesn't make sense to have this condition here even if it works. Because user may not enter a custom address.

widget/embedded/src/utils/swap.ts
```diff
@@ -255,22 +256,19 @@ export function canComputePriceImpact(
...
-    }
-    lastAddedWallet = wallets[wallets.length - 1];
-    if (currentStepToBlockchain != lastAddedWallet) {
-      wallets.push(currentStepToBlockchain);
-    }
-  });
-  return wallets;
+export function getRequiredWallets(quote: SelectedQuote) {
+  return removeDuplicateFrom(quote.swaps.map((swap) => swap.from.blockchain));
 }
```
RanGojo, line 272: We have a function named `getRequiredChains`. I think you are doing something similar here. Also, you didn't consider internal swaps blockchains here which are required to be considered e.g. for aggregated swaps in cosmos-based blockchains.

widget/embedded/src/utils/swap.ts
```diff
@@ -753,25 +751,43 @@ export function shouldRetrySwap(pendingSwap: PendingSwap) {
...
 
-export function confirmSwapDisabled(
+export function isConfirmSwapDisabled(
   fetching: boolean,
   showCustomDestination: boolean,
   customDestination: string | null,
   quote: SelectedQuote | null,
   selectedWallets: { walletType: string; chain: string }[],
   lastStepToBlockchain?: BlockchainMeta
```
RanGojo, line 760: shouldn't `lastStepToBlockchain` be required arg?

RanGojo (changes requested):

RanGojo:
Is it correct that last step wallet is still selected in ui even if user enters a custom address?

Recording 2024-05-07 at 11 58 52

widget/embedded/src/utils/wallets.ts
```diff
@@ -199,34 +199,35 @@ export function prepareAccountsForWalletStore(
...
-      wallets.push(currentStepFromBlockchain);
-    }
-    if (!wallets.includes(currentStepToBlockchain)) {
-      wallets.push(currentStepToBlockchain);
-    }
+    wallets.add(currentStepFromBlockchain);
 
     // Check if internalSwaps array exists
-    if (swap.internalSwaps && Array.isArray(swap.internalSwaps)) {
+    if (swap.internalSwaps) {
```
RanGojo, line 215: I think we could still have this check.

Ikari-Shinji-re: It's redundant. The `internalSwaps` type is a union of null and array. If `swap.internalSwaps` is not falsy, TypeScript recognizes it as an array within the if block.

RanGojo (approved):
LGTM.

PR #718, opened 2024-05-11
fix: update design for not-selected blockchain or token

mikasackermn:
# Summary

Update design for not-selected blockchain or token (and color hover has been updated based on Figma.) 

Fixes # (issue)

# Checklist:


samobasquiat:
It seems that the texts should be changed as well 
image

samobasquiat:
when there is no output amount in Figma `Zero` color looks like inputAmount (a gray kind) .
image

samobasquiat (changes requested):

RyukTheCoder:
When no token and chain is selected and I hover on it, I think the blockchain icon should be transparent but it is white:

image

image

RyukTheCoder:
I think when no token is selected, balance should not be visible.

image

RyukTheCoder (approved):
LGTM

PR #719, opened 2024-05-12
feat: update wallets page

mikasackermn:
# Summary

We have to implement a filter for wallets to be filltered by transaction types (like what we have in Blockchain Selector page)

Fixes # (issue)

Note 1:
Skeleton should be updated.

Note 2:
There are some cases that all the avalilable list shouldn't be shown which means some wallets are excluded from config, in this case we only should show categories that has more than one wallet.

 Check both the widget and the playground (make sure it works properly by excluding wallets from the playground).

# Checklist:


widget/embedded/src/utils/wallets.ts
```diff
@@ -82,7 +88,11 @@ export function mapWalletTypesToWalletInfo(
         img: image,
         installLink,
         showOnMobile,
+        supportedChains,
       } = getWalletInfo(type);
+      const types = supportedChains.map((item) => item.type);
+      const blockchainTypes = [...new Set(types)];
```
Ikari-Shinji-re, line 94: I believe we have a helper function called `removeDuplicateFrom(array)`. You can utilize it in this context.

widget/embedded/src/utils/wallets.ts
```diff
@@ -533,3 +544,23 @@ export const isFetchingBalance = (
...
+      }
+    });
+    return types;
+  }, []);
+
+  const filteredBlockchains = blockchains.filter((item) =>
+    blockchainTypes.includes(item.type)
+  );
+
+  return filteredBlockchains;
```
Ikari-Shinji-re, line 565: You can use `Set` and `forEach` for better readability:
```ts
export function filterBlockchainsByWalletTypes(
  wallets: ModalWalletInfo[],
  blockchains: BlockchainMeta[]
) {
  const uniqueBlockchainTypes = new Set<TransactionType>();

  wallets.forEach((wallet) => {
    wallet.blockchainTypes.forEach((type) => {
      uniqueBlockchainTypes.add(type);
    });
  });

  const filteredBlockchains = blockchains.filter((blockchain) =>
    uniqueBlockchainTypes.has(blockchain.type)
  );

  return filteredBlockchains;
}
```

widget/embedded/src/pages/WalletsPage.tsx
```diff
@@ -64,17 +76,41 @@ export function WalletsPage() {
   const selectedWalletState =
     selectedWallet?.state || WalletState.NOT_INSTALLED;
 
+  const filteredBlockchains = filterBlockchainsByWalletTypes(list, blockchains);
```
Ikari-Shinji-re, line 79: Until the metadata response is received, there will be no wallet list, and consequently, no skeleton for the wallet list either :

mikasackermn: This wasn't an issue for me. I appreciate you checking once more.
Screenshot from 2024-05-14 11-24-06

widget/embedded/src/pages/WalletsPage.tsx
```diff
@@ -90,14 +90,17 @@ export function WalletsPage() {
...
+    fetchMetaStatus === 'loading'
+      ? list
+      : list.filter((wallet) => {
+          for (const type of wallet.blockchainTypes) {
+            if (filterByType(type, blockchainCategory)) {
+              return true;
+            }
+          }
+          return false;
+        });
```
Ikari-Shinji-re, line 103: I believe this condition alters the default behavior. If the meta request fails, the skeleton list will be empty.

Ikari-Shinji-re (approved):

widget/embedded/src/pages/WalletsPage.tsx
```diff
@@ -74,17 +87,44 @@ export function WalletsPage() {
   const selectedWalletState =
     selectedWallet?.state || WalletState.NOT_INSTALLED;
 
+  const filteredBlockchains = filterBlockchainsByWalletTypes(list, blockchains);
+  const countActiveCategories = getCountCategories(filteredBlockchains);
+  const showCategory = countActiveCategories !== 1;
+
+  const filteredWallets = list.filter((wallet) => {
+    if (blockchainCategory === BlockchainCategories.ALL) {
```
RyukTheCoder, line 95: this comparison is done here and also again in the `filterByType` function.

widget/ui/src/components/Wallet/Wallet.types.ts
```diff
@@ -14,6 +15,7 @@ export type WalletInfo = {
   image: string;
   type: string;
   showOnMobile?: boolean;
+  blockchainTypes: TransactionType[];
```
RyukTheCoder, line 18: I think maybe it's better to eliminate the paradigm shift between blockchainType and TransactionType and instead name this variable transactionTypes

RyukTheCoder (changes requested):
Thanks, I just added some minor comments.

widget/embedded/src/pages/WalletsPage.tsx
mikasackermn: this line of code was added to solve an issue where wallet skeleton isn't displayed and meta wasn't loaded yet.We also use filterByType for the blockchain page

widget/ui/src/components/Wallet/Wallet.types.ts
mikasackermn: I disagree because it doesn't give the meaning of transaction type here

RyukTheCoder (approved):
LGTM

PR #720, opened 2024-05-12
fix: excluding ledger on mobile

mikasackermn:
# Summary

Shouldn't show Ledger in mobile view.
The ledger was causing problems for the Injected wallet
NOTE: Injected wallet for when we don't support an evm wallet and the user wants to connect with that wallet

Fixes # (issue)

Test on mobile display

For instance, you could try removing Metamask from the provider all and connecting to Metamask using the injected wallet to test the injected wallet.

# Checklist:

### DIFF
--- a/widget/embedded/src/hooks/useWalletList.ts
+++ b/widget/embedded/src/hooks/useWalletList.ts
@@ -56,7 +56,8 @@ export function useWalletList(params: Params) {
 
   wallets = detectMobileScreens()
     ? wallets.filter(
-        (wallet) => wallet.showOnMobile || state(wallet.type).installed
+        (wallet) =>
+          wallet.showOnMobile !== false && state(wallet.type).installed
       )
     : wallets;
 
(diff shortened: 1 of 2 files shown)


Ikari-Shinji-re (approved):

widget/embedded/src/hooks/useWalletList.ts
```diff
@@ -55,9 +55,12 @@ export function useWalletList(params: Params) {
...
 
   wallets = detectMobileScreens()
-    ? wallets.filter(
-        (wallet) => wallet.showOnMobile || state(wallet.type).installed
-      )
+    ? wallets.filter((wallet) => {
+        if (wallet.showOnMobile === false) {
+          return false;
+        }
+        return wallet.showOnMobile || state(wallet.type).installed;
```
nikaaru, line 62: Can we remove `wallet.showOnMobile` in ` wallet.showOnMobile || state(wallet.type).installed` condition?

because in prev condition, `wallet.showOnMobile` was checked.

mikasackermn: The issue is that although showOnMobile might include the true value, but it isn't installed.

nikaaru: in this condition: `wallet.showOnMobile || state(wallet.type).installed`,
`wallet.showOnMobile `  is always `true`,
so we can remove it or combine the two conditions as follows:

`wallets.filter(wallet => wallet.showOnMobile !== false && state(wallet.type).installed`

nikaaru (approved):

PR #703, opened 2024-04-29
chore: add analyze bundle workflow

Ikari-Shinji-re:
# Summary

Include a workflow check to analyze the widget bundle.

# How did you test this change?

Utilize a package such as Material UI within the widget and submit your changes to this pull request. The workflow will execute and display the alteration in the widget bundle's size.

# Checklist:

- [x] I have performed a self-review of my code

PR #725, opened 2024-05-14
test analyze bundle workflow

Ikari-Shinji-re:
related to: https://github.com/rango-exchange/rango-client/pull/703

### DIFF
--- a/.github/workflows/checks.yaml
+++ b/.github/workflows/checks.yaml
@@ -24,6 +24,22 @@ jobs:
           node ./scripts/build-libs/command.mjs
           yarn run test
 
+      - name: Hash Current Branch Name
+        id: hash_curret_branch_name
+        run: |
+          hashed_name=$(printf "%s" "${{ github.ref }}" | base64)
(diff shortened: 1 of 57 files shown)


.github/workflows/checks.yaml
```diff
@@ -86,3 +96,27 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+          ref: ${{ github.base_ref }}
+
+      - name: Prepare
+        uses: ./.github/actions/prepare
+
+      - name: Build Target Branch
```
yeager-eren, line 112: we should download artifact for target branch as well, if it doesn't exists, this action should be skipped.

yeager-eren (changes requested):
Thanks. I put some general points here, but this changnes should be moved to your original PR for analyzing. You can check more details out [here](https://github.com/rango-exchange/rango-client/pull/703#pullrequestreview-2064602200).

PR #733, opened 2024-05-19
feat: add custom solana rpc url to config

RyukTheCoder:
# Summary

Added custom solana rpc in signer config to widget config.

Fixes # RF-1504

Tested it by adding the below object to widget config:

```
signerConfig: { customSolanaRPC: 'customSolanaRPC' }
```

# Checklist:

### DIFF
--- a/signers/signer-solana/src/config.ts
+++ b/signers/signer-solana/src/config.ts
@@ -1,0 +1,20 @@
+type SolanaSignerConfig = {
+  customRPC?: string;
+};
+
+const solanaSignerConfig: SolanaSignerConfig = {};
+
+export function setSolanaSignerConfig<Key extends keyof SolanaSignerConfig>(
+  key: Key,
+  value: SolanaSignerConfig[Key]
+) {
+  solanaSignerConfig[key] = value;
+
+  return value;
+}
+
+export function getSolanaSignerConfig<Key extends keyof SolanaSignerConfig>(
+  key: Key
+): SolanaSignerConfig[Key] {
+  return solanaSignerConfig[key];
+}
--- a/widget/embedded/src/containers/WidgetProvider/WidgetProvider.tsx
+++ b/widget/embedded/src/containers/WidgetProvider/WidgetProvider.tsx
@@ -1,7 +1,8 @@
 import type { PropTypes } from './WidgetProvider.types';
 import type { PropsWithChildren } from 'react';
 
-import React, { useEffect, useMemo } from 'react';
+import { setSolanaSignerConfig } from '@rango-dev/signer-solana';
+import React, { useEffect } from 'react';
 
 import { DEFAULT_BASE_URL, RANGO_PUBLIC_API_KEY } from '../../constants';
 import useFontLoader from '../../hooks/useFontLoader';
@@ -13,13 +14,6 @@ import { WidgetInfo } from '../WidgetInfo';
 export function WidgetProvider(props: PropsWithChildren<PropTypes>) {
   const { onUpdateState, config } = props;
 
-  useMemo(() => {
-    initConfig({
-      API_KEY: config?.apiKey || RANGO_PUBLIC_API_KEY,
-      BASE_URL: config?.apiUrl || DEFAULT_BASE_URL,
-    });
-  }, [config.apiKey, config.apiUrl]);
-
   const fontFamily = props.config?.theme?.fontFamily;
 
   const { handleLoadCustomFont } = useFontLoader();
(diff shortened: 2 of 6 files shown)


signers/signer-solana/src/config.ts
```diff
@@ -0,0 +1,13 @@
+type Config = {
+  customRPC: string;
+};
+
+let config: Config | undefined;
+
+export function setConfig(c: { customRPC: string }) {
```
yeager-eren, line 7: please use full names. and use `Config` instead of inline typing.

signers/signer-solana/src/config.ts
```diff
@@ -0,0 +1,13 @@
...
+  config = c;
+}
+
+export function getConfig() {
```
yeager-eren, line 11: you can accept a param here and also return a proper type and infer the type using generics.

widget/embedded/src/containers/WidgetProvider/WidgetProvider.tsx
```diff
@@ -12,13 +13,21 @@ import { WidgetInfo } from '../WidgetInfo';
 export function WidgetProvider(props: PropsWithChildren<PropTypes>) {
   const { onUpdateState, config } = props;
 
-  useMemo(() => {
+  useEffect(() => {
```
yeager-eren, line 16: Why you've changed memo to effect?

widget/embedded/src/types/config.ts
```diff
@@ -209,4 +213,5 @@ export type WidgetConfig = {
   features?: Features;
   variant?: WidgetVariant;
   enableCentralizedSwappers?: boolean;
+  signerConfig?: SignerConfig;
```
yeager-eren, line 216: It's already in a `config` so adding a suffix is not needed. I mean you can rename `signerConfig` to `signers` or whatever makes more sense to you.

yeager-eren (changes requested):
Thanks. added some comments.

signers/signer-solana/src/config.ts
yeager-eren, line 7: Please consider to always type function's return. (helps performance)

widget/embedded/src/containers/WidgetProvider/WidgetProvider.tsx
RyukTheCoder: To improve the code quality. That was a wrong way of using 'useMemo'.

signers/signer-solana/src/config.ts
RyukTheCoder: Tnx, Added.

signers/signer-solana/src/config.ts
RyukTheCoder: I don't get why a getter function should accept a param.

yeager-eren:
```js

function getConfig(name){
// ...
}

const rpc = getConfig("customRPC"); // result: string | undefined

// without param
const rpc = getConfig()?.customRPC;
```

yeager-eren: And for consistency, we implemented same thing in other places like `widget/embedded/src/utils/configs.ts`

widget/embedded/src/containers/WidgetProvider/WidgetProvider.tsx
yeager-eren: They are running in 2 different phases, one in render and effect after a render in commit phase. This could break the behavior here. Did you check it?
Although I agree useMemo is for caching a result so maybe it's a misuse here.

signers/signer-solana/src/config.ts
RyukTheCoder: Fixed

widget/embedded/src/containers/WidgetProvider/WidgetProvider.tsx
RyukTheCoder: Yes I think it is a misuse of useMemo and it should be corrected. Also the effect of 'initConfig' with apiKey and apiUrl is not necessary for render phase and it can be happen after the render. The only unintended behavior can be api calls with the wrong apiKey and apiUrl. I tested that and didn't detect this unintended behavior. Considering this, also I don't think we need to use some alternative like useLayoutEffect here.

signers/signer-solana/src/config.ts
```diff
@@ -0,0 +1,18 @@
+type SolanaSignerConfig = {
+  customRPC?: string;
+};
+
+const solanaSignerConfig: SolanaSignerConfig = {};
+
+export function setSolanaSignerConfig(
+  name: keyof SolanaSignerConfig,
+  value: any
```
yeager-eren, line 9: it shouldn't be `any`. you can infer `value` type by `name`

widget/embedded/src/containers/WidgetProvider/WidgetProvider.tsx
yeager-eren: I will leave this to @Ikari-Shinji-re to check and resolve this comment.

yeager-eren (changes requested):

widget/embedded/src/containers/WidgetProvider/WidgetProvider.tsx
RyukTheCoder: Considering that 'API_KEY' and 'BASE_URL' are not needed during the render phase, and no memoization is happened here I think it wont have any side effect. check please @Ikari-Shinji-re

yeager-eren (approved):
thanks. lgtm

Ikari-Shinji-re (approved):

PR #736, opened 2024-05-20
feat: update our colors to new colors system

mikasackermn:
# Summary

The colors should to be changed in accordance with Figma.
Additionally, colors can be generated using the tool we provided.

Note: This new formula is used for blockchain tags in the token list.

Fixes # (issue)

Playground's styles tab allows you to test.

# Checklist:

- [x] Update colors
- [x] Check our colors to be matched with Figma
- [x] Update our color generator to use new forrmula

### DIFF
--- a/widget/embedded/src/components/TokenList/TokenList.tsx
+++ b/widget/embedded/src/components/TokenList/TokenList.tsx
@@ -1,6 +1,6 @@
 /* eslint-disable @typescript-eslint/no-magic-numbers */
 import type { PropTypes, RenderDescProps } from './TokenList.types';
-import type { Token } from 'rango-sdk';
+import type { BlockchainMeta, Token } from 'rango-sdk';
 
 import { i18n } from '@lingui/core';
 import {
@@ -19,7 +19,7 @@ import React, { useEffect, useState } from 'react';
 
 import { useAppStore } from '../../store/AppStore';
 import { useWalletsStore } from '../../store/wallets';
-import { generateShade, isBright } from '../../utils/colors';
+import { createTintsAndShades } from '../../utils/colors';
 import { formatBalance } from '../../utils/wallets';
 
 import { LoadingTokenList } from './LoadingTokenList';
@@ -133,25 +133,20 @@ export function TokenList(props: PropTypes) {
           const address = token.address || '';
           const blockchain = blockchains.find(
             (blockchain) => blockchain.name === token.blockchain
-          );
-          const lightShade = generateShade(
-            blockchain?.color || '',
-            isBright(blockchain?.color || '') ? 5 : 125
-          );
-          const darkShade = generateShade(blockchain?.color || '', -100);
-
+          ) as BlockchainMeta;
+          const colors = createTintsAndShades(blockchain.color, 'main');
           const customCssForTag = {
-            $$color: lightShade,
+            $$color: colors.main250,
             [`.${darkTheme} &`]: {
-              $$color: darkShade,
+              $$color: colors.main850,
             },
             backgroundColor: '$$color',
           };
 
           const customCssForTagTitle = {
-            $$color: darkShade,
+            $$color: colors.main600,
             [`.${darkTheme} &`]: {
-              $$color: lightShade,
+              $$color: colors.main200,
             },
             color: '$$color',
           };
--- a/widget/embedded/src/utils/colors.ts
+++ b/widget/embedded/src/utils/colors.ts
@@ -1,141 +1,184 @@
-import type { WidgetColors, WidgetColorsKeys } from '../types';
+/* eslint-disable @typescript-eslint/no-magic-numbers */
+// Types
+import type { ThemeColors, WidgetColors, WidgetColorsKeys } from '../types';
 
-export function generateShades(
-  baseColor: string,
-  prefix: string,
-  reverse?: boolean
-): { [key: string]: string } {
-  const BASE_SHADE_INDEX = 100;
-  const NUMBER_OF_COLORS = 9;
-  const SHADING_INCREMENT_MAX = 10; // For bright colors
-  const SHADING_INCREMENT_MIN = 6; // For dark colors
-
-  const HALF_NUMBER_OF_COLORS = 4;
-
-  const shadeIncrements: number[] = [];
-  let increment = 90;
-  for (let i = 0; i < NUMBER_OF_COLORS; i++) {
-    shadeIncrements.push(increment);
-    if (i + 1 === HALF_NUMBER_OF_COLORS) {
-      increment = 0;
-    } else {
-      increment -=
-        i >= HALF_NUMBER_OF_COLORS
-          ? SHADING_INCREMENT_MIN
-          : SHADING_INCREMENT_MAX;
-    }
-  }
-
-  // If 'reverse' is true, that means the color should be from dark to light
-  const decrementAmounts = reverse
-    ? shadeIncrements.slice().reverse()
-    : shadeIncrements;
-
-  const shades: { [key: string]: string } = {};
-
-  // Generate shades
-  for (let i = 0; i < NUMBER_OF_COLORS; i++) {
-    const shadeIndex = BASE_SHADE_INDEX + i * BASE_SHADE_INDEX;
-    const decrementAmount = decrementAmounts[i];
-
-    const shade: { [key: string]: string } = {
-      [`${prefix}${shadeIndex}`]: generateShade(baseColor, decrementAmount),
-    };
-    Object.assign(shades, shade);
-  }
-
-  return { ...shades, [prefix]: baseColor };
-}
-
-export function generateShade(
-  baseColor: string,
-  decrementAmount: number
-): string {
-  const [r, g, b] = convertToRGB(baseColor);
-  return `#${clamp(r + decrementAmount)}${clamp(g + decrementAmount)}${clamp(
-    b + decrementAmount
-  )}`;
-}
-
-export function convertToRGB(baseColor: string): number[] {
-  const COLOR_SLICE_START = 1;
-  const COLOR_SLICE_END = 3;
-  const BASE_COLOR_LENGTH = 7;
-
-  // Convert the base color to RGB components
-  const r: number = parseInt(
-    baseColor.slice(COLOR_SLICE_START, COLOR_SLICE_END),
-    16
-  );
-  const g: number = parseInt(
-    baseColor.slice(COLOR_SLICE_END, COLOR_SLICE_END + 2),
-    16
-  );
-  const b: number = parseInt(
-    baseColor.slice(COLOR_SLICE_END + 2, BASE_COLOR_LENGTH),
-    16
-  );
-
-  return [r, g, b];
-}
-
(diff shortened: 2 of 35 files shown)


widget/embedded/src/utils/colors.ts
```diff
@@ -1,141 +1,151 @@
...
+// Generate a color palette based on main colors and widget colors
+export function generateColors(
+  mainColors: { [key: string]: string },
   reverseNeutralRange: boolean,
-  colors?: WidgetColors
-) => {
-  if (!colors || !Object.entries(colors).length) {
+  widgetColors?: WidgetColors
+): { [key: string]: string } {
+  if (!widgetColors || Object.keys(widgetColors).length === 0) {
```
yeager-eren, line 124: I think it's kinda weird to make `widgetColors` optional then add a check here to if it's not passed return nothing.
You can make `widgetColors` mandatory (type) and remove this if block here. you will need to update where `generateColors()` has used probably which is ok.

widget/embedded/src/utils/colors.ts
```diff
@@ -1,141 +1,151 @@
...
+  for (let i = 0; i < length; i++) {
+    const index = BASE_INDEX + (reverse ? length - 1 - i : i) * STEP;
+    colorMap[`${prefix}${index}`] = combinedColors[i];
+  }
+
+  return colorMap;
+}
+
+// Generate a color palette based on main colors and widget colors
+export function generateColors(
```
yeager-eren, line 119: Let's simplify this function and make it more clear:
1. Rename the function from `generateColros` to `extendThemeColors`.
2. Remove `shouldUpdate` logic and this function always should return colors.
3. As I said in a separate comment, `widgetColors` is mandatory for this function. Also rename the param to `extendedColors`.
4. Rename `mainColors` to `baseColors`.
4. Reorder params to be matched with this final function signature: `extendThemeColors(baseColors, extendedColors, options: { reverseNeutralRange  })`

yeager-eren: The final logic will only gets a base colors and will extend the base colors using second parameter (`extendedColors`) and always will return an object includes colors (so no empty object `{}`).

yeager-eren (changes requested):
Thanks, I've put some comments that needs to be addressed.

yeager-eren (changes requested):

widget/embedded/src/components/TokenList/TokenList.tsx
```diff
@@ -133,24 +133,24 @@ export function TokenList(props: PropTypes) {
           const address = token.address || '';
           const blockchain = blockchains.find(
             (blockchain) => blockchain.name === token.blockchain
+          ) as BlockchainMeta;
+          const colors = createTintsAndShades(
+            blockchain.color,
+            blockchain.shortName,
```
yeager-eren, line 139: you don't need to pass `shortName` here. you can use a fixed name like `main` or `primary`. it helps you to change `$$color: colors[`${blockchain.shortName}250`],` to `$$color: colors.main250`

widget/embedded/src/utils/ui.ts
```diff
@@ -28,47 +28,53 @@ export function customizedThemeTokens(
...
-  let dark: CustomizedTheme['dark'] = undefined;
-
-  if (hasDefaultLightColors) {
-    const tokens = { colors: lightColorsWithDefaults };
+  const baseDarkColors = {
+    ...baseColors,
+    ...defaultDarkColors,
+  };
+
+  const darkColorChanges = !!colors?.dark
```
yeager-eren, line 36: I suggest to change the code structure to something like this (I only do for `light`, `dark` will be same.).

```ts
if(color.light){

   const expandColors = findChangedColors(baseColors, colors.light);   
   if(expandColors){
      const lightColors = expandedThemeColors(baseColors, expandColors);
      const tokens = { colors: lightColors };
      // ...
   }
   
}
```

widget/embedded/src/utils/colors.ts
```diff
@@ -1,141 +1,185 @@
...
+type ShadeOrTintFunction = (rgb: RGB, step: number) => RGB;
+
+const HEX_BASE = 16;
+const MAX_COLOR_VALUE = 255;
+const PERCENT = 0.11;
+
+// Utility Functions
+
+// Determines if there are color changes between base and expanded colors.
+export function findChangedColors(
```
yeager-eren, line 25:
```
/*
We only need to expand those colors that doesn't match with baseColors.
Our default theme colors don't follow our formula for generating colors and actually they are custom colors. By checking `baseColors` and `expandColors` we will remove those colors that are default colors to avoid generating colors for them. 
*/

function removeDefaultColorsFromExpand(...){
// ...
}
```

yeager-eren: Please rename the function and add the comment.

widget/embedded/src/utils/colors.ts
```diff
@@ -1,141 +1,185 @@
...
-          ...listOfColors,
-          ...generateShades(color, colorKey, reverse),
-        };
-      }
+
+  return colorMap;
+}
+
+// Generate a color palette based on base colors and "Expand Colors"
+export function expandedThemeColors(
```
yeager-eren, line 138: I think `expandToGenerateThemeColors` is more meaningful. Sorry for keep changing this function name for multiple times, trying to find the best name to describe what it's doing.

yeager-eren (changes requested):

widget/embedded/src/utils/ui.ts
```diff
@@ -27,48 +30,54 @@ export function customizedThemeTokens(
...
-
-  if (hasDefaultLightColors) {
-    const tokens = { colors: lightColorsWithDefaults };
-    const id = `${THEME_CLASS_NAME_PREFIX}-light-${toHash(tokens)}`;
-    light = {
-      id,
-      tokens,
-    };
+  if (colors?.light) {
+    const expandColors = removeDefaultColorsFromExpand(
```
yeager-eren, line 41: As we talked we can remove this step and move it to playground somehow,

yeager-eren (changes requested):

widget/embedded/src/utils/colors.ts
```diff
@@ -1,141 +1,173 @@
...
-          ...generateShades(color, colorKey, reverse),
-        };
-      }
+
+  return colorMap;
+}
+
+// Generate a color palette based on base colors and "Expand Colors"
+export function expandToGenerateThemeColors(
+  baseColors: { [key: string]: string },
```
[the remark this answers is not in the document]
mikasackermn: This function has been removed

widget/playground/src/constants/presets.ts
```diff
@@ -281,7 +164,8 @@ export const WIDGET_COLORS: { key: WidgetColorsKeys; label: string }[] = [
     label: 'Foreground',
   },
 ];
-export const DEFAULT_COLORS = {
+
+export const DEFAULT_THEME = {
```
yeager-eren, line 168: theme includes other properties like `spacing`, ... . So please rename this to `DEFAULT_THEME_COLORS`

widget/ui/src/theme.ts
```diff
@@ -228,3 +225,16 @@ export const darkTheme = createTheme('dark-theme-ui', {
     s: '0px 3px 5px 3px #222, 0px 6px 10px 3px #222, 0px 1px 18px 3px #222',
   },
 });
+
+export const RangoDarkColors = {
```
yeager-eren, line 229: Please follow other naming convention in this file. it should be `rangoDarkColors`

widget/playground/src/types.ts
```diff
@@ -1 +1,10 @@
+import type { WidgetColors } from '@rango-dev/widget-embedded';
+
 export type Type = 'Destination' | 'Source';
+
+export type COLORS = { light?: WidgetColors; dark?: WidgetColors };
```
yeager-eren, line 5: We are using UPPERCASE for constants, not types.

yeager-eren: Same applies to `PRESET`

widget/playground/src/utils/colors.ts
```diff
@@ -1,25 +1,76 @@
...
     }
-    if (isLightTab && theme.singleTheme && theme?.colors?.light) {
+    if (isLightTab && theme?.singleTheme && theme?.colors?.light) {
       return theme?.colors?.light[key];
     }
   }
   return undefined;
 };
+
+export const isSelectedPreset = (
```
yeager-eren, line 34:
```suggestion
export const isPresetSelected = (
```

yeager-eren (changes requested):

yeager-eren (approved):
I only did the code review and didn't test it. We can pass this to QA and also second reviewer to make sure everything looks good.

widget/embedded/src/utils/colors.ts
```diff
@@ -1,141 +1,159 @@
...
+function rgbToHex(rgb: RGB) {
+  return `#${intToHex(rgb.red)}${intToHex(rgb.green)}${intToHex(rgb.blue)}`;
 }
 
-export function convertToRGB(baseColor: string): number[] {
-  const COLOR_SLICE_START = 1;
-  const COLOR_SLICE_END = 3;
-  const BASE_COLOR_LENGTH = 7;
+// Convert a hex color string to an RGB object
+function hexToRGB(hex: string): RGB {
```
RyukTheCoder, line 48: You are handling 6 digit hex colors but 3 digit hex colors ae not handled here. Therefore, if the secondary color is set to something like `#129`, it won't work properly.

RyukTheCoder (changes requested):
Thanks, I just added a comment.

Also in the background color of playground will not be updated by changing the theme. Is that the required behavior?

widget/playground/src/containers/StyleLayout/StyleLayout.Preset.tsx
```diff
@@ -78,15 +77,15 @@ export function Preset(props: PresetTypes) {
...
-    light: WidgetColors;
-  }) => {
-    onSelectTheme(customTheme);
-    setSelectedPreset(customTheme);
+  const onSelectPreset = (customTheme: ColorsType) => {
+    const selectedTheme = isAutoTab ? customTheme : { [tab]: customTheme[tab] };
+    onSelectTheme(selectedTheme);
+    setSelectedPreset(selectedTheme);
     onChangeTheme({ name: 'singleTheme', value: isAutoTab ? undefined : true });
     if (!isAutoTab) {
```
RyukTheCoder, line 85: This if and else if pointless

```suggestion
onChangeTheme({ name: 'mode', value: tab });
```

RyukTheCoder (changes requested):

mikasackermn:
> Thanks, I just added a comment.
> 
> Also in the background color of playground will not be updated by changing the theme by selecting from presets in dark and light tab. I think you set wrong background and foreground colors for some presets.

This is because the default theme is being read and we haven't modified the neutral colors yet.

RyukTheCoder (approved):

RanGojo (changes requested):
I've reported the bug offline.

RanGojo (approved):

PR #737, opened 2024-05-25
fix: exclude the gas price from evm transaction specifications in version 2

Ikari-Shinji-re:
# Summary

For EVM transaction v2, the `gasPrice` should be omitted, and only the `maxFeePerGas` and `maxPriorityFeePerGas` should be provided to the provider for signing the transaction.

# Checklist:

### DIFF
--- a/signers/signer-evm/src/signer.ts
+++ b/signers/signer-evm/src/signer.ts
@@ -1,16 +1,18 @@
-import {
+import type {
   TransactionRequest,
   TransactionResponse,
 } from '@ethersproject/abstract-provider';
 import type { GenericSigner } from 'rango-types';
-import { EvmTransaction } from 'rango-types/lib/api/main';
+import type { EvmTransaction } from 'rango-types/lib/api/main';
+
 import { providers } from 'ethers';
-import { cleanEvmError, getTenderlyError, waitMs } from './helper';
 import {
   RPCErrorCode as RangoRPCErrorCode,
   SignerError,
   SignerErrorCode,
 } from 'rango-types';
+
+import { cleanEvmError, getTenderlyError, waitMs } from './helper';
 import { RPCErrorCode } from './types';
 
 type ProviderType = ConstructorParameters<typeof providers.Web3Provider>[0];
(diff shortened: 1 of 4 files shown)


signers/signer-evm/src/signer.ts
```diff
@@ -53,6 +61,42 @@ export class DefaultEvmSigner implements GenericSigner<EvmTransaction> {
...
+        ...tx,
+        gasPrice: '0x' + parseInt(evmTx.gasPrice).toString(TO_STRING_BASE),
+      };
+    }
+    if (evmTx.maxFeePerGas && !disableV2) {
+      tx = { ...tx, maxFeePerGas: evmTx.maxFeePerGas };
+    }
+    if (evmTx.maxPriorityFeePerGas && !disableV2) {
+      tx = { ...tx, maxPriorityFeePerGas: evmTx.maxPriorityFeePerGas };
+    }
```
RanGojo, line 96: This is the correct logic:

```suggestion
    if (!disableV2 && evmTx.maxFeePerGas && evmTx.maxPriorityFeePerGas) {
      tx = { 
        ...tx, 
        maxFeePerGas: evmTx.maxFeePerGas, 
        maxPriorityFeePerGas: evmTx.maxPriorityFeePerGas 
      };
    } else if (evmTx.gasPrice) {
      tx = {
        ...tx,
        gasPrice: '0x' + parseInt(evmTx.gasPrice).toString(TO_STRING_BASE),
      };
    }
```

signers/signer-evm/src/signer.ts
```diff
@@ -1,18 +1,21 @@
...
 import {
   RPCErrorCode as RangoRPCErrorCode,
   SignerError,
   SignerErrorCode,
 } from 'rango-types';
+
+import { cleanEvmError, getTenderlyError, waitMs } from './helper';
 import { RPCErrorCode } from './types';
 
+// eslint-disable-next-line @typescript-eslint/no-magic-numbers
```
RanGojo, line 18: Why did you add this? 🤔

RanGojo (approved):

PR #745, opened 2024-05-27
chore: eliminate any newline characters from the artifact names

Ikari-Shinji-re:
# Summary

eliminate any newline characters from the artifact names


yeager-eren (approved):
lgtm

PR #746, opened 2024-05-27
 fix: some bugs in playground

mikasackermn:
# Summary

Fix walletConnectProjectId:

Don't insert 'walletConnectProjectId' in exported code if wallet-connect is excludedCurrently if the user exclude 'wallet-connect' in the list of wallets, we insert walletConnectProjectId in the exported code anyway. It should be removed if 'wallet-connect' is excluded.

Fix external wallets and supported wallets mismatch:

Reproduce

- Enable external wallet option
- Click on supported wallets (metamask is selected by default), Click on Select All and then Confirm.
- Click on export code button, it eliminated wallets and only keeps externalWallets: true.

Possible solution:
When external wallets has been enabled, we can disable Supported Wallets input to stop user from doing incorrect behaviour.
fix Enable Multi Wallets Simultaneously

Reproduce:
- Disable and enable  Enable Multi Wallets Simultaneously option.
- It will add multiWallets: true,

Solution:
When the option has been enabled, we can eliminate multiWallets: true, from exported config.

Fix slider bug:

- When the config was reset, the slider wasn't reset
- The color of the slider wasn't changed with the theme

Fixes # (issue)

You can test on the playground

# Checklist:

### DIFF
--- a/widget/playground/src/components/Slider/Slider.tsx
+++ b/widget/playground/src/components/Slider/Slider.tsx
@@ -3,6 +3,8 @@ import type { PropTypes } from './Slider.types';
 import { Typography } from '@rango-dev/ui';
 import React, { useEffect } from 'react';
 
+import { DEFAULT_THEME_COLORS, PLAYGROUND_CONTAINER_ID } from '../../constants';
+
 import {
   Content,
   RangeWrapper,
@@ -11,6 +13,7 @@ import {
 } from './Slider.styles';
 
 const MAX_VALUE = 100;
+const DEFAULT_COLOR = DEFAULT_THEME_COLORS.light.secondary;
 function Slider(props: PropTypes) {
   const {
     title,
@@ -20,6 +23,7 @@ function Slider(props: PropTypes) {
     variant = 'custom',
     min,
     max = MAX_VALUE,
+    color = DEFAULT_COLOR,
     id,
   } = props;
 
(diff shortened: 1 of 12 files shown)


RyukTheCoder:
Thanks. There is just a minor bug in sliders which can be observed in this video


[screen-capture.webm](https://github.com/rango-exchange/rango-client/assets/154742563/7ecda726-1520-48ad-9374-b477e823bff1)


Also the commit message should be corrected.

widget/playground/src/components/Slider/Slider.tsx
```diff
@@ -20,19 +22,28 @@ function Slider(props: PropTypes) {
...
   const progressScript = () => {
     const sliderEl = document.querySelector(`#${id}`) as HTMLInputElement;
     const sliderValue = parseInt(sliderEl.value);
     const mainValue = sliderValue * (MAX_VALUE / (max as number));
-    sliderEl.style.background = `linear-gradient(to right, #5BABFF ${mainValue}%, #C8E2FF ${mainValue}%)`;
+    // Get CSS variables
+    const referenceElement = document.querySelector(
+      `#${PLAYGROUND_CONTAINER_ID}`
+    ) as Element;
+    const sliderColorActive = color;
```
RyukTheCoder, line 37:
```suggestion
    const sliderActiveColor = color;
```

widget/playground/src/containers/StyleLayout/StyleLayout.General.tsx
```diff
@@ -139,6 +142,7 @@ export function General() {
           </FieldTitle>
           <Divider size={16} />
           <Slider
+            color={colors?.[mode]?.secondary || '#2284ED'}
```
RyukTheCoder, line 145: I think Slider component should have a default color instead of passing a hex code color if 'secondary' is undefined.

RyukTheCoder (changes requested):
Thanks, I just added some minor comments.

widget/playground/src/components/Slider/Slider.tsx
```diff
@@ -13,6 +13,7 @@ import {
 } from './Slider.styles';
 
 const MAX_VALUE = 100;
+const DEFAULT_COLOR = '#2284ED';
```
RyukTheCoder, line 16: considering that this value should be equal to the secondary color of 'DEFAULT_THEME_COLORS', please check if you can use it directly.

RyukTheCoder (changes requested):

widget/playground/src/components/Slider/Slider.tsx
mikasackermn: fixed

RyukTheCoder (approved):
LGTM

Ikari-Shinji-re:
when I exclude Wallet Connect in Playground, this happens:

Ikari-Shinji-re:
when the wallets config is not touched and I enable external wallets, this happens:

Ikari-Shinji-re:
when some wallets are selected and I enable the external wallets, the exported code is correct, but the list of wallets in the widget doesn't match the selected wallets:

mikasackermn:
> when some wallets are selected and I enable the external wallets, the exported code is correct, but the list of wallets in the widget doesn't match the selected wallets:  

All wallets should be included in the Playground, and since we utilized the Playground widget provider when we choose the external wallet option, we are unable to override wallets.

PR #748, opened 2024-05-27
feat: make update settings optional to make it enable in playground

mikasackermn:
# Summary
Source of truth of the values should be config object in Playground, but if any changes happens inside the widget, it has more priority than config. We need to reset all the widget values in Playground on reload to make sure configs will be applied.

Note: It seemed that the problem was from Zustand's persistor, but the problem was that the setting wasn't updated by the config

Fixes # (issue)

Change the font in the playground first, then the widget. After changing the font once more in the playground, the widget's font should also change.

# Checklist:

### DIFF
--- a/widget/embedded/src/types/config.ts
+++ b/widget/embedded/src/types/config.ts
@@ -219,4 +219,9 @@ export type WidgetConfig = {
   __UNSTABLE_OR_INTERNAL__?: {
     walletConnectListedDesktopWalletLink?: string;
   };
+
+  // Internal configuration options. Not intended for public use or modification.
+  _INTERNAL_SETTINGS_?: {
+    autoUpdateSettings?: boolean; // If true, settings will be updated automatically based on the configuration.
+  };
 };
--- a/widget/embedded/src/store/slices/settings.ts
+++ b/widget/embedded/src/store/slices/settings.ts
@@ -169,6 +169,9 @@ export const createSettingsSlice: StateCreator<
       ...(isLiquidityHidden && {
         disabledLiquiditySources: nextConfig.liquiditySources || [],
       }),
+      ...(nextConfig?._INTERNAL_SETTINGS_?.autoUpdateSettings && {
+        language: nextConfig.language || DEFAULT_LANGUAGE,
+      }),
     });
   },
 });
--- a/widget/playground/src/App.tsx
+++ b/widget/playground/src/App.tsx
@@ -23,6 +23,9 @@ export function App() {
     features: {
       theme: 'hidden',
     },
+    _INTERNAL_SETTINGS_: {
+      autoUpdateSettings: true,
+    },
   };
 
   /*


widget/embedded/src/types/config.ts
```diff
@@ -219,4 +219,9 @@ export type WidgetConfig = {
   __UNSTABLE_OR_INTERNAL__?: {
     walletConnectListedDesktopWalletLink?: string;
   };
+
+  // Note: It is advised against adding this to the configuration.
+  __INTERNAL_DONT_USE__?: {
```
nikaaru, line 224: NIT: I think it is better to use a more appropriate and clear name and comment.

suggestion:

```
// Internal configuration options. Not intended for public use or modification.
_internalSettings?: {
  autoUpdateSettings?: boolean; // If true, settings will be updated automatically based on the configuration.
};
```

nikaaru (approved):
LGTM

widget/embedded/src/types/config.ts
```diff
@@ -220,8 +220,8 @@ export type WidgetConfig = {
     walletConnectListedDesktopWalletLink?: string;
   };
 
-  // Note: It is advised against adding this to the configuration.
-  __INTERNAL_DONT_USE__?: {
-    shouldUpdateSettingsWithConfig?: boolean;
+  // Internal configuration options. Not intended for public use or modification.
+  _INTERNAL_SETTINGS_?: {
+    autoUpdateSettings?: boolean; // If true, settings will be updated automatically based on the configuration. };
```
nikaaru, line 225: please remove `};` at the end of this line.

nikaaru (approved):
LGTM

RyukTheCoder:
The description of the task says that you should make the persisting of data in zustand optional, but here you made the language in zustand synced with the config. Can you please explain how this change can solve that problem? Does this should only be done for the language and not for other settings? 

Also I see a bug which is represeted in the below video. I don't know if it should be fixed in this PR.

[screen-capture (2).webm](https://github.com/rango-exchange/rango-client/assets/154742563/3441137c-9b4c-411c-b210-df13ca6aff16)

mikasackermn:
> The description of the task says that you should make the persisting of data in zustand optional, but here you made the language in zustand synced with the config. Can you please explain how this change can solve that problem? Does this should only be done for the language and not for other settings?
> 
> Also I see a bug which is represeted in the below video. I don't know if it should be fixed in this PR.
> 
>  screen-capture.2.webm

Becuse it seemed that the problem was from Zustand's persistor, but the problem was that the setting wasn't updated by the config
Currently, we can only modify the language by modifying the configuration using Polygrand; but, in the future, we will be able to add other elements as well.
Since we just set the default language in Polygrand, the video you submitted functions as intended.

RyukTheCoder (approved):
LGTM

PR #762, opened 2024-06-08
chore: add log for debugging wallet connect signer errors

RanGojo:
# Summary

We are experiencing some unknown errors with the Wallet Connect signer. In this PR, I've added a warning log for debugging these errors, which will help us identify issues by sending pairing and session namespaces to the sentry.

By testing a transaction on wallet connect (and rejecting it).

# Checklist:


wallets/provider-walletconnect-2/src/signers/evm.ts
```diff
@@ -126,7 +91,25 @@ class EVMSigner implements GenericSigner<EvmTransaction> {
         hash,
       };
     } catch (error) {
-      throw cleanEvmError(error);
+      const modifiedError = cleanEvmError(error);
```
yeager-eren, line 94: I suggest a different approach here since with this logic events can be sent twice, one from here and another one from `queue-manager/rango-preset/src/helpers.ts` and `signTransaction`.

What I suggest is creating a custom error class and let providers to only more context to it. We can put the class in `wallets-shared` to be accessible for all the providers (includes WC2).

Here is what I think of:


```typescript
// somwhere in wallets-shared.
export class ProviderError extends Error {
    name = 'ProviderError';
    context = {}
    constructor(msg, context) {
        super(msg);

        this.context = context;
    }
}


// current file (wallets/provider-walletconnect-2/src/signers/evm.ts)

throw new ProviderError(msg, {
    namspaces: session?.namespaces,
    peering: session?.peer?.metadata,
})

// queue-manager/rango-preset/src/helpers.ts

// Logging RPC errors
if (error instanceof ProviderError) {
    warn(error, {
        context: error.context,
        tags: ...
    })
} else {
    const log_msg = new RangoPresetError(
        error?.trace?.stack || error?.trace || error?.root || error
    );

    // ...
}


```

yeager-eren: I also suggest to not using `cleanEvmError` here, since it can eliminate some useful context like error code. For example the map in that function for user is rejecting the action is `4001` (which is for Metamask) but for WC is `5000`.

My final suggestion would be something like this:



```typescript
const msg = error instanceof Error ? error.message : YOUR_JSON_STRINGIFY;
// or if we tested and errors are not instance of Error, we can use `error.message || YOUR_JSON_STRINGIFY`
 

const context = {
    code: error.code
    namspaces: session?.namespaces,
    peering: session?.peer?.metadata,
}


throw new ProviderError(msg, context)
```

yeager-eren (changes requested):

wallets/provider-walletconnect-2/src/signers/evm.ts
yeager-eren: We can also keep the original error when we are rethrowing using ProviderError, not sure sentry support or not, but we can test to see the result.
Error has a property called `cause`, documentation can be found here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/cause#rethrowing_an_error_with_a_cause

RanGojo: I've updated PR based on our offline discussion.

yeager-eren (approved):
lgtm

PR #767, opened 2024-06-11
feat: support tomo wallet

RanGojo:
# Summary

In this PR, I've added support for Tomo wallet extension for the EVM blockchains. This wallet doesn't have document for cosmos yet and it's interface is not fully compatible with Keplr, so we don't integrate its cosmos yet.

By installing Tomo extension and testing different scenarios including signing the transaction, switch network and so on.

# Checklist:

### DIFF
--- a/wallets/provider-tomo/package.json
+++ b/wallets/provider-tomo/package.json
@@ -1,0 +1,31 @@
+{
+  "name": "@rango-dev/provider-tomo",
+  "version": "0.1.0",
+  "license": "MIT",
+  "type": "module",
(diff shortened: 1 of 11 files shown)


wallets/provider-tomo/package.json
```diff
@@ -0,0 +1,31 @@
+{
+  "name": "@rango-dev/provider-tomo",
+  "version": "0.35.0",
```
RyukTheCoder, line 3: version should start from '0.1.0'

RyukTheCoder (approved):
LGTM

PR #769, opened 2024-06-13
fix: fix StepDetalis types import path

RaccoonskY:
Fixes # (issue)
Replaced SwapInputPropTypes source path to relative, because having 'src/..' causes problems when having the same name 'src' inside project.

# Checklist:


RanGojo:
@RaccoonskY Thanks for the PR. Seems good to me. Please update the commit message to a conventional commit format, so we could merge it.
something like this:
`fix: fix StepDetals.types import path`

RanGojo:
@RaccoonskY You've forget to squash and fix commits as well. Please squash your commits into a single conventional commit.

PR #774, opened 2024-06-24
fix: bug in updating input & output amount

mikasackermn:
# Summary
solved two issues:
1. Incorrect updating of the input amount occurs when the source token is eliminated.
2. The dollar value at the destination is not updated accurately when tokens are removed.

Fixes # (issue)

The issue was that the values weren't updated appropriately when the tokens were removed.
In order to improve display, I added a condition to the inputs component that states that values will be presented correctly in the event that the token is removed or the USD value is null.

It is preferable to take the test on the playground:

1. First, indicate the token and input's default values.
2. Navigate to the token selection page. Choose all options without choosing a token.
3. Return to the home screen.
In this instance, the USD values ought to be shown appropriately.

# Checklist:

### DIFF
--- a/widget/embedded/src/containers/Inputs/Inputs.tsx
+++ b/widget/embedded/src/containers/Inputs/Inputs.tsx
@@ -40,7 +40,6 @@ export function Inputs(props: PropTypes) {
   } = useQuoteStore();
   const { connectedWallets, getBalanceFor } = useWalletsStore();
   const fromTokenBalance = fromToken ? getBalanceFor(fromToken) : null;
-
   const fromTokenFormattedBalance =
     formatBalance(fromTokenBalance)?.amount ?? '0';
 
@@ -69,6 +68,7 @@ export function Inputs(props: PropTypes) {
     !inputUsdValue || !outputUsdValue || !outputUsdValue.gt(0)
       ? null
       : getPriceImpact(inputUsdValue.toString(), outputUsdValue.toString());
+
   return (
     <Container>
       <FromContainer>
--- a/widget/embedded/src/store/quote.ts
+++ b/widget/embedded/src/store/quote.ts
@@ -160,11 +160,12 @@ export const useQuoteStore = createSelectors(
 
           return {
             fromBlockchain: chain,
+            inputUsdValue: new BigNumber(0),
             ...(state.fromToken && {
               selectedQuote: null,
               fromToken: null,
               outputAmount: null,
-              outputUsdValue: null,
+              outputUsdValue: new BigNumber(0),
             }),
           };
         });
@@ -195,7 +196,7 @@ export const useQuoteStore = createSelectors(
               selectedQuote: null,
               toToken: null,
               outputAmount: null,
-              outputUsdValue: null,
+              outputUsdValue: new BigNumber(0),
             }),
           };
         });
@@ -260,8 +261,10 @@ export const useQuoteStore = createSelectors(
         set(() => ({
           fromToken: null,
           fromBlockchain: null,
+          outputUsdValue: new BigNumber(0),
+          inputUsdValue: new BigNumber(0),
+          inputAmount: '',
           outputAmount: null,
-          outputUsdValue: null,
           selectedQuote: null,
         })),
       resetToBlockchain: () =>
@@ -269,7 +272,7 @@ export const useQuoteStore = createSelectors(
           toToken: null,
           toBlockchain: null,
           outputAmount: null,
-          outputUsdValue: null,
+          outputUsdValue: new BigNumber(0),
           selectedQuote: null,
         })),
       setQuoteWalletConfirmed: (flag) =>


widget/embedded/src/containers/Inputs/Inputs.tsx
```diff
@@ -69,6 +69,11 @@ export function Inputs(props: PropTypes) {
     !inputUsdValue || !outputUsdValue || !outputUsdValue.gt(0)
       ? null
       : getPriceImpact(inputUsdValue.toString(), outputUsdValue.toString());
+
+  const inputUsd =
```
nikaaru, line 73:
```suggestion
  const inputUsdAmount =
```

widget/embedded/src/containers/Inputs/Inputs.tsx
```diff
@@ -69,6 +69,11 @@ export function Inputs(props: PropTypes) {
...
+    !inputUsdValue || !fromToken ? new BigNumber(0) : inputUsdValue;
+  const outputUsd =
```
nikaaru, line 75:
```suggestion
  const outputUsdAmount =
```

nikaaru:
I think the value of realUsdValue should also be updated
Because now when we show zero dollar amount, the tooltip is still showing the previous amount.

widget/embedded/src/containers/Inputs/Inputs.tsx
mikasackermn: fixed

widget/embedded/src/containers/Inputs/Inputs.tsx
mikasackermn: fixed

mikasackermn:
> I think the value of realUsdValue should also be updated Because now when we show zero dollar amount, the tooltip is still showing the previous amount.
> 
> Please check if it needs to be updated elsewhere

I fixed, now updated real Usd Value when delete selected token

widget/embedded/src/containers/Inputs/Inputs.tsx
```diff
@@ -140,11 +146,11 @@ export function Inputs(props: PropTypes) {
           usdValue: priceImpactOutputCanNotBeComputed
             ? undefined
             : numberToString(
-                outputUsdValue,
+                outputUsdAmount,
                 USD_VALUE_MIN_DECIMALS,
                 USD_VALUE_MAX_DECIMALS
               ),
-          realValue: formatTooltipNumbers(outputAmount),
+          realValue: formatTooltipNumbers(outputUsdAmount),
```
nikaaru, line 153: this is real output value, not real output USD value,
now, the tooltip for the real output value is incorrect,
Please revert this to its previous state.

Please make sure that all input and output items are correct for both the main value and its tooltip value and the dollar value and its tooltip value.

widget/embedded/src/containers/Inputs/Inputs.tsx
```diff
@@ -55,20 +60,21 @@ export function Inputs(props: PropTypes) {
 
   const priceImpactInputCanNotBeComputed = !canComputePriceImpact(
```
nikaaru, line 61: the second argument of the `canComputePriceImpact` function is the input amount, not the input USD amount.

widget/embedded/src/containers/Inputs/Inputs.tsx
```diff
@@ -55,20 +60,21 @@ export function Inputs(props: PropTypes) {
...
     selectedQuote,
-    inputAmount,
+    inputUsdAmount.toString(),
     inputUsdValue
   );
 
   const priceImpactOutputCanNotBeComputed = !canComputePriceImpact(
     selectedQuote,
-    inputAmount,
+    inputUsdAmount.toString(),
```
nikaaru, line 69: It should be like the input here:
the second argument of the `canComputePriceImpact` function is the input amount, not the input USD amount.

I think this bug already existed in production because we passed the input amount instead of the output amount.

widget/embedded/src/containers/Inputs/Inputs.tsx
mikasackermn: test

widget/embedded/src/containers/Inputs/Inputs.tsx
mikasackermn: fixed

widget/embedded/src/containers/Inputs/Inputs.tsx
mikasackermn: fixed

widget/embedded/src/containers/Inputs/Inputs.tsx
mikasackermn: fixed

mikasackermn:
> I think the value of realUsdValue should also be updated Because now when we show zero dollar amount, the tooltip is still showing the previous amount.
> 
> Please check if it needs to be updated elsewhere

fixed

nikaaru (approved):
thanks,
LGTM

widget/embedded/src/containers/Inputs/Inputs.tsx
```diff
@@ -40,10 +41,14 @@ export function Inputs(props: PropTypes) {
...
   const { connectedWallets, getBalanceFor } = useWalletsStore();
   const fromTokenBalance = fromToken ? getBalanceFor(fromToken) : null;
-
   const fromTokenFormattedBalance =
     formatBalance(fromTokenBalance)?.amount ?? '0';
 
+  const inputUsdAmount =
+    !inputUsdValue || !fromToken ? new BigNumber(0) : inputUsdValue;
+  const outputUsdAmount =
+    !outputUsdValue || !toToken ? new BigNumber(0) : outputUsdValue;
```
Ikari-Shinji-re, line 50: I believe it is better to fix this issue in the store, rather than using derived state. We have two functions for resetting the state of the quote store that will be invoked in this situation. You can set `outputUsdValue` to `null` in `resetFromBlockchain`.

mikasackermn: fixed

PR #779, opened 2024-06-30
feat: add a new event for the connect wallet button click

Ikari-Shinji-re:
# Summary

We need to detect when a user clicks on the "Connect Wallet" button at the bottom of the widget's homepage, prevent navigation to the wallet connection page, and allow for adding a custom callback from outside the widget.

Fixes # (issue)

```ts
  useEffect(() => {
    const handleEvent = (event: UiEventData) => {
      if (event.type === UiEventTypes.CLICK_CONNECT_WALLET) {
        console.log('ui events', event);
        event.payload.preventDefault();
      }
    };
    eventEmitter.on(WidgetEvents.UiEvent, handleEvent);

    return () => eventEmitter.off(MainEvents.UiEvent, handleEvent);
  }, [eventEmitter]);
```

# Checklist:

### DIFF
--- a/widget/embedded/src/utils/events.ts
+++ b/widget/embedded/src/utils/events.ts
@@ -1,0 +1,27 @@
+import { eventEmitter } from '../services/eventEmitter';
+import { type UiEventData, WidgetEvents } from '../types';
+
+type UiEvent = {
+  type: UiEventData['type'];
+  payload?: Omit<UiEventData['payload'], 'preventDefault'>;
+};
+
+export function emitPreventableEvent(event: UiEvent, action: () => void): void {
+  let defaultPrevented = false;
+
+  const extendedPayload = {
+    preventDefault() {
+      defaultPrevented = true;
+    },
+    ...(event.payload === undefined && { payload: event.payload }),
+  };
+
+  eventEmitter.emit(WidgetEvents.UiEvent, {
+    type: event.type,
+    payload: extendedPayload,
+  });
+
+  if (!defaultPrevented) {
+    action();
+  }
+}
--- a/widget/embedded/src/index.ts
+++ b/widget/embedded/src/index.ts
@@ -4,6 +4,7 @@ import type {
   BlockchainAndTokenConfig,
   QuoteEventData,
   Tokens,
+  UiEventData,
   WalletEventData,
   WalletInfoWithExtra,
   WidgetColors,
@@ -71,6 +72,7 @@ import { widgetEventEmitter } from './services/eventEmitter';
 import {
   WidgetEvents as MainEvents,
   QuoteEventTypes,
+  UiEventTypes,
   WalletEventTypes,
   WidgetEvents,
 } from './types';
@@ -108,6 +110,7 @@ export type {
   WidgetVariant,
   WalletEventData,
   QuoteEventData,
+  UiEventData,
   WalletInfoWithExtra,
 };
 export {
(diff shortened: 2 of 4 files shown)


widget/embedded/src/utils/common.ts
```diff
@@ -41,6 +41,22 @@ export function debounce(fn: Function, time: number) {
   }
 }
 
+export function createCustomEvent<T extends Record<string, unknown>>(
```
yeager-eren, line 44: I suggest this function to something like this:

```js
function emitPreventableEvent(type, payload, action){
  let defaultPrevented = false;
  const extendedPayload = {
    preventDefault(){ ... }
    ...payload, 
  };
  
  // `mitt` will run syncronously after emit an event. so we are sure an event can be prevented before running action.
  eventEmitter.emit(type, extendedPayload);
  
  if(!defaultPrevented){
    action()
  }
}
```

yeager-eren: In this way, you don't need to put a reference in useRef. You can use this util whenever you want to emit the event.

yeager-eren: Also i suggest to move this util to `/utils/event.ts`

widget/embedded/src/utils/common.ts
```diff
@@ -41,6 +41,22 @@ export function debounce(fn: Function, time: number) {
...
+
+  return {
+    preventDefault() {
+      defaultPrevented = true;
+    },
+    get defaultPrevented() {
+      return defaultPrevented;
+    },
+    ...eventProps,
+  } as CustomEventPayload<T>;
```
yeager-eren, line 57: Generally, assertion shouldn't be used and be solved in type itself (except exceptions). The problem is `eventProps` can be undefined. I think that could be solved in a way but if you do the previous change (`emitPreventableEvent`) I requested, I think this will go away from here.  So I just pointed at to make sure we are considering it for future PRs.

yeager-eren (changes requested):
Thanks. Like the solution, Added some comments to improve the code.

widget/embedded/src/utils/common.ts
Ikari-Shinji-re: I agree with these changes to avoid repetitive checks when using this function. However, to achieve this functionality, you need to store the result of this function somewhere for a specific event to preserve its closure.

widget/embedded/src/utils/events.ts
```diff
@@ -0,0 +1,34 @@
...
+      UiEventData['payload'],
+      'preventDefault' | 'defaultPrevented'
+    >;
+  },
+  action: () => void
+): void {
+  let defaultPrevented = false;
+
+  const extendedPayload = {
+    get defaultPrevented() {
```
yeager-eren, line 17: Not sure why we are exposing this to subscriber. Am I missing something here? If not needed please remove from here and type as well.

widget/embedded/src/utils/events.ts
```diff
@@ -0,0 +1,34 @@
+import { eventEmitter } from '../services/eventEmitter';
+import { type UiEventData, WidgetEvents } from '../types';
+
+export function emitPreventableEvent(
+  event: {
```
yeager-eren, line 5: Define a separate type instead of inline typing.

yeager-eren (changes requested):

yeager-eren (approved):
lgtm

PR #783, opened 2024-07-05
fix: fix flicker in playground

RyukTheCoder:
# Summary

Fixed a flicker that was happening in playground when navigating between pages in widget.

Fixes # rf-1611

Tested by navigating between widget pages in playground.

# Checklist:


mikasackermn (approved):

mikasackermn (commented):
Screenshot from 2024-07-07 13-00-31


The connect Metamask function is broken when External Wallets is true.

PR #796, opened 2024-07-10
fix: fix modal animation

RyukTheCoder:
# Summary

Fixed a lag in modal animation.

Fixes #1661

The problem was emerged in connect wallet modal in dapp so I linked "@rango-dev/ui" in app project and tested openning and closing that modal.

# Checklist:

### DIFF
--- a/widget/ui/src/components/Modal/Modal.styles.ts
+++ b/widget/ui/src/components/Modal/Modal.styles.ts
@@ -48,7 +48,7 @@ export const ModalContainer = styled('div', {
   display: 'flex',
   flexDirection: 'column',
   zIndex: 9999999,
-  transform: 'translateY(100%)',
+  transform: 'translateY(100vh)',
   transition: 'transform .45s ease-in-out',
 
   variants: {
@@ -61,7 +61,7 @@ export const ModalContainer = styled('div', {
       right: {
         borderTopRightRadius: '0',
         borderBottomRightRadius: '0',
-        transform: 'translateX(100%)',
+        transform: 'translateX(100vw)',
       },
     },
     active: {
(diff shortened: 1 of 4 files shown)


widget/ui/src/components/Modal/Modal.styles.ts
```diff
@@ -48,7 +48,7 @@ export const ModalContainer = styled('div', {
   display: 'flex',
   flexDirection: 'column',
   zIndex: 9999999,
-  transform: 'translateY(100%)',
+  transform: 'translateY(100vh)',
```
Ikari-Shinji-re, line 51: It partially solves the problem. I believe the other issue with the modal component is its implementation. Using timeouts can make it buggy. The modal animation duration is 450 ms, and there's a variable called CLOSED_DELAY that refers to the time the modal will be removed from the DOM after the user action, which has a different value (600 ms). I think we can remove these timeouts and use the native `onAnimationEnd` event on the modal container. Something like this:

```ts
const Modal = ({ open } : {open: boolean}) => {
  const [isMounted, setIsMounted] = useState(open);
  const [animationClass, setAnimationClass] = useState('');

  useEffect(() => {
    if (open) {
      setIsMounted(true);
      setAnimationClass('mount');
    } else {
      setAnimationClass('unmount');
    }
  }, [open]);

  const handleAnimationEnd = () => {
    if (!open) {
      setIsMounted(false);
    }
  };

  return (
    isMounted && (
      <ModalContainer
       //Modify this to be more suitable for Stitches.js
        className={animationClass}
        onAnimationEnd={handleAnimationEnd}
      >
     .............................
      </ModalContainer>
    )
  );
};
```

Ikari-Shinji-re (approved):

Ikari-Shinji-re (approved):

PR #798, opened 2024-07-13
fix: fix sort logic of notifications

RyukTheCoder:
# Summary

Fixed sort logic of notifications so they get displayed sorted based on transactionCreationTime.

Fixes # 1662

Tested this by creating a transaction. So when that tx is in the check status state, if I get a new notification, it will appear on top of the old check status notification in notifications list.

# Checklist:

### DIFF
--- a/widget/embedded/src/store/notification.ts
+++ b/widget/embedded/src/store/notification.ts
@@ -36,6 +36,7 @@ export const useNotificationStore = createSelectors(
             creationTime: Date.now(),
             requestId: route.requestId,
             route: {
+              creationTime: parseInt(route.creationTime),
               from: {
                 blockchain: fromStep.fromBlockchain,
                 address: fromStep.fromSymbolAddress,
@@ -67,7 +68,18 @@ export const useNotificationStore = createSelectors(
         getNotifications: () => {
           const { isSynced, notifications } = get();
           return isSynced
-            ? notifications.sort((a, b) => b.creationTime - a.creationTime)
+            ? notifications.sort((a, b) => {
+                if (a.route.creationTime && !b.route.creationTime) {
+                  return -1;
+                }
+                if (!a.route.creationTime && b.route.creationTime) {
+                  return 1;
+                }
+                if (!a.route.creationTime && !b.route.creationTime) {
+                  return b.creationTime - a.creationTime;
+                }
+                return b.route.creationTime - a.route.creationTime;
+              })
             : [];
         },
         syncNotifications: (swaps) => {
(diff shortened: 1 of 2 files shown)


widget/embedded/src/types/notification.ts
```diff
@@ -21,5 +21,6 @@ type NotificationRoute = {
 export type Notification = Pick<Route, 'requestId'> & {
   event: RouteEvent | StepEvent;
   creationTime: number;
+  transactionCreationTime: number;
```
Ikari-Shinji-re, line 24: I believe it is more related to the route information, so it would be better to move this property to the route object:
```ts
type NotificationRoute = {
  transactionCreationTime: number;
  from: {
    blockchain: Step['fromBlockchain'];
    symbol: Step['fromSymbol'];
    address: Step['fromSymbolAddress'];
  };
  to: {
    blockchain: Step['toBlockchain'];
    symbol: Step['toSymbol'];
    address: Step['toSymbolAddress'];
  };
};
```

widget/embedded/src/store/notification.ts
```diff
@@ -67,7 +68,13 @@ export const useNotificationStore = createSelectors(
...
           const { isSynced, notifications } = get();
           return isSynced
-            ? notifications.sort((a, b) => b.creationTime - a.creationTime)
+            ? notifications.sort((a, b) =>
+                !a.transactionCreationTime ||
+                !b.transactionCreationTime ||
+                a.transactionCreationTime === b.transactionCreationTime
+                  ? b.creationTime - a.creationTime
+                  : b.transactionCreationTime - a.transactionCreationTime
+              )
```
Ikari-Shinji-re, line 77: We know that if the `transactionCreationTime` property exists, the transaction is newer than an item that does not have this property. Therefore, we can change this to:

```ts
notifications.sort((a, b) => {
      if (a.transactionCreationTime && !b.transactionCreationTime) {
        return -1;
      }
      if (!a.transactionCreationTime && b.transactionCreationTime) {
        return 1;
      }
      if (!a.transactionCreationTime && !b.transactionCreationTime) {
        return b.creationTime - a.creationTime;
      }
      return b.transactionCreationTime - a.transactionCreationTime;
    })
```

widget/embedded/src/types/notification.ts
RyukTheCoder: Tnx Fixed

widget/embedded/src/store/notification.ts
RyukTheCoder: Tnx Fixed

Ikari-Shinji-re (approved):

PR #805, opened 2024-07-20
fix: address the issue where the token list does not update when the balance changes

Ikari-Shinji-re:
# Summary

If we’re on the Token Selector page and the balance of the selected blockchain changes, the token list sort order fails to update.

- [x] Connect a wallet that supports the auto-connect feature. Then, navigate to the Token Selector page, select a blockchain with a balance in your wallet, and refresh the page. After the wallet auto-connects, the token list sort order should be correct. 
- [x]  Use the Widget Playground and an external wallet for this test. Go to the Token Selector, choose a blockchain with a known balance in your `MetaMask` wallet, and use the external wallet feature to connect and disconnect `MetaMask`. The token list sort order should be correct after both connecting (and fetching the balance) and disconnecting the wallet.

# Checklist:

### DIFF
--- a/widget/embedded/src/components/TokenList/TokenList.tsx
+++ b/widget/embedded/src/components/TokenList/TokenList.tsx
@@ -17,6 +17,7 @@ import {
 } from '@rango-dev/ui';
 import React, { useEffect, useState } from 'react';
 
+import { useObserveBalanceChanges } from '../../hooks/useObserveBalanceChanges';
 import { useAppStore } from '../../store/AppStore';
 import { useWalletsStore } from '../../store/wallets';
 import { createTintsAndShades } from '../../utils/colors';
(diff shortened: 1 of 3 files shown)


nikaaru (approved):
thanks,
LGTM

widget/embedded/src/components/TokenList/TokenList.tsx
```diff
@@ -110,7 +112,7 @@ export function TokenList(props: PropTypes) {
 
   useEffect(() => {
     setTokens(list.slice(0, PAGE_SIZE));
-  }, [list.length, selectedBlockchain]);
+  }, [list.length, selectedBlockchain, balanceKey]);
```
RyukTheCoder, line 115:
```suggestion
  }, [JSON.stringify(list), selectedBlockchain]);
```

Tnx, I appreciate your contribution, but I think you just wrote a complex hook with some unnecessary dependencies to EventEmitter and other things to just prepare a value for a useEffect dependency. Can you please explain why we can not solve the problem by changing this single line of code?

PR #813, opened 2024-07-23
chore: migrate core to nodenext module system

yeager-eren:
# Summary

Add a `tsconfig.libnext.json` which is [recommended by TS](https://www.typescriptlang.org/docs/handbook/modules/guides/choosing-compiler-options.html#im-writing-a-library) and make `wallets/core` and `wallets/shared` compatible with those configs.

To make these packages I had to fix hoisting problem of `trezor-provider`'s dependency called `@types/web` which has conflicts with `@types/dom` and also moved `core/src` to `core/src/legacy` to make it ready for upcoming changes by Hub.

Please run build and dev commands and make sure nothing has broken. Also check usage of these packages to ensure their regarding types is working as expected as well. 

# Checklist:


RyukTheCoder:
Some related info can be found at below address:

[https://stackoverflow.com/a/71473145](https://stackoverflow.com/a/71473145)

PR #821, opened 2024-07-31
chore: refactor connect modals which includes namespace and derivation

yeager-eren:
# Summary

Refactor connect wallets modal to avoid duplication and be reusable.

Sometimes we can not directly connect to a wallet and we need more information like what namespace user is trying to connect (e.g. Ledger) or for hardware wallets we need to get a derivation path then we can connect to wallet. We have multiple places that let user connect their wallets (e.g. Wallets page), this logic had been duplicated in these places over time. This PR makes them into one component called `StatefulConnect` and all logics or sub-components can be exported and reused in other places (like dApp).

Fixes RF-1684

**Note:**  The behavior should be matched with our `next` version. you compare my PR and the last `next` deployment to review and test. 

All places a user can connect wallets are affected by these changes.

These are 3 main places for user to connect wallet:
1. Wallet page: you can click on the wallet icon in header.
2. Confirm wallets page: Try to do a swap, on confirm page, user is confirming what wallet is going to be used for swapping. There is a `Show more wallet` button, by clicking on the button you can test that page as well.
3. Swap page: Try to do a swap and when you've been navigated to `Swap` page try to reload page and then disconnect your wallet then you will see a "connect wallet" in your swap page.

Our widget integrators can config wallets to use a `single wallet` only. This may be affected.

When user is trying to connect a wallet, if it's connecting quickly, we shouldn't show the modal that shows `Connecting...`.  Usually, auto connect and already connected wallets (switching between connected and disconnected) are the use case. I changed some code regarding this, we should ensure it works as before.

Experimental chain modals also has been moved. I didn't touched the code, only restructure and moved them but please consider to check and test them as well. You can use your Keplr wallet and BitSong to see the expermental chain. 

One last thing I want to mention, previously all modals (namespace, derivation path, ...) were showing separately. In this PR, I changed the structure and we open a modal, and only changing the contents of modal if needed(not closing a modal and opening another modal).

Notes for reviewer:

- I moved some lingui text from components to helpers, we should ensure they will not be affected.
- I removed the need of having delay for ModalStatefulConnect. I made some changes on  `ui/modal`. 

# Checklist:


RyukTheCoder (changes requested):

widget/embedded/src/hooks/useStatefulConnect/useStatefulConnect.ts
```diff
@@ -0,0 +1,222 @@
...
+    wallet: WalletInfoWithExtra,
+    selectedNamespaces: Namespace[]
+  ): Promise<Result> => {
+    const isSingleNamespace = wallet.singleNamespace;
+    const needsDerivationPath = wallet.needsDerivationPath;
+    const firstSelectedNamespace = selectedNamespaces[0];
+
+    if (!firstSelectedNamespace) {
+      throw new Error(
+        'To confirm a namespace, you should select at least one namespace.'
```
RyukTheCoder, line 137: When the wallet is singleNamespace, the `at least one namespace` part of the message is not correct, because the user should select exactly one namespace.

widget/embedded/src/hooks/useStatefulConnect/useStatefulConnect.ts
```diff
@@ -0,0 +1,222 @@
...
+      });
+
+      return { status: ResultStatus.DerivationPath };
+    }
+
+    /**
+     * If execution reaches here, means we don't have any more optional step (like namespace and derivation)
+     * So we should start connecting wallet.
+     */
+    const type = connectState.namespace!.providerType;
```
RyukTheCoder, line 163: Forbidden non-null assertion.

widget/embedded/src/hooks/useStatefulConnect/useStatefulConnect.ts
```diff
@@ -0,0 +1,222 @@
...
+
+    return await runConnect(type, namespaces);
+  };
+
+  const handleDerivationPath = async (
+    derivationPath: string
+  ): Promise<Result> => {
+    if (!derivationPath) {
+      throw new Error(
+        "Derivation path is empty. Please make sure you've filled the field correctly."
```
RyukTheCoder, line 176:
```suggestion
        "Derivation Path can not be empty."
```

widget/embedded/src/hooks/useStatefulConnect/useStatefulConnect.ts
```diff
@@ -0,0 +1,222 @@
...
+    }
+
+    const isStateOnDerivationPathStep = !!connectState.derivationPath;
+
+    if (!isStateOnDerivationPathStep) {
+      throw new Error(
+        'It seems you are filling derivation path without setting namespace before doing that. Please retry to connect.'
+      );
+    }
+    const type = connectState.derivationPath!.providerType;
```
RyukTheCoder, line 187: Forbidden non-null assertion.

widget/embedded/src/hooks/useWalletList.ts
```diff
@@ -175,9 +145,6 @@ export function useWalletList(params: Params) {
     list: sortedWallets.filter(
       (wallet) => !shouldExcludeWallet(wallet.type, chain ?? '', blockchains)
     ),
-    error,
-    handleClick,
-    disconnectConnectingWallets,
-    disconnectWallet,
+    terminateConnectingWallets,
```
RyukTheCoder, line 148: Considering that you moved all the logic related to handling click, connect and disconnect to `useStatefulConnect` hook, why did you keep `terminateConnectingWallets` here?

widget/embedded/src/components/WalletStatefulConnect/ExperimentalChain.tsx
```diff
@@ -0,0 +1,36 @@
+import { i18n } from '@lingui/core';
+import { Button, Divider, MessageBox } from '@rango-dev/ui';
+import React from 'react';
+
+interface PropTypes {
+  displayName?: string;
+  onConfirm: () => void;
+}
+
+export function ExperimentalChain(props: PropTypes) {
```
RyukTheCoder, line 10: I think name of this component does not specify what it is. Obviously this component is not an experimental chain.

widget/embedded/src/components/ModalStatefulConnect/ModalStatefulConnect.tsx
```diff
@@ -0,0 +1,142 @@
...
+      onExit={() => {
+        resetModalState();
+        terminateConnectingWallets();
+
+        if (props.onClose) {
+          props.onClose();
+        }
+      }}
+      container={
+        document.getElementById(WIDGET_UI_ID.SWAP_BOX_ID) || document.body
```
RyukTheCoder, line 120: Please use `getContainer` from utils instead

widget/embedded/src/components/ModalStatefulConnect/ModalStatefulConnect.tsx
```diff
@@ -0,0 +1,142 @@
...
+      }>
+      {isOnStatus ? (
```
RyukTheCoder, line 122: I think using a function to render the content of modal is better than this complex conditional rendering.

widget/embedded/src/components/ModalStatefulConnect/ModalStatefulConnect.tsx
```diff
@@ -0,0 +1,142 @@
...
+        <>
+          <ConnectStatus wallet={props.wallet!} error={error} />
+          <Divider direction="vertical" size={32} />
+        </>
+      ) : null}
+      {isOnNamespace && (
+        <Namespaces
+          onConfirm={handleConfirmNamespaces}
+          value={getState().namespace!}
```
RyukTheCoder, line 131: Forbidden non-null assertion.

RyukTheCoder (changes requested):

widget/ui/src/components/Modal/Modal.tsx
```diff
@@ -27,6 +27,8 @@ export function Modal(props: PropsWithChildren<ModalPropTypes>) {
     title,
     open,
     onClose,
+    onEnter,
```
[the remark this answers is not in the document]
yeager-eren: These are general APIs that can be used. onExit and onEnter has a meaning in context of animations that you can see in other libraries as well.

widget/ui/src/components/Modal/Modal.tsx
```diff
@@ -44,6 +46,17 @@ export function Modal(props: PropsWithChildren<ModalPropTypes>) {
...
+  const modalContainerRef = useRef<HTMLElement | null>(null);
+  const exitCallbackRef = useRef<ModalPropTypes['onExit']>();
+  const enterCallbackRef = useRef<ModalPropTypes['onEnter']>();
+
+  useEffect(() => {
+    exitCallbackRef.current = onExit;
+  }, [onExit]);
+
+  useEffect(() => {
+    exitCallbackRef.current = onEnter;
```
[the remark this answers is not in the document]
yeager-eren: Good catch! thanks. fixed.

widget/embedded/src/components/ModalStatefulConnect/index.ts
```diff
@@ -0,0 +1 @@
+export { ModalStatefulConnect } from './ModalStatefulConnect';
```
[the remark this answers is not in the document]
yeager-eren: We are prefixing our modals using `Modal....`, to be consistent with our convention that would be better to keep it same. 
Though, I can change it to `ModalStatefulConnectWallet` if you prefer. I'm ok with both but I think the only connect functionality we have is connecting wallet, So for keeping the name shorter, I removed it. but we can explicitly using it name as you suggested.

widget/embedded/src/pages/WalletsPage.tsx
```diff
@@ -228,28 +95,11 @@ export function WalletsPage() {
...
-            open={!!derivationPathModalState?.open}
-            selectedNamespace={derivationPathModalState?.namespace}
-            type={derivationPathModalState?.providerType}
-            image={derivationPathModalState?.providerImage}
-            onClose={handleCloseDerivationPathModal}
-            onConfirm={handleDerivationPathConfirm}
+          <ModalStatefulConnect
+            wallet={selectedWalletToConnect}
+            onClose={() => {
+              setSelectedWalletToConnect(undefined);
```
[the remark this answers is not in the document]
yeager-eren: It has different meaning in different context. What I consider is `undefined` when you are initializing value and `null` when you're data is empty.  
For example consider a fetch request, it has 3 states,:`undefined` means request hasn't been sent at all, `null` means request has sent and we didn't get anything from response and also the third one is the actual data.

In this case I only has two state, has data or it's undefined which means modal shouldn't be shown in page. there is no third state to set an empty value, it doesn't make any sense here.

So I prefer to keep it as `undefined` here.

widget/embedded/src/hooks/useWalletList.ts
```diff
@@ -22,26 +21,31 @@ import {
   sortWalletsBasedOnConnectionState,
 } from '../utils/wallets';
 
+import { useStatefulConnect } from './useStatefulConnect/useStatefulConnect';
+
 const ALL_SUPPORTED_WALLETS = Object.values(WalletTypes);
 
 interface Params {
   chain?: string;
-  onBeforeConnect?: (walletType: string) => void;
```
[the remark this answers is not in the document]
yeager-eren: Good point! As you can see in `widget/embedded/src/index.ts`, we marked it as deprecated so it shouldn't be used generally. But I see your point on keeping the interface to don't have breaking change. But I believe these changes will simplify our code and the added value here with consideration I mentioned earlier can make an exception here.

widget/embedded/src/hooks/useWalletList.ts
```diff
@@ -22,26 +21,31 @@ import {
...
-  onError?: (error?: string) => void;
+}
+
+interface API {
+  list: WalletInfoWithExtra[];
+  terminateConnectingWallets: () => void;
 }
 
 /**
  * gets list of wallets with their information and an action for handling click callback fo UI
```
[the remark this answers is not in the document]
yeager-eren: Good catch! thanks. I wrote an updated description.

widget/embedded/src/hooks/useStatefulConnect/useStatefulConnect.ts
```diff
@@ -0,0 +1,222 @@
...
+  ) => Promise<Result>;
+  handleDerivationPath: (path: string) => Promise<Result>;
+  handleDisconnect: (type: WalletType) => Promise<Result>;
+  getState(): State;
+  resetState(section?: 'derivation'): void;
+}
+
+/**
+ *
+ * This hook maintain an state and detects if the wallet needs what information
```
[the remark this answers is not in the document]
yeager-eren: thanks. updated.

widget/embedded/src/hooks/useStatefulConnect/useStatefulConnect.ts
yeager-eren: Original text has a suggested action next to the error. I prefer to keep the original text.

widget/embedded/src/hooks/useWalletList.ts
yeager-eren: the logic has broken into two separated part. one is list, and one is for handling to connect to any of them. the list (`useWalletsList`) and connect logic (`useStatefulConnect`). `terminateConnectingWallets` is actually an action the whole list, not related to a single wallet.

widget/embedded/src/components/WalletStatefulConnect/ExperimentalChain.tsx
yeager-eren: It was `ModalExpermintalChain`, I only removed the `Modal`. That would be great to give me your suggestion as well.

widget/embedded/src/components/ModalStatefulConnect/ModalStatefulConnect.tsx
yeager-eren: Thanks. I used the util.

widget/embedded/src/hooks/useStatefulConnect/useStatefulConnect.ts
yeager-eren: It should be an unreachable sate for the case you are mentioning. I just used a general message. it will be used for phantom soon as well (which is multi namespaces).
If we want to change it we should do something like this:

```js
if(! firstSelectedNamespace){
  let message = "To confirm a namespace, you should select at least one namespace."
  if(singleNamespace){
    message = "To confirm a namespace, you should select a namespace."
  }
  throw new Error(message)
}
```

The trade off here is:
1. Showing an exact error message for each cases. One of them should be unreachable. add more logic..
2. Showing a general message, and keep the code simple.

What do you think?

widget/embedded/src/components/ModalStatefulConnect/ModalStatefulConnect.tsx
yeager-eren: I used a typeguard here. should be fixed.

widget/embedded/src/pages/WalletsPage.tsx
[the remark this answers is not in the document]
RyukTheCoder: Tnx I respect your opinion its not that important at the final result. But just for the record, null is a primitive value that represents the intentional absence of any object value and undefined is a primitive value used when a variable has not been assigned a value. So we do not assign a variable with `undefined`.

widget/embedded/src/hooks/useStatefulConnect/useStatefulConnect.ts
RyukTheCoder: It was just a reminder, I'm ok with the way it is.

widget/embedded/src/components/ModalStatefulConnect/ModalStatefulConnect.tsx
RyukTheCoder: I think of something like this:

```

  const renderModalContent = () => {
    if (isOnStatus(getState, props.wallet)) {
      return (
        <>
          <ConnectStatus wallet={props.wallet} error={error} />
          <Divider direction="vertical" size={32} />
        </>
      );
    } else if (isOnNamespace(getState)) {
      return (
        <Namespaces
          onConfirm={handleConfirmNamespaces}
          value={getState().namespace}
        />
      );
    }
    return (
      <DerivationPath
        onConfirm={handleDerivationPathConfirm}
        value={getState().derivationPath}
      />
    );
  };

  return (
    <WatermarkedModal
      open={isConnecting}
      onClose={handleClosingModal}
      onExit={() => {
        resetModalState();
        terminateConnectingWallets();

        if (props.onClose) {
          props.onClose();
        }
      }}
      container={getContainer()}>
      {renderModalContent()}
    </WatermarkedModal>
  );
  ```

File: widget/embedded/src/pages/WalletsPage.tsx (hunk 13 above)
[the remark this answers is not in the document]
Comment (yeager-eren) [reply widget/embedded/src/pages/WalletsPage.tsx:101]: That is exactly my point, `undefined` here is using for checking whether modal exists or not. which matches is what you've said. `null` usually will be used for objects where they are expected but hasn't any value, in that context undefined means a property doesn't exist in schema but null means it exists but don't have any value. Here the value is string.

File: widget/embedded/src/hooks/useStatefulConnect/useStatefulConnect.ts (hunk 2 above)
Comment (yeager-eren) [reply widget/embedded/src/hooks/useStatefulConnect/useStatefulConnect.ts]: Fixed them by adding a typegurad.

File: widget/embedded/src/hooks/useStatefulConnect/useStatefulConnect.ts (hunk 4 above)
Comment (yeager-eren) [reply widget/embedded/src/hooks/useStatefulConnect/useStatefulConnect.ts]: Fixed them by adding a typegurad.

File: widget/embedded/src/components/ModalStatefulConnect/ModalStatefulConnect.tsx (hunk 8 above)
Comment (yeager-eren) [reply widget/embedded/src/components/ModalStatefulConnect/ModalStatefulConnect.tsx]: It doesn't add any value (like readability or improvement) and seems same as the original code to me. So I prefer to not change.

**Comment by yeager-eren** [comment]
@RyukTheCoder I also did some refactoring on Modal. Please, make sure you will not miss the change.

File: widget/ui/src/components/Modal/Modal.tsx [hunk 17]
```diff
@@ -81,7 +94,35 @@ export function Modal(props: PropsWithChildren<ModalPropTypes>) {
         );
       }
     }
+
```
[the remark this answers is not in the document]
yeager-eren: I rebased my branch and did some changes on Modal. We shouldn't have any problems after this.

widget/embedded/src/components/ModalStatefulConnect/index.ts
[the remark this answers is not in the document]
yeager-eren: Missed this one. I updated the PR again to rename ModalStatefulConnect to StatefulConnectModal

RyukTheCoder (approved):
LGTM

PR #825, opened 2024-08-07
chore: move wallets/react logics into 2 separate hooks

yeager-eren:
# Summary

Separate legacy logics into two hooks to make it ready for adding Hub adapter.

For making `wallets/react` compatible we should keep the interface same and also our providers will be migrated to Hub gradually. So we need to keep the logic as well. In this PR I mostly move codes and my changes was minor.

Fixes RF-1732

I mentioned my changes by putting a comment. It shouldn't affect any other places since I just moved some codes.

# Checklist:


wallets/react/src/legacy/useLegacyProviders.ts
```diff
@@ -0,0 +1,218 @@
...
+  // Get (or add) wallet instance (`provider`s will be wrapped in a `Wallet`)
+  const getWalletInstance = useInitializers(
+    makeEventHandler(dispatch, props.onUpdateState)
+  );
+
+  // Getting providers from props and put all of them in a `Map` with an appropriate interface.
+  const listOfProviders = props.providers;
+  const wallets = checkWalletProviders(listOfProviders);
+
+  useAutoConnect({
```
yeager-eren, line 32: In original code, it was an useEffect (the last one in `provider.tsx`). I moved it to `useAutoConnect` and use it here.

wallets/react/src/provider.tsx
```diff
@@ -1,228 +1,15 @@
...
-      !autoConnectInitiated.current;
-
-    if (shouldTryAutoConnect) {
-      autoConnectInitiated.current = true;
-      void (async () => {
-        await autoConnect(wallets, getWalletInstance);
-      })();
-    }
-  }, [props.autoConnect, props.allBlockChains]);
+  const legacyApi = useLegacyProviders(props);
```
yeager-eren, line 9: I moved all the logics into `useLegacyProviders`. In a separate PR, I will add hub providers as a hook as well and there will be a adapter to make both of them work without breaking anything.

RyukTheCoder (approved):
LGTM

PR #838, opened 2024-08-15
chore: restructure shared and react to use /core package

yeager-eren:
# Summary

Update all imports from `core` to import from `core/legacy` and move some functionality from `/shared` to `/core`.

In long term we will remove `wallets-shared`, a part of this PR is moving `core`-related functionality to `wallets-ocre`. And also this PR prefixed all the legacy core with `legacyThenSomething` to make sure we are flagging them and give a notice they are deprecated and will be removed in next major version.

### Aliasing 

In those packages who imported types or functions from legacy code, you will see I'm aliasing the `legacySomething` to `legacySomething as something`, this is because I wanted to update only imports and not the usages. I believe this is more appropriate way to reduce the changes.

### `core/legacy/package.json`

module system other than `node16` and `bundler` doesn't have support for `package.json`'s `exports` field. For backward-compatibility reasons we needed a workaround. I checked some repositories and all of them used the way thar I used in this PR. [here is an example](https://github.com/apollographql/apollo-server/blob/main/packages/server/errors/package.json).

It shouldn't have any side effect on user side. it can affect our building process. so make sure the project will be build correctly. 

# Related PRs

- https://github.com/rango-exchange/rango-client/pull/623

# Checklist:

### DIFF
--- a/examples/queue-manager-demo/src/flows/rango/helpers.ts
+++ b/examples/queue-manager-demo/src/flows/rango/helpers.ts
@@ -32,7 +32,7 @@ import type {
 } from 'rango-types';
 
 import { SUPPORTED_ETH_CHAINS as XDEFI_WALLET_SUPPORTED_EVM_CHAINS } from '@rango-dev/provider-xdefi';
-import { readAccountAddress } from '@rango-dev/wallets-react';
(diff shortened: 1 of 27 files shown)


RyukTheCoder (approved):
LGTM

PR #842, opened 2024-08-18
chore: upgrade ethers to v6

RyukTheCoder:
# Summary

Upgraded ethers dependency package from @5.7.2 to @6.13.2.
Here is a link to a migration guide: https://docs.ethers.org/v6/migrating/

Migration was straight forward in most places and is done by just upgrading it or a related package except in `signer-evm`. Here is the list of changes in this package according to the migration guide:

- `providers.Web3Provider` replaced by `BrowserProvider`
- `provider.getSigner` now returns a promise so it can not be used in constructor
- nonce should be parsed to integer
- `getChainId` is not available in AbstractSigner so getting and checking the validity of it has changed

Notes: 

- `@ledgerhq/evm-tools` which is a dependency of `@ledgerhq/hw-app-eth` package which is used in `provider-ledger` still uses `ethers@5.7.2` and could not be upgraded. So, it configured to be bundled using `external` option.
- `ethereum-waffle` package which is used in providing a mockProvider to be used in tests was using `ethers@5.7.2` and could not be upgraded. So, I removed it and used `Wallet` of `ethers` instead.

Fixes # 1466

Tested by signing an evm transaction (on Polygon) using MetaMask. EVM transactions on providers which use defaultEvmSigner (like Metmask, Rabby, Tomo, Frontier, etc) should be tested to sign and broadcast transactions properly. Also Safe wallet should be tested because of upgrading a dependency. 

safe wallet tested and the transaction with below id executed successfully:
d0adf2e7-e6ef-40bf-9af3-71a5ded0609b

# Checklist:


examples/queue-manager-demo/src/wallet.js
```diff
@@ -1,44 +1,58 @@
...
-// Make a transaction and run
-// check status from server
-// notif the resul
\ No newline at end of file
+/*
+ * Switch network
+ * Make a transaction and run
+ * check status from server
+ * notif the resul
+ */
```
RanGojo, line 58: You could remove this comment.

signers/signer-evm/src/signer.ts
```diff
@@ -52,12 +49,15 @@ const waitWithMempoolCheck = async (
...
+  private provider: BrowserProvider;
 
   constructor(provider: ProviderType) {
-    this.provider = new providers.Web3Provider(provider);
-    this.signer = this.provider.getSigner();
+    /*
+     * Connect to the MetaMask EIP-1193 object. This is a standard
+     * protocol that allows Ethers access to make all read-only
+     * requests through MetaMask.
+     */
```
RanGojo, line 59: I recommend either removing this comment or shortening it.

```suggestion
    // get read-only access to the provider
```

signers/signer-evm/src/signer.ts
```diff
@@ -102,7 +102,13 @@ export class DefaultEvmSigner implements GenericSigner<EvmTransaction> {
 
   async signMessage(msg: string): Promise<string> {
     try {
-      return await this.signer.signMessage(msg);
+      /*
+       * It provides an opportunity to request access to write
+       * operations, which will be performed by the private key
+       * that MetaMask manages for the user.
+       */
```
RanGojo, line 109: same here.


```suggestion
      // request access to write operations
```

RanGojo:
Thanks for your well-documented PR. I didn't test your PR yet but I've put some comments about the code changes.

RanGojo (changes requested):

signers/signer-evm/src/signer.ts
RyukTheCoder: Improved comment

signers/signer-evm/src/signer.ts
RyukTheCoder: Improved comment

examples/queue-manager-demo/src/wallet.js
RyukTheCoder: Removed

signers/signer-evm/src/signer.ts
```diff
@@ -15,10 +12,10 @@ import {
 import { cleanEvmError, getTenderlyError, waitMs } from './helper.js';
 import { RPCErrorCode } from './types.js';
 
-type ProviderType = ConstructorParameters<typeof providers.Web3Provider>[0];
+type ProviderType = ConstructorParameters<typeof BrowserProvider>[0];
 
 const waitWithMempoolCheck = async (
-  provider: providers.Web3Provider,
+  provider: BrowserProvider,
```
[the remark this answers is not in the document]
RyukTheCoder: this function is not exported and it is only used in this file in `DefaultEvmSigner` so I think it is better to be typed as the `provider` in `DefaultEvmSigner`.

signers/signer-evm/src/signer.ts
```diff
@@ -52,12 +49,15 @@ const waitWithMempoolCheck = async (
 };
 
 export class DefaultEvmSigner implements GenericSigner<EvmTransaction> {
-  private signer: providers.JsonRpcSigner;
-  private provider: providers.Web3Provider;
+  private provider: BrowserProvider;
 
   constructor(provider: ProviderType) {
```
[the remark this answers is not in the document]
RyukTheCoder: Changed to `Eip1193Provider`.

Considering that our tests are running successfully, it means this works in node.js environment.

RanGojo:
@RyukTheCoder Please fix the conflicts, I would like to test the latest changes with vercel when deployed.

signers/signer-evm/src/signer.ts
```diff
@@ -15,10 +17,8 @@ import {
 import { cleanEvmError, getTenderlyError, waitMs } from './helper.js';
```
RanGojo, line 17: Please update cleanEvmError code based on this document:
https://docs.ethers.org/v6/api/utils/errors/

RyukTheCoder: I checked it, shortMessage contains a suitable error messages in most cases except "UNKNOWN_ERROR" which is handled by checking the message of the inner error object.

RanGojo: You could also update catch code in wait in evm signer like this:

```
    } catch (error) {
      if (isError(error, "TRANSACTION_REPLACED")) {
        const reason = error.reason
        if (reason === "cancelled") {
          throw new SignerError(
            SignerErrorCode.SEND_TX_ERROR,
            undefined,
            'Transaction replaced and canceled by user',
            undefined,
            error
          );
        }
        return { hash: error.replacement.hash, response: error.replacement };
      } else if (isError(error, "CALL_EXCEPTION")) {
        const tError = await getTenderlyError(chainId, txHash);
```

widget/app/public/index.html
```diff
@@ -7,6 +7,7 @@
       content="width=device-width, initial-scale=1, maximum-scale=1"
     />
     <title>Rango Widget</title>
+    <link rel="manifest" href="./manifest.json" />
```
RanGojo, line 10: This manifest file is fine. The reason that still doesn't work with safe site is that parcel is transforming `manifest.json`to `manifest.webmanifest` during the build phase and safe web app doesn't accept it.

You should check parcel document to avoid it or you could update build script to re-transform it. (for test purpose)

something like this:
`    "build": "parcel build --cache-dir=.parcel-cache && sed -i 's/manifest.webmanifest/manifest.json/g' ./dist/index.html && cp ./dist/manifest.webmanifest ./dist/manifest.json",
`

signers/signer-evm/src/signer.ts
RanGojo: Can you give me an example of reproducing provider error?

RyukTheCoder: I couldn't reproduce such error, I said that based on the existing code. So I removed the existing code and used ethers for error handling. I think `rango-types` package should be updated as well. Also updated `wait` function as you said.
[fix: improve errors](https://github.com/rango-exchange/rango-client/pull/842/commits/dccb0673a12cc9a64787230e5bdbb42ec2c68a6f)

widget/app/public/index.html
RyukTheCoder: I tried that with placing manifest.json in dist directory and changing index.html file but it didn't work.

signers/signer-evm/src/signer.ts
```diff
@@ -210,12 +209,9 @@ export class DefaultEvmSigner implements GenericSigner<EvmTransaction> {
...
       const error = err as any; // TODO find a proper type
-      if (
-        error?.code === RPCErrorCode.TRANSACTION_REPLACED &&
-        error?.replacement
-      ) {
-        return { hash: error?.replacement?.hash, response: error?.replacement };
-      } else if (error?.code === RPCErrorCode.CALL_EXCEPTION) {
+      if (isError(error, 'TRANSACTION_REPLACED')) {
+        return { hash: error.replacement.hash, response: error.replacement };
+      } else if (isError(error, 'CALL_EXCEPTION')) {
```
RanGojo, line 214: Please also handle canceled scenario like this.
```ts
      if (isError(error, "TRANSACTION_REPLACED")) {
        const reason = error.reason
        if (reason === "cancelled") {
          throw new SignerError(
            SignerErrorCode.SEND_TX_ERROR,
            undefined,
            'Transaction replaced and canceled by user',
            undefined,
            error
          );
        }
        return { hash: error.replacement.hash, response: error.replacement };
      }
```

signers/signer-evm/src/signer.ts
```diff
@@ -210,12 +209,9 @@ export class DefaultEvmSigner implements GenericSigner<EvmTransaction> {
       return { hash: txHash };
     } catch (err) {
       const error = err as any; // TODO find a proper type
```
RanGojo, line 211: We could remove this line.

signers/signer-evm/src/signer.ts
RyukTheCoder: Tnx fixed.

signers/signer-evm/src/signer.ts
RyukTheCoder: Removed

RanGojo (approved):
LGTM

wallets/provider-ledger/package.json
```diff
@@ -14,28 +14,30 @@
...
+    "ethers": "^6.13.2",
     "rango-types": "^0.1.69"
   },
+  "devDependencies": {
+    "@ledgerhq/errors": "^6.16.4",
+    "@ledgerhq/hw-app-eth": "^6.36.0",
+    "@ledgerhq/hw-app-solana": "^7.1.6",
+    "@ledgerhq/hw-transport-webhid": "^6.28.6",
+    "@ledgerhq/types-cryptoassets": "^7.11.0",
+    "@ledgerhq/types-devices": "^6.24.0"
```
RanGojo, line 38: @RyukTheCoder I'm not sure about this change. Please double check with @yeager-eren if you didn't already.

yeager-eren: moving to `devDependencies` will break our `tsc` users. please revert them back.

RyukTheCoder: reverted moving ledgerhq to devDependencies.
[fix: revert moving ledgerhq to dev dependencies](https://github.com/rango-exchange/rango-client/pull/842/commits/788ae2356cdf505c17bd6a469054c7863054fd7f)

PR #845, opened 2024-08-19
fix: update `signMessage` in the default Solana signer to return a base58 string instead of base64

Ikari-Shinji-re:
# Summary

update `signMessage` in the default Solana signer to return a base58 string instead of base64 .

You can test this change using the `wallets-demo` package.

# Checklist:

### DIFF
--- a/signers/signer-solana/src/signer.ts
+++ b/signers/signer-solana/src/signer.ts
@@ -1,6 +1,7 @@
 import type { SolanaExternalProvider } from './utils/types.js';
 import type { GenericSigner, SolanaTransaction } from 'rango-types';
 
+import base58 from 'bs58';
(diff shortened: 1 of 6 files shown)


signers/signer-solana/src/signer.ts
```diff
@@ -21,7 +22,7 @@ export class DefaultSolanaSigner implements GenericSigner<SolanaTransaction> {
           message: encodedMessage,
         },
       });
-      return Buffer.from(signature).toString('base64');
+      return base58.encode(signature);
```
RanGojo, line 25: I think we have similar codes in solflare signer.

RanGojo (approved):
I didn't test it, but code looks good to me. I will test it after fixing of solflare problem you've mentioned. (it should be fix in next branch)

PR #847, opened 2024-08-19
Chore/rf 1782 avoid running patch package on host machines

yeager-eren:
# Summary

`patch-packge` shouldn't be needed to run on host machines.

We should make sure the final build will work without needing to patching any node_modules packages. There are 3 packages to fix building issues we did use the `patch-package` for them. In this PR, I'm trying to avoid this.

Fixes # (issue)

- `solflare` doesn't need to be patched since we are using `bundler`;
- wallet connect has fixed in a tricky way, make sure you will read `notes.md`.
- I also added an option to our build script. tried to make it documented on the file.
- I remove `postinstall` from `signer-cosmos`, hopefully host will fix the long package itself.

# Checklist:


RyukTheCoder (approved):
LGTM

PR #824, opened 2024-08-07
test: add tests for wallets/react

yeager-eren:
# Summary

Add integration tests for `walelt-react`.

This PR includes some integration tests for legacy provider and also some todo tests for doing for future.

Fixes RF-1731

# How did you test this change?

Run `yarn test`

# Related PRs

- https://github.com/rango-exchange/rango-client/pull/852
- https://github.com/rango-exchange/rango-client/pull/623

# Checklist:

- [ ] I have performed a self-review of my code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] Implemented a user interface (UI) change, referencing our Figma design to ensure pixel-perfect precision.

# Notes
First I need to change `connect` interface for legacy.

PR #852, opened 2024-08-21
feat: hub adapter and refactor wallets store

yeager-eren:
# Summary

Add adapter to use hub providers alongside legacy providers.

This PR includes adding hub adapter in `/react/` and refactor `wallets` store.  From now on, we should be able to develop with our new wallet management system, Hub. 

You can test the changes by running `yarn dev:widget`.

Details on how to review and what to test has been written in next section.

# Notes for reviewers

## Store
__Assignee: @Ikari-Shinji-re__

We had a separate store for `wallets`, in order to continue our work to migrate our stores to a single store and make current stores a slice of that, I completely removed `wallets` store and made it an slice.

The structure is mostly same, but there are one major difference. Previously, we had a `connectedWallets` which each of them had a `balance` itself. To simplify the process, I added a separate `_balances` and keep them separate from `connectedWallet`.

This change helps us do calculation much simpler (like calculating total balances) and treat this part differently, for example on connecting wallet it has two phases, one for adding the wallet to `connectedWallets` and another action for getting balances.

#### Other notable changes

- We were slicing tokens list to simulate pagination (to not showing a long scrollbar), I removed it since it was buggy and also make it more complicated without bringing much value.
-  I'm keeping the raw balance data in store, which means whenever we need to show balance to user, we need to make sure we are using `shiftedBy` in order to show a correct number to user. There were places like `Max` button needs to be updated which I did. I just mentioned this, maybe there are more places like this I'm unaware.

#### Area of changes

Mostly `embedded/src/store`, you can take a look at `embedded` as well and they will be updating usage of `wallets` store.

Also removing `balance` from `rango-preset` is related to wallets store.

## AutoConnect
__Assignee: @Ikari-Shinji-re__

For having auto connect in `Hub`, in addition to wallet type/name, we need `namespace` as well. So I added a separate key in storage.

I tried to reuse the legacy codes as much as I can, `eagerConnectHandler`, `LastConnectedWalletsFromStorage`, same filenames are some effort to achieve this.

`LastConnectedWalletsFromStorage` is actually a shared interface to interact with both legacy and hub storage.

#### Area of changes

You can start by reviewing `react/src/legacy/`, I tried to make the legacy code reusable by separating the logic into `autoConnect`, `useAutoConnect`. Auto connect is calling in `useLegacyProvider`.

And then you can start taking a look at `react/src/hub/`, I tried to use same filename and structure in legacy to make it more consistent and understandable.

When you are reviewing this, you will notice there are some new functions like `LastConnectedWallets` which are actually a general solutions on legacy codes that needs to be reused in adapter.

These files should be interested to you:
- `autoConnect,ts`
- `constancts.ts`
- `lastConnectedWallets.ts`
- `useHubAdapter.ts`

#### In Future

- Having a migration for migrating legacy auto connect storage to hub storage

## Adapter
__Assignee: @RyukTheCoder __

In previous PRs, we've prepared the filesystem structure for adding an adapter. In this PR, We've added a new `/react/src/hub` directory  alongside `/react/src/legacy`.

The idea here is keeping the legacy interface to not needing to update clients (like embedded), to achieve this I created a new hook called `useProviders` which provides the legacy (`ProviderContext`) at the end. It simply will load both hub and legacy providers and it's switch between them. For example you are going to use `connect` in `embedded`, when it's going to run, it will check the request wallet is a legacy provider or hub provider, then it will call the method on that one.

Each legacy and hub has a hook to implement the legacy interface (`ProviderContext`), they are `useLegacyProviders` and `useHubProviders`.

For backward compatibility, there is an special namespace called `DISCOVER_MODE`. Alongside `DISCOVER_MODE`, `network` will be set as well. here we are manually matching networks to namespaces. This will help us keep the legacy interface and have what hub needs as well. Explained some more details on next section.

#### Other notable changes

- To call a hub provider, having `namespace` is required. Since the legacy hasn't this concept, we only have `network` there. Adapter has a mapping between `Networks` and Hub's `Namespace`. Take a look at `discoverNamespace` to see the mapping. That would be great if @RanGojo check this mapping as well.
- `Network` can not be passed to wallet's instances (e.g. `window.ethereum`) directly, for example for switching network, we need to convert `Network` enum to `chain id`. take a look at `tryConvertNamespaceNetworkToChainInfo` in this regard.

#### Area of changes

Hub enabled in `widget/app/src/App.tsx`, and there are some update regarding to `connect`, it needs to have `DISCOVER_MODE` ( e.g `widget/embedded/src/QueueManager.tsx`).

Adapter has been written in `/react/src/hub`. Some parts of the changes is for `auto connect`. It separately has been assigned to @Ikari-Shinji-re .

Changes to `/react/src/legacy` are mostly for auto connect feature.

#### Breaking Changes

- Second param for `connect` in `/react` has changed. I updated the usage in `embedded`.

## Phantom
__Assignee: @RyukTheCoder __

Phantom has been migrated to hub and also in addition to Solana namespace, EVM namespace has been added too.

#### Area of changes

You can take a look at `/provider-phantom`.

## Other changes

- Hub can be enabled and disabled by setting `configs.isExperimentalEnabled`. I made it `enable` for now.

# TODO

- [ ]  add package.json for subpath and remove `utils` from` /core/mod.ts`
- [ ]  Write some guide on hub

# In Future

- Check hot reload for multiple entrypoins, seems Parcel doesn't support ESM exports for hot reload properly.
- `suggestAndConnect` hasn't implemented in Hub adapter. 
- hub adapter is using`canSwitchNetworkTo`, `canEagerConnect`, `getSigners`, `getWalletInfo().supportedChains`, `getInstance` from `legacy` provider.

# Related PRs

- https://github.com/rango-exchange/rango-client/pull/826
- https://github.com/rango-exchange/rango-client/pull/824
- https://github.com/rango-exchange/rango-client/pull/623


RyukTheCoder (changes requested):

RyukTheCoder:
I get this error when I try to connect to Solana on an account which does not have Solana address. Error message should be improved.

image

RyukTheCoder:
I can not do anything when I connect to Ethereum on Phantom and try to swap on Solana with Phantom. I should be able to disconnect Phantom and connect it with a different namespace.

[screen-capture (8).webm](https://github.com/user-attachments/assets/4b9f6acf-2706-43af-8839-b8897089faae)

RyukTheCoder:
I can connect Ethereum and Polygon when connecting to EVM namespace but in this modal only Ethereum is shown.

image

wallets/provider-phantom/src/mod.ts
```diff
@@ -0,0 +1,11 @@
+import { defineVersions } from '@rango-dev/wallets-core/utils';
+
+import { legacyProvider } from './legacy/index.js';
+import { provider } from './provider.js';
+
+const versions = defineVersions()
```
RyukTheCoder, line 6: As a suggestion, can't we handle this by exporting `legacy` and `hub` separately instead of this `versions` approach? Maybe that way we can prevent including `legacy` or `hub` code based on the value of `experimentalWallet` in `config`.

wallets/provider-phantom/src/legacy/index.ts
```diff
@@ -24,7 +28,15 @@ export const config = {
 };
 
 export const getInstance = phantom_instance;
-export const connect: Connect = getSolanaAccounts;
+const connect: Connect = async ({ instance, meta }) => {
+  const solanaInstance = instance.get(Networks.SOLANA);
+  const result = await getSolanaAccounts({
```
RyukTheCoder, line 33: It seems that you are trying to keep support of legacy providers but you didn't add `connect` and `sign` for `evm` namespace. If it is not possible, at least it should be documented.

wallets/react/src/hub/utils.ts
```diff
@@ -0,0 +1,388 @@
...
+  type: string
+): LegacyProviderInterface {
+  const [legacy] = separateLegacyAndHubProviders(allProviders);
+  const provider = legacy.find((legacyProvider) => {
+    return legacyProvider.config.type === type;
+  });
+
+  if (!provider) {
+    console.warn(
+      `You have a provider that hasn't legacy provider. it causes some problems since we need some legacy functionality. Provider Id: ${type}`
```
RyukTheCoder, line 321:
```suggestion
      `You have a provider that doesn't have legacy provider. It causes some problems since we need some legacy functionality. Provider Id: ${type}`
```

wallets/react/src/useProviders.ts
```diff
@@ -0,0 +1,120 @@
...
+    providers(): Providers {
+      let output: Providers = {};
+      if (hubProviders.length > 0) {
+        output = { ...output, ...hubApi.providers() };
+      }
+      if (legacyProviders.length > 0) {
+        output = { ...output, ...legacyApi.providers() };
+      }
+
+      return output;
```
RyukTheCoder, line 95: What is the problem with this?
```suggestion
      return { ...hubApi.providers(), ...legacyApi.providers() };
```

RyukTheCoder (changes requested):

widget/embedded/src/containers/Wallets/Wallets.tsx
```diff
@@ -139,14 +138,19 @@ function Main(props: PropsWithChildren<PropTypes>) {
     }),
     []
   );
+  const isExperimentalEnabled =
+    props.config.features?.experimentalWallet === 'enabled';
```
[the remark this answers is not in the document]
yeager-eren: Not much useful util but I changed for consistency.
I can not remember the exact reason, but I think we are passing `config` object as string in iframe. so "false" would be simply true on `!!"false"` conditions.   not sure though, I just followed `experimentalRoute` convention here as well.

wallets/provider-phantom/src/legacy/index.ts
```diff
@@ -70,6 +86,25 @@ export const getWalletInfo: (allBlockChains: BlockchainMeta[]) => WalletInfo = (
       DEFAULT: 'https://phantom.app/',
     },
     color: '#4d40c6',
-    supportedChains: solana,
+    // if you are adding a new namespace, don't forget to also update `properties`
+    supportedChains: [
+      ...solana,
+      ...evms.filter((chain) =>
+        EVM_SUPPORTED_CHAINS.includes(chain.name as Networks)
+      ),
```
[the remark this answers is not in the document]
yeager-eren: `allBlockChains ` includes 60 items and EVM ones < 20. the code had about 20 more iterations which shouldn't be expensive.
To reduce iterations, we could also filtering both solana and evm in same loop, but for sure code readability would affected to save some nanoseconds.

Anyway I changed the code to reduce the iterations as you suggested.

wallets/provider-phantom/src/legacy/signer.ts
```diff
@@ -1,14 +1,18 @@
...
 
 export default async function getSigners(
   provider: any
 ): Promise<SignerFactory> {
   const solProvider = getNetworkInstance(provider, Networks.SOLANA);
+  const evmProvider = getNetworkInstance(provider, Networks.ETHEREUM);
   const signers = new DefaultSignerFactory();
-  const { DefaultSolanaSigner } = await import('@rango-dev/signer-solana');
   signers.registerSigner(TxType.SOLANA, new DefaultSolanaSigner(solProvider));
+  signers.registerSigner(TxType.EVM, new DefaultEvmSigner(evmProvider));
```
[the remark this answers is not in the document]
yeager-eren: Good catch! they are lazy now.

wallets/provider-phantom/src/utils.ts
```diff
@@ -0,0 +1,55 @@
...
+
+  return instances;
+}
+
+export function evmPhantom(): EvmProviderApi {
+  const instances = phantom();
+
+  const evmInstance = instances?.get(LegacyNetworks.ETHEREUM);
+
+  if (!instances || !evmInstance) {
```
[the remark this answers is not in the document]
yeager-eren: good point. remvoed.

wallets/provider-phantom/src/mod.ts
yeager-eren: We've talked about this offline, I mentioned about parcel has struggling with `exports` filed and also versioning maybe removed in future versions since it's kinda experimental.

So I prefer to keep it this way yet.

wallets/provider-phantom/src/legacy/index.ts
yeager-eren: Good point. I added a comment on code to explain we will not add new functionality to legacy anymore.

wallets/react/src/hub/utils.ts
yeager-eren: updated.

wallets/react/src/useProviders.ts
yeager-eren: Not sure. Only god knows.
updated.

yeager-eren: OK. I can officially say god is not alone anymore.
After checking more, when `isExperimentalEnabled` set to false, `separateLegacyAndHubProviders` will return only legacy providers.
Your code causes hub providers to be returned always, but when it's disabled it we shouldn't run `...hubApi.providers()`.

So I revert your suggestion and kept the original code.

wallets/provider-phantom/src/mod.ts
```diff
@@ -0,0 +1,11 @@
...
+  .version('0.0.0', legacyProvider)
```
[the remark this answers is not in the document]
yeager-eren: `.version` first parameter is union, so making the value a `const` in code to make sure we are reusing same string across the usage, doesn't add any value here.

yeager-eren:
> I get this error when I try to connect to Solana on an account which does not have Solana address. Error message should be improved.
> 
> image

Good catch! Unfortunately we can not do much in this case since Phantom is throwing `-32603` error code which is [internal error](https://docs.phantom.app/solana/errors) so we can not detect what's really going on. 

We should leave it as it is for now, I will open an issue on Phantom repo. I also added a todo in code to make we don't miss this one.

yeager-eren:
> I can not do anything when I connect to Ethereum on Phantom and try to swap on Solana with Phantom. I should be able to disconnect Phantom and connect it with a different namespace.
> 
>  screen-capture.8.webm

Thanks for the report. I fixed it in my last commit.

RyukTheCoder (approved):
I reveiwed Adapter and Phantom. LGTM

Ikari-Shinji-re:
It seems the retry function for fetching balances hasn't been implemented. After a wallet connects, we send a request to the server to retrieve the balance of the connected wallet. We use a batch endpoint to get the balances across multiple blockchains in a single request and response. If the server is unable to fetch the balance for any blockchain's address within the expected time, it returns an error for that specific blockchain in the response. We should handle this by checking for errors and retrying the balance fetch for the failed blockchains.

Ikari-Shinji-re:
When a wallet is connected but no blockchain is selected, and we navigate from the home page to the token and blockchain selection page, there's a noticeable lag during the transition and when rendering the token list. To mitigate this, we leveraged the balances stored in the wallet store, which helps avoid this issue. The store values are updated when wallets are connected or disconnected, and accessing an asset has a time complexity of O(1).

https://github.com/user-attachments/assets/22781a79-532d-435a-9172-ae364e6fa974

Ikari-Shinji-re:
The automatic selection of connected wallets is no longer functioning as it used to, and now users have to manually select their wallets. When a wallet connects, we should check all blockchain accounts associated with that wallet, and if no other wallet is selected for a specific blockchain, we automatically select the connected wallet for that blockchain. Similarly, if the wallet is disconnected, we should review the list of connected wallets for that blockchain and mark the first one in the list as selected. This approach reduces the number of clicks required from the user, improving the overall UX.
The user can change the selected wallet for a specific blockchain in the "Confirm Wallets" modal by clicking on the desired wallet and confirming the selection in the modal.

https://github.com/user-attachments/assets/a2be988e-18c7-41dd-a46d-da91878df1db

Ikari-Shinji-re:
When a wallet is connected with multiple namespaces, there may be cases where a wallet is incorrectly displayed in the list of supported wallets for a blockchain in the "Confirm Wallets" modal. In the attached video, I connect the Solana namespace for the Phantom wallet, then navigate to a route that requires a Polygon wallet to be connected. In the supported wallets section, the Phantom wallet is shown as connected, even though it has no account for Polygon. While Phantom supports the Polygon blockchain, its EVM namespace is not connected. In this situation, we should either exclude the wallet or display it in a disconnected state, allowing the user to connect the relevant namespace for the wallet and proceed with the transaction.
We should discuss this issue with the team members.

https://github.com/user-attachments/assets/2af0cdd6-357f-4b2a-b593-14691a93ec89

Ikari-Shinji-re:
When multiple wallets are connected with the same account, disconnecting one of them results in the balance for that address being completely removed, even though other wallets with the same account are still connected.

https://github.com/user-attachments/assets/1ef6176e-1721-41cd-a98f-38fafd932bb8

wallets/react/src/legacy/useLegacyProviders.ts
```diff
@@ -30,22 +42,48 @@ export function useLegacyProviders(props: ProviderProps): ProviderContext {
...
+        (
+          ns
+        ): ns is Exclude<
+          LegacyNamespaceInputForConnect,
+          LegacyNamespaceInputWithDiscoverMode
+        > => {
+          return ns.namespace !== 'DISCOVER_MODE';
+        }
+      );
+      // Legacy providers doesn't implemented multiple namespaces, so it will always be one value.
```
RyukTheCoder, line 72: I don't think that this a valid assumption, why should there be such limitation for legacy providers to only implement one single namespace?

wallets/provider-phantom/src/legacy/index.ts
RyukTheCoder: But you have added `ETHEREUM` and `POLYGON` to supported chains.

RyukTheCoder (changes requested):

wallets/provider-phantom/src/legacy/index.ts
yeager-eren: It will not be needed eventually when adapter removed its dependency to `wallentInfo.supportedChains`. I mentioned `In Future` section.

widget/embedded/src/containers/WidgetInfo/WidgetInfo.types.ts
```diff
@@ -13,7 +13,7 @@ export interface WidgetInfoContextInterface {
     details: ConnectedWallet[];
     totalBalance: string;
     isLoading: boolean;
-    refetch: (accounts: Wallet[], tokens: Token[]) => void;
```
mojtabast, line 16: this can be a breaking change. i don't know why we'd got a second param and didn't use in implementation.

yeager-eren, line 16: this can be a breaking change since the hook is public. i don't know why we'd got a second param and didn't use in implementation.

yeager-eren:
> 0879cd0

You're right, I'd removed the retry logic by mistake. I've added a retry in my last commit  `0879cd000eb4a36a44a26775873fd6158d4cc307`.

yeager-eren:
> It seems the retry function for fetching balances hasn't been implemented. After a wallet connects, we send a request to the server to retrieve the balance of the connected wallet. We use a batch endpoint to get the balances across multiple blockchains in a single request and response. If the server is unable to fetch the balance for any blockchain's address within the expected time, it returns an error for that specific blockchain in the response. We should handle this by checking for errors and retrying the balance fetch for the failed blockchains.

You're right, I'd removed the retry logic by mistake. I've added a retry in my last commit  `0879cd000eb4a36a44a26775873fd6158d4cc307`.

yeager-eren:
> When a wallet is connected but no blockchain is selected, and we navigate from the home page to the token and blockchain selection page, there's a noticeable lag during the transition and when rendering the token list. To mitigate this, we leveraged the balances stored in the wallet store, which helps avoid this issue. The store values are updated when wallets are connected or disconnected, and accessing an asset has a time complexity of O(1).
> 
>  1.mp4

Good catch. I fixed it in e99356d88a9ef64ebaf715912c547a247c7833c4

yeager-eren:
> When multiple wallets are connected with the same account, disconnecting one of them results in the balance for that address being completely removed, even though other wallets with the same account are still connected.
> 
>  2.mp4

This should've been fixed in my latest commit `e99356d88a9ef64ebaf715912c547a247c7833c4 `

widget/embedded/src/containers/WidgetInfo/WidgetInfo.types.ts
RanGojo: It's needed in dApp. We shouldn't break it.

yeager-eren:
> 8aae1ea

Fixed in 8aae1eae39466a256bc39ef37ec86904c668280d

yeager-eren:
@RanGojo Thanks. I fixed the NaN bug in my last commit ab16aea43d0259b66e6f4bdbb9b337dc976374a9.
The rest of items have been discussed offline.

widget/embedded/src/containers/WidgetInfo/WidgetInfo.types.ts
yeager-eren: Thanks for the heads up. I opened a PR there.

yeager-eren:
going to merge this. we will fix some reported bugs in separate PRs.

PR #854, opened 2024-08-23
chore: use radix to improve modal component

RyukTheCoder:
# Summary

Use Radix UI to improve modal component.
In result of this change:
- Modal will close by pressing escape button
- Focus will be trapped in modal

Fixes # 1777

Notes:
Currently, we are handling the logic related to showing animation on opening and closing modal and positioning the content of modal based on `anchor` prop. Before exploring the Radix UI document, we were optimistic about that it can handle these animation logics and positioning the content of modal, but after checking the documentation I didn't find such solutions so I kept it as it is.

Tested this change by checking most of modals.

# Checklist:

### DIFF
--- a/widget/ui/src/components/Modal/Modal.styles.ts
+++ b/widget/ui/src/components/Modal/Modal.styles.ts
@@ -8,13 +8,13 @@ const DialogOverlayAnimateIn = keyframes({
     backgroundColor: 'transparent',
   },
   '100%': {
-    backgroundColor: 'color-mix(in srgb, $neutral500 70%, transparent)',
+    backgroundColor: '$overlay',
   },
 });
 
 const DialogOverlayAnimateOut = keyframes({
   '0%': {
-    backgroundColor: 'color-mix(in srgb, $neutral500 70%, transparent)',
+    backgroundColor: '$overlay',
   },
   '100%': {
     backgroundColor: 'transparent',
@@ -35,7 +35,7 @@ export const DialogOverlay = styled(Dialog.Overlay, {
   backgroundColor: 'transparent',
 
   "&[data-state='open']": {
-    backgroundColor: 'color-mix(in srgb, $neutral500 70%, transparent)',
+    backgroundColor: '$overlay',
     animation: `${DialogOverlayAnimateIn} .45s ease-in-out`,
   },
   "&[data-state='closed']": {
--- a/widget/ui/src/components/Modal/Modal.styles.ts
+++ b/widget/ui/src/components/Modal/Modal.styles.ts
@@ -179,6 +179,10 @@ export const Content = styled('div', {
   overflowX: 'hidden',
 });
 
+export const DialogTitle = styled(Dialog.DialogTitle, {
+  margin: 0,
+});
+
 export const Footer = styled('div', {
   padding: '0 $20 $10',
   '& .footer__logo': {
--- a/widget/ui/src/components/Modal/Modal.styles.ts
+++ b/widget/ui/src/components/Modal/Modal.styles.ts
@@ -1,7 +1,25 @@
 import * as Dialog from '@radix-ui/react-dialog';
 
-import { styled } from '../../theme.js';
+import { keyframes, styled } from '../../theme.js';
 import { IconButton } from '../IconButton/index.js';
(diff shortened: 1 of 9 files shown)


yeager-eren:
Thanks. Your changes looks good to me. For handling animation Radix seems using [data attributes](https://www.radix-ui.com/primitives/docs/components/dialog#content). [Here is an example](https://github.com/shadcn-ui/ui/blob/9ef7967b0d39e7084f9e319aec0e97ca4a31a739/apps/www/registry/default/ui/dialog.tsx) from Shadcn doing samething on top of Radix. Please take a look at the approach and see is there anyway to remove our js approach to handle animations.

RyukTheCoder:
> Thanks. Your changes looks good to me. For handling animation Radix seems using [data attributes](https://www.radix-ui.com/primitives/docs/components/dialog#content). [Here is an example](https://github.com/shadcn-ui/ui/blob/9ef7967b0d39e7084f9e319aec0e97ca4a31a739/apps/www/registry/default/ui/dialog.tsx) from [Shadcn](https://ui.shadcn.com/docs/components/dialog) doing samething on top of Radix. Please take a look at the approach and see is there anyway to remove our js approach to handle animations.

Please check my latest changes. I used data attributes to handle animations.

RyukTheCoder:
In [this](https://github.com/rango-exchange/rango-client/pull/854/commits/1fa28fb5540c311c3eb5e2c833efbf1315eb67df) commit I tried to prevent an error related to `radix` which indicates that "`DialogContent` requires a `DialogTitle` for the component to be accessible for screen reader users.". It still exits for modals with a custom header. This error will completely be removed by removing custom headers and using `ModalHeader` component.

widget/ui/src/components/Modal/Modal.styles.ts
```diff
@@ -179,6 +179,10 @@ export const Content = styled('div', {
   overflowX: 'hidden',
 });
 
+export const DialogTitle = styled(Dialog.DialogTitle, {
+  margin: 0,
```
yeager-eren, line 183: I think this will affect a11y and they did the trick for that. We can ignore 1px difference.

yeager-eren (approved):
LGTM.

widget/ui/src/components/Modal/Modal.styles.ts
RyukTheCoder: This is not to handle 1px difference. Actually with position set to `absolute`, there is no 1px difference. When we wrap `title` with `Dialog.DialogTitle` it will be wrapped in a `h2` element which have a default margin so I added this to remove the unwanted margin.

Ikari-Shinji-re:
Currently, we have a Modal component in our design system package. However, our implementation works differently depending on its container.

Based on our requirements for our widget and other applications, we have identified two types of modals:

A variant that occupies the entire viewport, blocking user interaction outside its content.
A variant that behaves like a modal but can receive a specific container, blocking interaction only within that container when open.
In this pull request, we aimed to integrate @radix-ui/react-dialog into our modal component to better handle animation, state management, and accessibility. However, due to limitations in their API and implementation, some issues couldn't be resolved cleanly. I also evaluated other libraries like MUI and Headless UI, but they align with Radix’s modal implementation, adhering to the W3C dialog modal pattern:
https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/

These libraries position the modal as fixed, blocking the entire viewport, which does not meet our requirement for container-based positioning (e.g., bottom sheets in our widget).

To address this, I propose splitting the Modal component into two separate components:

Modal: A standard modal built with external packages like Radix UI, incorporating most of the changes made in this pull request.
BottomSheet: A new component based on our current implementation. This would allow absolute positioning within a container while maintaining the behavior required for bottom sheets. We can enhance its accessibility by managing focus and adding keyboard shortcuts.
While reviewing other UI libraries, I noticed their drawer components function similarly to modals, blocking the entire viewport. Therefore, keeping our existing implementation for BottomSheet seems to be the most practical approach to meet our specific needs.

yeager-eren:
@Ikari-Shinji-re Thanks. I will close this issue and we will address this in our upcoming design system.

PR #856, opened 2024-08-26
fix: incorrect error message on Trezor wallet transaction rejection

mikasackermn:
# Summary

The error message "cancelled" appears on the Trezor wallet when a transaction is declined. What should appear instead is "Reason: User rejected the transaction." 

Fixes # (issue)

 Transact with Trezer, but don't verify it in the wallet.

# Checklist:


wallets/provider-trezor/src/helpers.ts
```diff
@@ -20,6 +20,9 @@ export async function getTrezorModule() {
 }
 
 export function getTrezorErrorMessage(error: any) {
+  if (error?.message === 'Cancelled') {
```
RyukTheCoder, line 23: Please check if you can use `cleanEVMError` function.

RyukTheCoder (changes requested):

wallets/provider-trezor/src/helpers.ts
mikasackermn: I had used `cleanEVMError` but it didn't support this error

mikasackermn:
```
export function getTrezorErrorMessage(error: any) {
  if (error?.message === 'Cancelled') {
    return new Error('User rejected the transaction.', { cause: error });
  }
  if (error?.shortMessage) {
    /*
     * Some error signs have lengthy, challenging-to-read messages.
     * shortMessage is used because it is shorter and easier to understand.
     */
    return new Error(error.shortMessage, { cause: error });
  }
  return cleanEvmError(error);
}
```

RyukTheCoder: `cleanEvmError` should support errors thrown by ethers and provider, we should not handle a single error using an `if` statement. Please check the type of error implement and see if there are other errors which are not supported properly and implement a general solution.

mikasackermn: Based on the error we received from the trezor side, I modified this section.

 Trezor error type:

```
export interface Unsuccessful {
    success: false;
    payload: {
        error: string;
        code?: string;
    };
}

```

RyukTheCoder (approved):
LGTM

PR #855, opened 2024-08-24
feat: implement lazy loading for some rango-client dependencies

mikasackermn:
# Summary

This PR implements lazy loading for certain dependencies within the rango-client to optimize the bundle size and improve application performance. By dynamically importing modules only when needed, we can reduce the initial load time and enhance the overall user experience, especially for users who may not require all features immediately.

Fixes # (issue)
The difference in bundle size before and after:

before

Screenshot from 2024-08-27 14-34-28

after

Screenshot from 2024-08-27 14-58-26

# How did you test this change?
 
 Because most of the changes have happened in Cosmos and Solana and Terra wallets and Signers, please test the wallets and Signers.

Packages that are lazily loaded: 
- @terra-money/terra.js
- @cosmjs/launchpad
- @cosmjs/stargate
- @keplr-wallet/cosmos
- @solana/web3.js
- cosmos-wallet
- @keplr-wallet/simple-fetch

# Checklist:

- [x] I have performed a self-review of my code

PR #861, opened 2024-09-07
perf: lazy load signers

RyukTheCoder:
# Summary

Lazy load all signer packages. Bundle analyze report compared to result of [related PR](https://github.com/rango-exchange/rango-client/pull/855) is provided in the following zip file:

[result.zip](https://github.com/user-attachments/files/16963859/result.zip)

## before:
image
## related PR:
image
## after:
image
## after resolve DefaultEvmSigner problem:
image

Fixes # 1796

Tested by signing a transaction using Metamask. Everything should be the same as production and the related bundle files should be downloaded on signing transactions.

# Notes:
`signer-solana` could not be lazy loaded because we call `setSolanaSignerConfig` of it in the first and also its dependency `@solana/web3.js` could not be lazy loaded because it is used in `@solflare-wallet/sdk` which is a dependency of `provider-solflare`.

# Lighthouse Insights:

## Before:
image

## After:
image

# Checklist:

### DIFF
--- a/queue-manager/rango-preset/src/actions/checkStatus.ts
+++ b/queue-manager/rango-preset/src/actions/checkStatus.ts
@@ -66,7 +66,9 @@ async function checkTransactionStatus({
     const txType = getCurrentStepTxType(currentStep);
     const sourceWallet = getRelatedWallet(swap, currentStep);
     if (txType && sourceWallet) {
-      signer = context.getSigners(sourceWallet.walletType).getSigner(txType);
+      signer = (await context.getSigners(sourceWallet.walletType)).getSigner(
+        txType
+      );
     }
   } catch (error) {
     /*
@@ -289,7 +291,8 @@ async function checkApprovalStatus({
     const txType = getCurrentStepTxType(currentStep);
     const sourceWallet = getRelatedWallet(swap, currentStep);
     if (txType && sourceWallet) {
-      signer = context.getSigners(sourceWallet.walletType).getSigner(txType);
+      const walletSigners = await context.getSigners(sourceWallet.walletType);
+      signer = walletSigners.getSigner(txType);
     }
   } catch (error) {
     /*
--- a/queue-manager/rango-preset/src/helpers.ts
+++ b/queue-manager/rango-preset/src/helpers.ts
@@ -991,9 +991,9 @@ export function isRequiredWalletConnected(
   return { ok: matched, reason: 'account_miss_match' };
 }
 
-export function signTransaction(
+export async function signTransaction(
   actions: ExecuterActions<SwapStorage, SwapActionTypes, SwapQueueContext>
-): void {
+): Promise<void> {
   const { setTransactionDataByHash } = inMemoryTransactionsData();
   const { getStorage, setStorage, failed, next, schedule, context } = actions;
   const { meta, getSigners, isMobileWallet } = context;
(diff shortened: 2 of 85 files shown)


RanGojo:
Could you also provide a diagram for the remaining parts of the bundle? @RyukTheCoder I'd like to get a clearer understanding of the key improvements for the future.

nikaaru:
I tested with metamask,
When I want to sign the transaction, EVM signers do not load in lazy loading mode.
It seems that it is in the initial bundle and is not properly lazy-loaded.

RanGojo:
Since we are going to close the previous PR, please also check related usage of `solana/web3.js` out of the signers. I think they were handled in previous PR. Also about the shared report, I'm not sure why some chnuks are duplicated in html report.

queue-manager/rango-preset/src/actions/checkStatus.ts
```diff
@@ -66,7 +66,9 @@ async function checkTransactionStatus({
     const txType = getCurrentStepTxType(currentStep);
     const sourceWallet = getRelatedWallet(swap, currentStep);
     if (txType && sourceWallet) {
-      signer = context.getSigners(sourceWallet.walletType).getSigner(txType);
+      signer = (await context.getSigners(sourceWallet.walletType)).getSigner(
```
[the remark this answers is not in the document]
RyukTheCoder: Change applied

RyukTheCoder:
> I tested with metamask, When I want to sign the transaction, EVM signers do not load in lazy loading mode. It seems that it is in the initial bundle and is not properly lazy-loaded.

This Problem is resolved for evm signer in 615424aacb1ebdfd680c467fd7bfeb905b9c2ce0

RyukTheCoder:
> @RyukTheCoder Since we are going to close the previous PR, please also check related usage of `solana/web3.js` out of the signers. I think they were handled in previous PR. Also about the shared report, I'm not sure why some chnuks are duplicated in related diagram.
> 
> Also, I'm not sure if we could something about `@ethersproject` or not.
> 
> Screenshot from 2024-09-10 23-34-25


signer-solana could not be lazy loaded because we call setSolanaSignerConfig of it in the first and also its dependency @solana/web3.js could not be lazy loaded because it is used in @solflare-wallet/sdk which is a dependency of provider-solflare. It was not removed in the previous PR either.

`@ethersproject` is removed in the last commit.

nikaaru (approved):
LGTM

queue-manager/rango-preset/src/helpers.ts
```diff
@@ -1109,71 +1108,74 @@ export function signTransaction(
...
-      onFinish();
-    },
-    (error) => {
-      if (swap.status === 'failed') {
-        return;
-      }
 
-      const { extraMessage, extraMessageDetail, extraMessageErrorCode } =
-        prettifyErrorMessage(error);
+  getSigners(sourceWallet.walletType).then((walletSigners) => {
```
yeager-eren, line 1112: I think making `signTransaction` async and use async/await here would be better:
1. It communicates it runs something asynchronously explicitly
2. History for these lines will be kept.

yeager-eren (approved):
I only reviewed code and didn't test it. make sure it will be tested enough before merging.
LGTM.

queue-manager/rango-preset/src/helpers.ts
RyukTheCoder: Done in [fix: make signTransaction async in rango-preset](https://github.com/rango-exchange/rango-client/pull/861/commits/b36e3181cbb33803fd2d65120387ae079b0df5c8)

PR #866, opened 2024-09-09
fix: ensure proper cleanup of the modal component after it is removed from the DOM

Ikari-Shinji-re:
# Summary

In certain cases, we want to conditionally remove the modal component from the DOM without using the close button or clicking the backdrop. Currently, the cleanup during the unmounting of the modal is not handled correctly. After the modal is removed from the DOM, the container's overflow remains set to "hidden," causing the scrollbar to stay hidden even after the modal is gone.

Fixes # (issue)

Removing the container's overflow property when the modal is unmounted.

You can test this by linking the UI package in the Dapp. Go to the profile page and attempt to log in. After logging in and the login modal is removed from the DOM, you'll notice that the body's overflow is incorrectly handled.

# Checklist:

### DIFF
--- a/widget/ui/src/components/Modal/Modal.tsx
+++ b/widget/ui/src/components/Modal/Modal.tsx
@@ -94,6 +94,13 @@ export function Modal(props: PropsWithChildren<ModalPropTypes>) {
   }, [open]);
 
   useEffect(() => {
+    return () => {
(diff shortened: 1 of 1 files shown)


RyukTheCoder (approved):
LGTM

PR #881, opened 2024-09-24
chore: remove ledgerhq dependencies from internal packages

RyukTheCoder:
# Summary

Making ledgerhq dependencies internal made bundle size to increase about 2MBs. So we decided to remove them from internal packages.

Fixes # (issue)

- [ ] Test A
- [ ] Test B

# Checklist:

### DIFF
--- a/wallets/provider-ledger/package.json
+++ b/wallets/provider-ledger/package.json
@@ -14,7 +14,7 @@
     "src"
   ],
   "scripts": {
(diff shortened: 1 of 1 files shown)


yeager-eren:
We've internalized some ledger dependencies due to mange dependencies and its problem on vite for local development. The problem was one of sub dependencies of ledger is using ethers@5 which means we need both ethers 5 and 6 at the same time. Vite had some struggle to use both version at the same time because of its optimization flows. 

I will merge this as a hotfix to our dApp be able to released normally again, but we need to investigate why vite doesn't do tree shaking and dynamic loading.

yeager-eren (approved):
We've internalized some ledger dependencies due to mange dependencies and its problem on vite for local development. The problem was one of sub dependencies of ledger is using ethers@5 which means we need both ethers 5 and 6 at the same time. Vite had some struggle to use both version at the same time because of its optimization flows. 

I will merge this as a hotfix to our dApp be able to released normally again, but we need to investigate why vite doesn't do tree shaking and dynamic loading.

PR #886, opened 2024-09-28
perf: enable code splitting in build process

RyukTheCoder:
# Summary

Enabled code splitting in build options of esbuild. Code splitting is still a work in progress in [esbuild docs](https://esbuild.github.io/api/#splitting) because it only works with `esm` output format (which we currently use) and it has a [known ordering issue](https://github.com/evanw/esbuild/issues/399). I'm not sure if we may encounter that issue but the is a [workaround](https://github.com/evanw/esbuild/issues/399#issuecomment-1458680887) introduced to prevent that. You can follow [the tracking issue](https://github.com/evanw/esbuild/issues/16) for updates about this feature. 

After applying this change and testing it through linking `widget-embedded` package in `app` repo, I observed that bundle size is reduced to 1.4MBs.

image

Fixes # 1905

- [ ] Test A
- [ ] Test B

# Checklist:

### DIFF
--- a/scripts/build/command.mjs
+++ b/scripts/build/command.mjs
@@ -24,6 +24,8 @@ async function run() {
     // When you want to make all the packages external, and only include some specific packages as your library bundle, this will be usefull.
     // Comma separated.
     { name: 'external-all-except', type: String },
+    // Enable code splitting
+    { name: 'splitting', type: Boolean },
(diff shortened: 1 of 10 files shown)


yeager-eren (approved):
lgtm

PR #920, opened 2024-10-26
fix: change crowdin machine translation to google translate

mikasackermn:
# Summary
To switch the translation engine, the Google Machine Translation Engine ID is required


nikaaru (approved):
LGTM

nikaaru:
Please use small letters and fix in the message commit.
like this:
fix: change crowdin machine translation to google translate

PR #924, opened 2024-10-30
fix: when host is using cjs, fallback to .default() for solfare sdk

yeager-eren:
# Summary

Get correct instance for CJS hosts.

We've got a report regarding `next@13` and `module: esnext`, `moduleResolution: node` can not be run and throwing an error on `provider-solflare`.


nikaaru (approved):
LGTM

PR #935, opened 2024-11-05
feat: add routing parameters to widget config

mikasackermn:
# Summary

We needed to add `avoidNativeFee` and `maxLength` to features, but after speaking with @RanGojo and @yeager-eren, we concluded that we should create a new parameter called `routing`. This parameter will contain `avoidNativeFee`, `maxLength`, and `experimental`.

### DIFF
--- a/widget/embedded/src/types/config.ts
+++ b/widget/embedded/src/types/config.ts
@@ -121,9 +121,6 @@ export type SignersConfig = {
  * @property {'visible' | 'hidden'} [language]
  * - The visibility state for the language feature. Optional property.
  *
- * @property {'disabled' | 'enabled'} [experimentalRoute]
- * - The enablement state for the experimental route feature. Optional property.
- *
  * @property {'visible' | 'hidden'} [connectWalletButton]
  * - The visibility state for the connect wallet button feature. Optional property.
  *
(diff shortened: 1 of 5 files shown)


widget/embedded/src/types/config.ts
```diff
@@ -143,8 +140,30 @@ export type Features = Partial<
...
   >
+>;
+
+/**
+ * `Routing`
+ *
+ * @property {'disabled' | 'enabled'} [avoidNativeFee]
+ * - The enablement state for the avoid native fee feature. Optional property.
+ *
+ * @property {number} [maxLength]
```
nikaaru, line 151: I think we can express these comments more clearly
Or, for example, take it from the document here:

https://api-docs.rango.exchange/reference/getbestroutes

widget/embedded/src/types/config.ts
```diff
@@ -143,8 +140,30 @@ export type Features = Partial<
     | 'customTokens',
     'visible' | 'hidden'
   >
+>;
+
+/**
+ * `Routing`
+ *
+ * @property {'disabled' | 'enabled'} [avoidNativeFee]
+ * - If you want to exclude Swappers that have native tokens as a fee, for example when calling it from an AA account, you could set this parameter to enabled.
```
RanGojo, line 149:
```suggestion
 * - If you want to exclude swappers that have native tokens as fee, for example when your call widget using an AA account, you could set this parameter to true.
```

widget/embedded/src/types/config.ts
```diff
@@ -143,8 +140,30 @@ export type Features = Partial<
...
+ *
+ * @property {number} [maxLength]
+ * - If you want to set the maximum acceptable length of a path, you could set this parameter to a numeric value.
```
RanGojo, line 152:
```suggestion
 * - If you want to limit the maximum acceptable length of a route, you could set this parameter.
```

widget/embedded/src/types/config.ts
```diff
@@ -143,8 +140,30 @@ export type Features = Partial<
...
+ *
+ * @property {'disabled' | 'enabled'} [experimental]
+ * - The enablement state for the experimental route feature. Optional property.
```
RanGojo, line 155:
```suggestion
 * - Use this property to enable experimental routes. This settings shouldn't be used in your production environment.
```

widget/embedded/src/types/config.ts
```diff
@@ -192,22 +211,24 @@ export type TrezorManifest = {
...
  *   - 'language': Visibility state for the language.
  *   - 'customTokens': Visibility state for the custom tokens.
- *   - 'experimentalRoute': Enablement state for the experimental route.
- *
+ * @property {Routing} routing - An optional object for configuring the enablement or numeric of various features.
+ *   Keys include:
+ *   - 'maxLength': Numeric state for the max length.
+ *   - 'avoidNativeFee': Enablement state for the avoid native fee.
+ *   - 'experimental': Enablement state for the experimental route.
+ *   - 'enableCentralizedSwappers': Enablement state for the centralized swappers.
```
RanGojo, line 227: Please also update here.

widget/embedded/src/types/config.ts
mikasackermn: fixed

widget/embedded/src/types/config.ts
mikasackermn: fixed

widget/embedded/src/types/config.ts
mikasackermn: fixed

widget/embedded/src/types/config.ts
mikasackermn: fixed

nikaaru (approved):
LGTM

PR #945, opened 2024-11-12
feat: display same source and destination token warning on widget

mikasackermn:
# Summary

If the source and destination tokens are selected as the same, we should show a warning to the user.

Fixes # (issue)

Select the same token for both the source and destination, and also enter an amount

# Checklist:

### DIFF
--- a/widget/embedded/src/components/SameTokensWarning/SameTokensWarning.tsx
+++ b/widget/embedded/src/components/SameTokensWarning/SameTokensWarning.tsx
@@ -1,0 +1,31 @@
+import { i18n } from '@lingui/core';
+import { Alert, Divider, NoRouteIcon, Typography } from '@rango-dev/ui';
+import React from 'react';
+
+import { useQuoteStore } from '../../store/quote';
+import { areTokensEqual } from '../../utils/wallets';
+
+import { Container } from './SameTokensWarning.styles';
+
+export function SameTokensWarning() {
+  const { fromToken, toToken } = useQuoteStore();
+  const showWarningMessage =
+    !!fromToken && !!toToken && areTokensEqual(fromToken, toToken);
+
+  return showWarningMessage ? (
+    <Container>
+      <Divider size={10} />
+      <NoRouteIcon size={24} color="gray" />
+      <Divider size={4} />
+      <Typography variant="title" size={'small'}>
+        {i18n.t('No Routes Found')}
+      </Typography>
+      <Divider size={4} />
+      <Alert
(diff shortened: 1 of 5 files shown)


widget/embedded/src/components/NoResult/NoResult.tsx
```diff
@@ -38,6 +38,8 @@ export function NoResult(props: PropTypes) {
     fetch
   );
 
+  console.log(info);
```
Ikari-Shinji-re, line 41: Please remove logs

widget/embedded/src/hooks/useSwapInput.ts
```diff
@@ -107,7 +105,14 @@ export function useSwapInput({
       affiliatePercent,
       affiliateWallets,
     } = params;
+    if (areTokensEqual(fromToken, toToken)) {
+      console.log('....');
```
Ikari-Shinji-re, line 109: Please remove logs

Ikari-Shinji-re:
In the current implementation, the warning appears after the debounce period for the input. If the user changes the input, the warning will be cleared and then reappear after the debounce delay. I suggest leaving the existing logic for the quote unchanged and simply adding a check and warning for this specific state in the relevant component, so the warning displays immediately without delay.

widget/embedded/src/components/SameTokensWarning/SameTokensWarning.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+import { useQuoteStore } from '../../store/quote';
+import { areTokensEqual } from '../../utils/wallets';
+
+import { Container } from './SameTokensWarning.styles';
+
+export function SameTokensWarning() {
+  const { fromToken, toToken } = useQuoteStore();
+  const showWarningMessage =
+    fromToken && toToken && areTokensEqual(fromToken, toToken);
```
Ikari-Shinji-re, line 13: The type of `showWarningMessage` is inferred as `boolean | null`. Using `!!fromToken` and `!!toToken` would ensure the inferred type is strictly `boolean`, making it better aligned with the variable's name.

Ikari-Shinji-re:
Please review the UI design again. The spacing around the newly added component does not align with the design specifications

Ikari-Shinji-re (approved):

widget/embedded/src/components/SameTokensWarning/SameTokensWarning.tsx
mikasackermn: Fixed

PR #948, opened 2024-11-12
fix: bug with the chart tooltip position

mikasackermn:
# Summary

There are two issues on mobile: the tooltip position isn’t applied correctly, and it stays visible while scrolling.

Fixes # (issue)

You can link and test it in the explorer

# Checklist:

### DIFF
--- a/widget/charts/src/components/BarChart/BarChart.tsx
+++ b/widget/charts/src/components/BarChart/BarChart.tsx
@@ -131,9 +131,17 @@ export const BarChart = (props: BarChartPropTypes) => {
         hideTooltip();
       }
     }
+    function handleScroll() {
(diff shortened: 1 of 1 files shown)


widget/charts/src/components/BarChart/BarChart.tsx
```diff
@@ -131,9 +131,17 @@ export const BarChart = (props: BarChartPropTypes) => {
         hideTooltip();
       }
     }
+    function handleScroll() {
+      if (isMobile) {
```
nikaaru, line 135: In my opinion, it should not be closed by scrolling, only where it is clicked outside the content of the tooltip, it should be closed.

PR #964, opened 2024-12-09
chore: change Namespace enum to stirng

yeager-eren:
# Summary

Using a single source of truth for Namespace.

We need to use a single type for both legacy and hub implementation. To make it more general and don't needing to always keeping the namespace updated by what Rango currently support, I changed it to `string`, instead of enum. To have namespace suggestion while coding, I used a trick to suggest our supported namespaces alongside the fact that it can be any string.

Fixes # (issue)

It should be compiled correctly, shouldn't affect the functionality of working with wallets.

# Checklist:


nikaaru (approved):
LGTM

PR #965, opened 2024-12-10
feat: storing network alongside namespace for hub localstorage

yeager-eren:
# Summary

Keep full namespace input in storage.

Previously we've stored the wallet name only, for hub we extended it to store namespace name as well. But we need the full namespace input (namespace name and network) to be stored.

You can take a look at localstorage data in your browser. Phantom is currently the only wallet that is using hub.

### DIFF
--- a/wallets/core/src/namespaces/evm/actions.ts
+++ b/wallets/core/src/namespaces/evm/actions.ts
@@ -57,8 +57,9 @@ export function changeAccountSubscriber(
 ): [Subscriber<EvmActions>, SubscriberCleanUp<EvmActions>] {
   let eventCallback: EIP1193EventMap['accountsChanged'];
 
+  // subscriber can be passed to `or`, it will get the error and should rethrow error to pass the error to next `or` or throw error.
   return [
-    (context) => {
+    (context, err) => {
       const evmInstance = instance();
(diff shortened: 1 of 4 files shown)


Ikari-Shinji-re:
I noticed two issues that might not be related to the latest changes:

If the user does not import an EVM-compatible account into Phantom and selects the EVM namespace for connection, the Phantom wallet throws an error, but the wallet modal remains stuck in the connecting state.

If an EVM-compatible account is available and the user selects the EVM namespace, then confirms and rejects the wallet popup, the EVM namespace is stored as the last connected wallet for Phantom.

yeager-eren:
> I noticed two issues that might not be related to the latest changes:
> 
> If the user does not import an EVM-compatible account into Phantom and selects the EVM namespace for connection, the Phantom wallet throws an error, but the wallet modal remains stuck in the connecting state.
> 
> If an EVM-compatible account is available and the user selects the EVM namespace, then confirms and rejects the wallet popup, the EVM namespace is stored as the last connected wallet for Phantom.

The first item reported in a separate issue. The second one should be fixed by my last commit.

Ikari-Shinji-re (approved):

PR #966, opened 2024-12-10
chore: remove discover mode from hub adapter

yeager-eren:
# Summary

Remove DISCOVER_MODE from hub's adapter.

Originally, I didn't want to change the usage of `wallets-core` in `embedded` but there is a difference between legacy and hub and that is hub needs to pass `namespace` for connecting. legacy didn't have the namespace concept at all and you could connect using `network`. To make these compatible with each other, I created a hardcoded `network` -> `namespace` map in adapter.
This works but it's not and ideal way to use and keep. In this PR I removed this mode from hub and did the necessary changes to `embedded` code (specifically, `rango-preset`) to have namespace and pass it to hub. 

Fixes RF-1967

# More notes for reviewer

There are two commits, first commit removed the DISCOVER_MODE from our wallet management. The second commit introduced the concept of `namespace` to `rango-preset` so we have both `namespace` and `network` from now on.
Then, I directly used whatever `rango-preset` is returning.

In second commit, you see `blockchain` has been renamed to `namespace` and the function signature has been updated accordingly. Also the rest of changes are updating the usage of those functions. `embedded` also updated to use the namespace returned by `rango-preset`. 

This code affected the `connect` only. There are two types of `connect`:

1. Connect the whole wallet (in Wallets page).
2. Switch network

You can test these changes by try to connect you wallets and also try to test switch network in swap flow. Don't forget to switch network has two scenarios, one, you trying to do a swap and your active network is something else, you will be asked to switch network. The other one is when you are refreshing page on the middle of you swap and swap will reach an state that wants to do you swap and your active network is something else.

# Checklist:

### DIFF
--- a/queue-manager/rango-preset/src/actions/checkStatus.ts
+++ b/queue-manager/rango-preset/src/actions/checkStatus.ts
@@ -17,7 +17,7 @@ import {
 import { httpService } from '../services';
 import { notifier } from '../services/eventEmitter';
 import {
-  getCurrentBlockchainOf,
+  getCurrentNamespaceOf,
   getNextStep,
   getRelatedWallet,
   getScannerUrl,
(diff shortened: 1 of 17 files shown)


nikaaru (approved):
LGTM

PR #969, opened 2024-12-15
chore: move namespaces info to provider and remove the hardcoded mapping in wallets-shared

yeager-eren:
# Summary

Adds namespace info into providers.

Overtime we needed to have another abstraction called Namespace. In legacy, we haven't support for namespace directly and added it for Ledger and Trezor in some kind of hacky way. One of these hacks were adding a map between some fields in `WalletInfo` and generate a list for namespaces in `Namespace` modal.

In this PR, I introduced two new fields called `needsNamespace` and `needsDerivationPath` for `WalletInfo`. Providers now can define the list  we show in Namespace by themselves.

Fixes RF-1935

It shouldn't affect the functionality. namespace and derivation path should works as before. Ledger, Trezor and Phantom have support for these functionalities i mentioned.

# Checklist:


widget/embedded/src/utils/hub.ts
```diff
@@ -1,22 +1,21 @@
 import type { CommonNamespaceKeys } from '@rango-dev/wallets-core';
-
-import { Namespace } from '@rango-dev/wallets-shared';
+import type { Namespace } from '@rango-dev/wallets-core/namespaces/common';
 
 export function convertCommonNamespacesKeysToLegacyNamespace(
```
mikasackermn, line 4: This function isn't used anywhere. If it isn't needed, it would be better to remove it.

yeager-eren: good catch! removed the file.

mikasackermn (approved):
LGTM

widget/embedded/src/utils/hub.ts
mikasackermn: I think you forgot to remove it

yeager-eren: you are right. i pushed a commit right now. https://github.com/rango-exchange/rango-client/pull/969/commits/fda14bd29731b1d67302af2079fbb4c699b00404

PR #984, opened 2025-01-07
fix: update connected namespaces in storage on switch account

RyukTheCoder:
# Summary

Fixed two bugs:

Update connected namespaces for wallets in storage on switch account

Fixes # (RF-2098)

Tested this change by connecting phantom wallet on solana namespace. Observed that solana namespace is set in local storage in `hub-v1-last-connected-wallets` for phantom wallet. After that I switch to an account which only contains ethereum address and observed that local storage updated as result of disconnecting wallet.

# Checklist:

### DIFF
--- a/wallets/react/src/hub/lastConnectedWallets.ts
+++ b/wallets/react/src/hub/lastConnectedWallets.ts
@@ -8,7 +8,7 @@ import {
 } from './constants.js';
 
 export interface NamespaceInput {
-  namsepace: Namespace;
+  namespace: Namespace;
   network: string | undefined;
 }
 
@@ -50,6 +50,12 @@ export class LastConnectedWalletsFromStorage {
       return this.#listFromHub();
     } else if (this.#storageKey === LEGACY_LAST_CONNECTED_WALLETS) {
       return this.#listFromLegacy();
+    }
+    throw new Error('Not implemented');
+  }
+  removeNamespacesFromWallet(providerId: string, namespaceIds: string[]) {
+    if (this.#storageKey === HUB_LAST_CONNECTED_WALLETS) {
+      return this.#removeNamespaceFromWalletHub(providerId, namespaceIds);
     }
     throw new Error('Not implemented');
   }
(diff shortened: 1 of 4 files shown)


wallets/react/src/hub/lastConnectedWallets.ts
```diff
@@ -53,6 +53,9 @@ export class LastConnectedWalletsFromStorage {
     }
     throw new Error('Not implemented');
   }
+  removeNamespacesFromWallet(providerId: string, namespaceIds: string[]) {
+    return this.#removeNamespaceFromWalletHub(providerId, namespaceIds);
```
yeager-eren, line 57: this class is going to be an adapter for hub and legacy storage. So like other methods you need to add a condition and run this on hub storage only. in your case, there is no implemention for legacy so the code will be like this probably:
```
   if (this.#storageKey === HUB_LAST_CONNECTED_WALLETS) {
      //
    } 
    throw new Error('Not implemented');
    ```

File: wallets/react/src/hub/lastConnectedWallets.ts [hunk 2]
```diff
@@ -105,6 +108,29 @@ export class LastConnectedWalletsFromStorage {
...
+    const storageState = persistor.getItem(this.#storageKey) || {};
+
+    const currentProviderNamespaces = storageState[providerId];
+    const newProviderNamespaces = currentProviderNamespaces.filter(
+      (namespace) => !namespaceIds.includes(namespace.namespace)
+    );
+
+    // If no namespace remains, remove wallet
+    if (!newProviderNamespaces.length) {
+      delete storageState[providerId];
```
yeager-eren, line 125: you could use `this.#removeWalletsFromHub([providerId])` to only have one place that have the logic for removing wallet.

yeager-eren: or may you can simplify the whole if/else[ by always removing wallet and add it again:

```
this.#removeWalletsFromHub([providerId]);
this.#addWalletToHub(providerId, newProviderNamespaces)
```

yeager-eren (changes requested):

wallets/react/src/hub/lastConnectedWallets.ts
RyukTheCoder: Tnx applied

RyukTheCoder: Tnx, applied

wallets/react/src/hub/lastConnectedWallets.ts
RyukTheCoder: Tnx, applied

yeager-eren (approved):
lgtm

PR #987, opened 2025-01-12
fix: ctrl wallet is changing 'keplr' named function to undefined

yeager-eren:
# Summary

Avoid CTRL to override the function named `keplr`.

We've faced with a weird bug where Having CTRL and Keplr at the same time causing an app crash. The first investigation was changing the getInstance from function declaration to arrow function in keplr provider will fix the issue. After more investigation I tried to see the difference between these two function declaring to find the root cause.

here is the transpiled verson for function declariont (simplified):

```js
var c = Object.defineProperty;
var r = (e, o) => c(e, "name", {
    value: o,
    configurable: !0
});

function s() {
    return window.keplr || null
}
r(s, "keplr");

// then export s as getInstance
```

and arrow function:
```js
r(() => window.keplr || null, "getInstance")

// then export.
```

the only difference is in the first one, the function will keep it's original name `keplr` but in the second one it will be named as `getInstance` (transpilers do this for better debugging).

I changed the name of function and issue has been solved! CTRL extension is changing `keplr` function to `undefined` somehow!

Fixes RF-2119

Try to active both CTRL and Keplr wallets. app shouldn't crash.

# Checklist:

### DIFF
--- a/wallets/core/src/legacy/wallet.ts
+++ b/wallets/core/src/legacy/wallet.ts
@@ -318,6 +318,13 @@ class Wallet<InstanceType = any> {
   }
 
   onInit() {
+    // some times functions can be overridden by wallets. see rf-2119
+    if (!this.actions.getInstance) {
+      throw new Error(
+        `Provider hasn't defined how to get wallet's instance. provider: ${this.options.config.type} on: onInit`
+      );
+    }
+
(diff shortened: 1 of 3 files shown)


RyukTheCoder (approved):
LGTM

PR #988, opened 2025-01-13
fix: remove namespace from storage on auto connect failure

RyukTheCoder:
# Summary

Remove namespace from storage on auto connect failure. 

Fixes # (RF-2118)

Tested this change by changin phantom account to an account containg only Solana address and setting `{"phantom":[{"namespace":"Solana"},{"namespace":"EVM"}]}` value for `hub-v1-last-connected-wallets` in localStorage and refreshing the app. Observed that Solana namespace connected successfully and also EVM namespace removed from localstorage.

# Checklist:

### DIFF
--- a/wallets/react/src/hub/autoConnect.ts
+++ b/wallets/react/src/hub/autoConnect.ts
@@ -15,9 +15,14 @@ import {
 
 import { HUB_LAST_CONNECTED_WALLETS } from '../legacy/mod.js';
 
-import { sequentiallyRun } from './helpers.js';
+import { runSequentiallyWithoutFailure } from './helpers.js';
 import { LastConnectedWalletsFromStorage } from './lastConnectedWallets.js';
 import { convertNamespaceNetworkToEvmChainId } from './utils.js';
+
+// Getting connected wallets from storage
+const lastConnectedWalletsFromStorage = new LastConnectedWalletsFromStorage(
+  HUB_LAST_CONNECTED_WALLETS
+);
 
 /**
  * Run `.connect` action on some selected namespaces (passed as param) for a provider.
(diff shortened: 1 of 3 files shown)


wallets/react/src/hub/autoConnect.ts
```diff
@@ -60,21 +65,49 @@ async function eagerConnect(
...
+  const connectNamespacesPromises = targetNamespaces.map(
+    ([info, namespace]) => {
+      const evmChain = legacyIsEvmNamespace(info)
+        ? convertNamespaceNetworkToEvmChainId(info, allBlockChains || [])
+        : undefined;
+      const chain = evmChain || info.network;
 
-    return async () => await namespace.connect(chain);
-  });
+      return async () => await namespace.connect(chain);
```
yeager-eren, line 75: since you are making a list of failed promises using `instanceof Error`, ensure you will throw an error in any cases by having something like this:
```
.catch(e => {
if(e instanceof Error){
throw e;
}
throw new Error(e)
})
```

yeager-eren: write this assumption as a comment on top of it also.

wallets/react/src/hub/autoConnect.ts
```diff
@@ -60,21 +65,49 @@ async function eagerConnect(
...
+  });
+
+  if (failedNamespaces.length > 0) {
+    lastConnectedWalletsFromStorage.removeNamespacesFromWallet(
+      type,
+      failedNamespaces.map((namespace) => namespace.namespace)
+    );
+  }
+
+  const atLeastOneNamespaceConnectedSuccessfully = successNamespaces.length > 0;
```
yeager-eren, line 105: you can remove `successNamespaces` and consoder it as:
```suggestion
  const atLeastOneNamespaceConnectedSuccessfully = connectNamespacesResult.length - failedNamespaces.length > 0;
```

I'm suggesting this since you only need the length, and not the input itselt.

wallets/react/src/hub/autoConnect.ts
```diff
@@ -60,21 +65,49 @@ async function eagerConnect(
...
+  if (!atLeastOneNamespaceConnectedSuccessfully) {
+    throw new Error(`No namespace connected for ${type}`);
+  }
+
+  return connectNamespacesResult.filter((result) => !(result instanceof Error));
```
yeager-eren, line 110:
```suggestion
const successfulResult = connectNamespacesResult.filter((result) => !(result instanceof Error));
  return successfulResult
```

wallets/react/src/hub/autoConnect.ts
```diff
@@ -161,26 +190,15 @@ export async function autoConnect(deps: {
           providerName,
         }).catch((e) => {
           walletsToRemoveFromPersistance.push(providerName);
-          throw e;
+          console.warn(e);
         })
       );
     });
 
     await Promise.allSettled(eagerConnectQueue);
```
yeager-eren, line 198: with your change on remving `throw`, the promise never fails, so you can change this to Promise.all  i guess

wallets/react/src/hub/helpers.ts
```diff
@@ -47,9 +47,13 @@ export async function sequentiallyRun<T extends () => Promise<unknown>>(
 ): Promise<Array<T extends () => Promise<infer R> ? R : never>> {
   const result = await promises.reduce(async (prev, task) => {
     const previousResults = await prev;
-    const taskResult = await task();
+    try {
+      const taskResult = await task();
 
-    return [...previousResults, taskResult];
+      return [...previousResults, taskResult];
+    } catch (error) {
```
yeager-eren, line 54: you've change the behavior and it doesn't reflect in the name. you can create a new function called `sequentiollyRunWithoutFailing`

yeager-eren (changes requested):

wallets/react/src/hub/autoConnect.ts
RyukTheCoder: Good point, applied

wallets/react/src/hub/autoConnect.ts
RyukTheCoder: Applied

wallets/react/src/hub/helpers.ts
RyukTheCoder: Tnx, changed to `runSequentiallyWithoutFailure`

wallets/react/src/hub/autoConnect.ts
RyukTheCoder: Yes throwing error was causing a crash. Changed to `Promise.all`

wallets/react/src/hub/autoConnect.ts
RyukTheCoder: Didn't understand. Can you make it more clear plz?

RyukTheCoder: Thanks for your insightful comment. Applied.

yeager-eren (approved):
nice job!

PR #989, opened 2025-01-13
fix: fix incorrect wallet state after switch account

RyukTheCoder:
# Summary

This pull request addresses an inconsistency related to disconnecting Phantom wallets after switching to an account that does not contain both Solana and EVM addresses. Previously, when transitioning from an account with both addresses to one containing only the EVM address, only the Solana account would disconnect. Conversely, moving from an account with both addresses to one with only the Solana address resulted in both Solana and EVM accounts disconnecting.

This PR resolves the issue by ensuring that when switching from an account containing both addresses to one with only either the EVM or Solana address, only the connection to the unavailable address is disconnected. This enhancement provides a more coherent and expected behavior during account transitions.

Fixes # 2141

- [ ] Test A
- [ ] Test B

# Checklist:

### DIFF
--- a/wallets/react/src/hub/utils.ts
+++ b/wallets/react/src/hub/utils.ts
@@ -110,6 +110,7 @@ export function checkHubStateAndTriggerEvents(
     let hasAccountChanged = false;
     let hasNetworkChanged = false;
     let hasProviderDisconnected = false;
+    let anyNamespaceIsConnected = false;
     // It will pick the last network from namespaces.
     let maybeNetwork = null;
     const disconnectedNamespacesIds: string[] = [];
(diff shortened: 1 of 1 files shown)


wallets/react/src/hub/utils.ts
```diff
@@ -138,15 +141,30 @@ export function checkHubStateAndTriggerEvents(
...
+        } else {
+          anyNamespaceDisconnected = true;
+        }
+
+        hasAccountChanged = true;
+      } else if (currentNamespaceState.accounts) {
+        anyNamespaceIsConnected = true;
+        const formattedAddresses = currentNamespaceState.accounts.map(
+          fromAccountIdToLegacyAddressFormat
+        );
```
yeager-eren, line 154: `accounts` is initialized with `[]` so it will always be true. what `else` do here?

wallets/react/src/hub/utils.ts
```diff
@@ -138,15 +141,30 @@ export function checkHubStateAndTriggerEvents(
           } else {
             accounts = [...formattedAddresses];
           }
+        } else {
+          anyNamespaceDisconnected = true;
+        }
+
+        hasAccountChanged = true;
+      } else if (currentNamespaceState.accounts) {
```
yeager-eren, line 149: What do you think to restructure the condition to be like this:

```js
if (currentNamespaceState.accounts) {
         if (
        previousNamespaceState.accounts?.slice().sort().toString() !==
        currentNamespaceState.accounts?.slice().sort().toString()
      ) {
      // ...
      }


}else{
anyNamespaceIsConnected = true;
}
```

wallets/react/src/hub/utils.ts
```diff
@@ -202,6 +220,7 @@ export function checkHubStateAndTriggerEvents(
       );
     }
     if (hasAccountChanged) {
+      onUpdateState(providerId, Events.ACCOUNTS, null, coreState, eventInfo);
```
yeager-eren, line 223: why is needed? why we firing two events in a row?

wallets/react/src/hub/utils.ts
RyukTheCoder: Applied in https://github.com/rango-exchange/rango-client/pull/989/commits/355c9eed2bceed9e089e4ec0527ff6f1a64891f9

wallets/react/src/hub/utils.ts
RyukTheCoder: This is due to a problem with  logic of wallets slice actions which I was not sure if I can change that. My requirement is that there is a wallet in store which I want to update accounts for it. The new accounts are calculated here and the accounts for wallet should just update to this calculated value. To make that happen I want to call `onUpdateState` with new accounts and expect that accounts get updated with new value but, it is not happening because the logic in `onUpdateState` and slice actions is like that if we call onUpdateState with (Events.ACCOUNTS, null) wallet gets disconnected and if we call it with (Events.ACCOUNTS, newAccounts), the new accounts just get added to last connected accounts in `addConnectedWallet` action an `wallets` slice. Also in `addConnectedWallet` there is a comment that explained that because of some assumptions we decided to implement it like the way it is.

yeager-eren: i see. thanks for the explanation. please add a comment on top of it.

wallets/react/src/hub/utils.ts
RyukTheCoder: Please check this commit https://github.com/rango-exchange/rango-client/pull/989/commits/fd4449863aa7dd034c90a6b65148652d64bae630. I changed the type of `accounts` to `string[]` because I didn't find out why it should be nullable at all. Also I checked every usage of `onUpdateState` and didn't find any usage for accounts field which we are propagating in `coreState` but considering that we are exporting `WidgetProvider` which accepts `onUpdateState` as a prop, I made a check on that to set `accounts` of `coreState` to `null` if it is empty.

yeager-eren: my bad. don't need to change anything

yeager-eren (approved):
lgtm

wallets/react/src/hub/utils.ts
RyukTheCoder: Tnx, comment added in https://github.com/rango-exchange/rango-client/pull/989/commits/fc917ac46e4298797223de6d64989b33b1cf05ec

PR #990, opened 2025-01-13
fix: close wallet connection modal after connect with namespace

RyukTheCoder:
# Summary

Wallet connection modal was not closing automatically after connecting a wallet which needs namespace selection or derivation path entering. It is fixed by calling `afterConnect` method after connecting wallet with namespace or derivation path.

Fixes # 2147

Tested by connecting a wallet which needs namespace or derivation path (such as phantom and ledger) and observing that wallet connection modal closes automatically after successful connect.

# Checklist:

### DIFF
--- a/widget/embedded/src/components/StatefulConnectModal/StatefulConnectModal.tsx
+++ b/widget/embedded/src/components/StatefulConnectModal/StatefulConnectModal.tsx
@@ -26,8 +26,8 @@ const DELAY_SHOWING_MODAL_FOR = 300;
 interface PropTypes {
   wallet: WalletInfoWithExtra | undefined;
   onClose: () => void;
-  // When `handleConnect` runs **successfully**, this will be called afterwards.
-  onConnect?: (result: Result) => void;
+  // When connecting wallet is executed **successfully**, this will be called afterwards.
+  onConnect?: () => void;
 }
 
 export function StatefulConnectModal(props: PropTypes) {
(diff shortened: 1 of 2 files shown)


widget/embedded/src/components/StatefulConnectModal/StatefulConnectModal.tsx
```diff
@@ -77,6 +79,27 @@ export function StatefulConnectModal(props: PropTypes) {
...
+    if (resultIsConnected && !isImmediatelyConnected) {
+      successModalTimerId.current = setTimeout(
+        handleClosingModal,
+        KEEP_SUCCESS_MODAL_FOR
+      );
+    } else if (resultIsDisconnected) {
+      handleClosingModal();
+    }
+
+    if (props.onConnect) {
```
yeager-eren, line 98: You spot the bug very well, `afterConnected`, there are one points on your approach:
`props.onConnect` will be called on each state/step, but it need only to be run on successful connection which means at the of the process, only once. 

I suggest to see it as an effect. keep your afterConnected and instead of calling `.then(afterConnected)`, do this:

```
useEffect(afterConnected, [getState().status])
```

Notes:
- you can also define a ref using useRef to update and access to `isConnected`.
- onConnect should be run once so in your afterConnected you need to have a condition like: !resultIsConnected || !resultIsDisconnected return; to stop proceeding.

yeager-eren (changes requested):

widget/embedded/src/components/StatefulConnectModal/StatefulConnectModal.tsx
RyukTheCoder: Tnx, I improved the code to call `props.onConnect` only at the end of the process. About the other suggestion, I didn't understand how we can see this as an effect. `isConnected` status is a part of the result of each of `handleConnect`, `handleNamespace` and `handleDerivationPath` and `getState().status` returns the status of modal which is of this type:
```
status: 'init' | 'namespace' | 'derivationPath';
```
So it seems that listening on the status of modal to call `afterConnected` can not be helpful.

yeager-eren (approved):
thanks.

PR #991, opened 2025-01-18
fix: adding balances to ConnectedWallet for WidgetInfo

yeager-eren:
# Summary

Add `balances` to WidgetInfo's `wallets.details`.
 
In https://github.com/rango-exchange/rango-client/pull/852, we refactored the wallets store and a part of that was logic related to `ConnectedWallet`. After refactoring, we've removed a field from the object called `balances`. since this was exposed to outside world using `WalletInfo` we need to revert it back for backward compatibility. (used in our dApp as well.)

Don't need to test, it will be tested in https://github.com/rango-exchange/app-v2/pull/145.

# Checklist:


RyukTheCoder (approved):
LGTM

PR #994, opened 2025-01-20
feat: add sending solana transaction on multiple nodes

RyukTheCoder:
# Summary

Added extra Solana RPC URLs to increase success rate of send Solana transactions.

- [ ] Test A
- [ ] Test B

# Checklist:

### DIFF
--- a/signers/signer-solana/src/utils/helpers.ts
+++ b/signers/signer-solana/src/utils/helpers.ts
@@ -1,19 +1,37 @@
+import type { RpcNodeUrls } from './types.js';
+
 import { Connection } from '@solana/web3.js';
 
 import { getSolanaSignerConfig } from '../config.js';
 
 const IS_DEV = !process.env.NODE_ENV || process.env.NODE_ENV === 'development';
-const SOLANA_RPC_URL = !IS_DEV
+// This main rpc url is used to fetch latest blockhash, simulate transaction, etc
+const DEFAULT_SOLANA_RPC_URL = !IS_DEV
   ? 'https://purple-practical-friday.solana-mainnet.quiknode.pro/d94ab067f793d48c81354c78c86ae908d9fc1582/'
   : 'https://fluent-still-scion.solana-mainnet.discover.quiknode.pro/fc8be9b8ac7aea382ec591359628e16d8c52ef6a/';
 
-export function getSolanaConnection(): Connection {
-  const customRPC = getSolanaSignerConfig('customRPC');
-
-  return new Connection(customRPC || SOLANA_RPC_URL, {
+export function getSolanaConnection(url: string): Connection {
+  return new Connection(url, {
     commitment: 'confirmed',
     disableRetryOnRateLimit: false,
   });
+}
+
+export function getSolanaRpcNodes(): RpcNodeUrls {
+  const customRPC = getSolanaSignerConfig('customRPC');
+  let list: string[] = [];
+
+  if (customRPC) {
+    if (!Array.isArray(customRPC)) {
+      list = [customRPC];
+    } else if (customRPC.length) {
+      list = customRPC;
+    }
+  } else {
+    list = [DEFAULT_SOLANA_RPC_URL];
+  }
+
+  return { main: list[0], list };
 }
 
 export const wait = async (time: number) =>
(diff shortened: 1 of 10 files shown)


signers/signer-solana/src/utils/send.ts
```diff
@@ -1,20 +1,36 @@
...
+    const connection = new Connection(url, {
+      commitment: 'confirmed',
+      disableRetryOnRateLimit: false,
+    });
+    await connection.sendRawTransaction(serializedTransaction, SEND_OPTIONS);
+  });
+}
+
 // https://github.com/jup-ag/jupiter-quote-api-node/blob/main/example/utils/transactionSender.ts
 export async function transactionSenderAndConfirmationWaiter({
```
yeager-eren, line 35: instead of `conncection: Connection` we can get a pool of connections. so we need to change to `pool: Connection[]`.

they all be should call at the same time and in parallel so failing one, should fails the other. instead of `await sendTransactionOnMultipleNodes(serializedTransaction);` you can do this:

```js
await Promise.allSetteled(pool.map(connection => connection.sendRawTransaction(serializedTransaction, SEND_OPTIONS).catch(e => console.warn(`Failed to resend transaction: ${e}`)))
```

and remove the try catch since we don't need neither result nor error.

signers/signer-solana/src/utils/helpers.ts
```diff
@@ -3,18 +3,34 @@ import { Connection } from '@solana/web3.js';
...
   ? 'https://purple-practical-friday.solana-mainnet.quiknode.pro/d94ab067f793d48c81354c78c86ae908d9fc1582/'
   : 'https://fluent-still-scion.solana-mainnet.discover.quiknode.pro/fc8be9b8ac7aea382ec591359628e16d8c52ef6a/';
+// This list is used for broadcasting transactions
+const SOLANA_RPC_URL_LIST = [
+  'https://purple-practical-friday.solana-mainnet.quiknode.pro/d94ab067f793d48c81354c78c86ae908d9fc1582/',
+  'https://mainnet.helius-rpc.com/?api-key=5b616869-7eed-4fbb-9d3f-f93959a67143',
+  'https://solana-api.instantnodes.io/token-vRnXXHywZM8VMbjTKm3UnKm0X5j8haLc',
+];
 
 export function getSolanaConnection(): Connection {
```
yeager-eren, line 17:
```suggestion
export function getSolanaConnection(url: string): Connection {
```

yeager-eren: then replace with `customRPC || MAIN_SOLANA_RPC_URL`

yeager-eren: then you will need a function to return rpc urls like what you did in getSolanaRpcList but need to be more generalized to returns a list and a main RPC.

it will be like:

```ts
interface RpcNodeUrls{
main: string;
list: string;
}

function getRpcNodes():RpcNodeUrls{
  const customRPC = getSolanaSignerConfig('customRPC');
  let list = [];
  if(customRPC){
   list = typeof customRPC === "string" ? [customRPC] : customRPC;
  }else{
   list = [MAIN_SOLANA_RPC_URL]
  }
  
  return {main: list[0], list}
}
```

yeager-eren (changes requested):
I think we need to take these steps in order:

1. extend widget config's customRpc from accepting `string` to `string | string[]` for backward compat.
2. we need a function to called getRpcNodes to return an string for single connection we need (`main`) and also returns a list of available nodes (`list`).
3. make `getSolanaConnection` more general by accepting a url.
4. update the usage of the function and call it like `getSolanaConnection(getRpcNodes().main)`
5. update `transactionSenderAndConfirmationWaiter` usage to pass a pool of connection instead of one, like:
```js
const pool = getRpcNodes.list.map(url => getSolanaConnection(url));
transactionSenderAndConfirmationWaiter({pool, serializedTransaction})
```

signers/signer-solana/src/utils/helpers.ts
RyukTheCoder: Tnx applied

signers/signer-solana/src/utils/send.ts
RyukTheCoder: Tnx applied

yeager-eren (approved):
lgtm

PR #998, opened 2025-01-22
fix: remove null from wallet name

arlert-armin:
# Summary

If the transaction gets blocked after it has started, it can't be resumed, and the user can't do anything more on Solana wallets.

Fixes # (issue)
I've removed null from the wallet name so the queue manager can resume its job. However, the transaction would be expired. 

# Checklist:

### DIFF
--- a/wallets/shared/src/helpers.ts
+++ b/wallets/shared/src/helpers.ts
@@ -216,7 +216,8 @@ export function detectMobileScreens(): boolean {
  */
 
 export function splitWalletNetwork(input: string): string[] {
-  const removedAddressInput = input?.split(':')[0] || '';
+  const removedNullInput = input?.split('-null').join('');
+  const removedAddressInput = removedNullInput?.split(':')[0] || '';
   const splittedInput = removedAddressInput.split('-');
   const network = splittedInput[splittedInput.length - 1];
   const walletNetwork = splittedInput.slice(0, -1);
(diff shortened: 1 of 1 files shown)


wallets/shared/src/helpers.ts
```diff
@@ -211,7 +211,8 @@ export function detectMobileScreens(): boolean {
  */
 
 export function splitWalletNetwork(input: string): string[] {
```
RyukTheCoder, line 213:
```suggestion
export function splitWalletNetwork(input: string): string[] {
  const removedAddressInput = input?.split(':')[0] || '';
  const splittedInput = removedAddressInput.split('-');
  const network = splittedInput[splittedInput.length - 1];
  const walletNetwork = splittedInput.slice(0, -1);

  if (
    walletNetwork[walletNetwork.length - 1] === network ||
    walletNetwork[walletNetwork.length - 1] === 'null'
  ) {
    walletNetwork.pop();
  }
  const wallet = walletNetwork.join('-');

  return [wallet, network];
}
```

RyukTheCoder (changes requested):

wallets/shared/src/helpers.ts
arlert-armin: I've applied the requested changes.

RyukTheCoder (approved):
LGTM

PR #993, opened 2025-01-18
fix: fix total balance calculation for WidgetInfo

yeager-eren:
# Summary

Fix total balance amount.

It should consider decimals first, then calculate the total balance.

# How did you test this change?

No need to test. will be tested on dApp.

# Checklist:

- [ ] I have performed a self-review of my code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] Implemented a user interface (UI) change, referencing our Figma design to ensure pixel-perfect precision.

PR #999, opened 2025-01-26
fix: usdValue shouldn't be formatted to calculate the total value

yeager-eren:
# Summary

In [another PR](https://github.com/rango-exchange/rango-client/pull/993) I tried to fix total balance for our dApp integration. I used `formatBalance` to convert amount (pair of amount and decimals) to be represented in decimal format. That function do this but also will format the output to separate numbers by comma (`1,350`). For the calculation I only need the represented value. 

Will be tested in https://github.com/rango-exchange/app-v2/pull/145

# Checklist:


RyukTheCoder (approved):
LGTM

PR #1005, opened 2025-01-27
fix: resolve issue with connecting to solana namespace

RyukTheCoder:
# Summary

Resolved an issue in the Phantom wallet related to connecting to the Solana namespace. Previously, after disconnecting the wallet and attempting to reconnect to both Solana and EVM namespaces, errors occurred with the Solana instance. This problem was fixed by modifying the connection process to the Solana namespace, ensuring that account requests are not sent if the public key is already available.

Connected to Solana namespace, then disconnected and then connected to both namespaces. Observed that both namespaces get connected without any error.

# Checklist:

### DIFF
--- a/wallets/shared/src/helpers.ts
+++ b/wallets/shared/src/helpers.ts
@@ -146,10 +146,15 @@ export const evmChainsToRpcMap = (
 };
 
 export const getSolanaAccounts: Connect = async ({ instance }) => {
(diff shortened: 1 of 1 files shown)


yeager-eren (approved):
lgtm. but make sure it will work in all the wallets that use this function, frontier, ledget, ...

PR #1014, opened 2025-02-02
fix: make connect namespaces sequential

RyukTheCoder:
# Summary

This pull request addresses an issue with connecting to namespaces on the Phantom wallet. That issue was happening in this scenario: Imagine you have two accounts on Phantom which one of them supports both evm and solana and the other one only supports solana address and widget does not have access to connect each of them. First, we try to connect to the account which only supports solana. After it gets connected successfully, we disconnect it and switch to the other account which supports both namespaces on Phantom. Now, if we try to connect to that account, it fails on connecting to solana namespace. After some investigation, it turned out that there is a problem on Phantom extension which fails to address two parallel requests in this scenario.

In this PR, connecting to namespaces has made sequential to address this issue. 

1. Repeat the scenario described above with both accounts.
2. Verify that connections to both namespaces are established successfully after the changes implemented in this PR.

# Checklist:

### DIFF
--- a/wallets/react/package.json
+++ b/wallets/react/package.json
@@ -37,7 +37,8 @@
   "dependencies": {
     "@rango-dev/wallets-core": "^0.40.1-next.11",
     "@rango-dev/wallets-shared": "^0.41.1-next.4",
-    "rango-types": "^0.1.78"
+    "rango-types": "^0.1.78",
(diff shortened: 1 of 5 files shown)


yeager-eren (approved):
lgtm

PR #1016, opened 2025-02-03
fix: fix balances not updating properly after switch account

RyukTheCoder:
# Summary

This pull request addresses an issue with updating balances after switching accounts on the Phantom wallet. That issue was happening in this scenario:  Imagine you have two accounts on Phantom which one of them supports both evm and solana and the other one only supports solana address. First, we connect on Phantom to the wallet containing both solana abd evm addresses. In result of this, balances for evm and solana accounts get fetched and updated. Now if we switch to the account containing only Solana address. At first Solana address gets updated and in result of that balances related to old Solana address get removed and fetch balance API gets called for new Solana address and old Evm addresses. Then, old evm addresses get removed and in result of that balances related to old Evm addresses get removed and fetch balance API gets called for new Solana address. In this time the first api call gets settled and updates the balances with values related to new Solana address and old Evm addresses. At this state, balances for old Evm addresses should not be included in widget store because Phantom is not connected to those addresses anymore.
This issue was resolved by refactoring wallet slice such that  fetch balance API call not getting called for balances which are available in store anymore. This makes the first fetch balance API get called only for new Solana address and resolves the issue.

1. Repeat the scenario described above with both accounts.
2. Verify that balances get updated properly.

# Checklist:

### DIFF
--- a/wallets/react/src/hub/utils.ts
+++ b/wallets/react/src/hub/utils.ts
@@ -223,8 +223,6 @@ export function checkHubStateAndTriggerEvents(
       );
     }
     if (hasAccountChanged) {
-      // This event is triggered to clear wallet state and after that set new accounts for wallet
-      onUpdateState(providerId, Events.ACCOUNTS, null, coreState, eventInfo);
       onUpdateState(
         providerId,
         Events.ACCOUNTS,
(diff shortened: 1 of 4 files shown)


yeager-eren:
Close this in favor of https://github.com/rango-exchange/rango-client/pull/1025

PR #1022, opened 2025-02-08
feat: implement search functionality for custom tokens

Ikari-Shinji-re:
# Summary

implement search functionality for custom tokens

If the user enters a query in the search input of the token list and the following criteria are met, we should call the search endpoint to find custom tokens:  

- The total number of tokens in the list is below 20.  
- The search query has at least 3 characters.  

To prevent unnecessary API calls, we can implement a debounce mechanism.  

Additional features:  
- Display a warning badge in the token selector and on the home page when a custom token is selected.  
- Implement an import function similar to the custom token feature in settings.  
- If a custom token is imported while on the token selector page, it should be automatically selected, and the user should be redirected back to the home page.

- [ ] Test A
- [ ] Test B

# Checklist:

### DIFF
--- a/widget/embedded/src/components/ImportCustomToken/ImportCustomToken.tsx
+++ b/widget/embedded/src/components/ImportCustomToken/ImportCustomToken.tsx
@@ -1,0 +1,121 @@
+import type { PropTypes } from './ImportCustomToken.types';
+
+import { i18n } from '@lingui/core';
+import { Button, Divider, MessageBox } from '@rango-dev/ui';
+import React, { useEffect, useState } from 'react';
+
+import { useAppStore } from '../../store/AppStore';
+import { getContainer } from '../../utils/common';
+import { WatermarkedModal } from '../common/WatermarkedModal';
+import { CustomTokenModal } from '../CustomTokenModal';
+
+export function ImportCustomToken(props: PropTypes) {
+  const {
+    token,
+    blockchain,
+    error,
+    address,
+    fetchCustomToken,
+    onCloseErrorModal,
+    onImport,
+    onExitErrorModal,
+    onExitImportModal,
+  } = props;
+  const { setCustomToken } = useAppStore();
+  const [showErrorModal, setShowErrorModal] = useState<boolean>(false);
+  const [showImportModal, setShowImportModal] = useState<boolean>(false);
+  const [retryClicked, setRetryClicked] = useState(false);
+
+  const getCustomToken = () => {
+    if (blockchain) {
+      void fetchCustomToken?.({
+        blockchain: blockchain.name,
+        tokenAddress: address,
+      });
+    }
+  };
+
+  const closeErrorModal = () => {
+    if (error?.type !== 'network-error') {
+      onCloseErrorModal?.();
+    }
+    setShowErrorModal(false);
+  };
+
+  const handleErrorModalButtonClick = () => {
+    setRetryClicked(true);
+    closeErrorModal();
+  };
+
+  const handleExit = () => {
+    if (retryClicked && error?.type === 'network-error') {
+      setRetryClicked(false);
+      void getCustomToken();
+    }
+    onExitErrorModal();
+  };
+
+  const handleSubmit = () => {
+    if (token) {
+      setCustomToken(token);
+      onImport();
+    }
+  };
+
+  useEffect(() => {
+    if (!!error) {
+      setShowErrorModal(true);
+    }
+  }, [error]);
+
+  useEffect(() => {
+    if (blockchain && token) {
+      setShowImportModal(true);
+    }
+  }, [blockchain, token]);
+
+  return (
+    <>
+      <WatermarkedModal
+        open={showErrorModal}
+        dismissible
+        onClose={closeErrorModal}
+        onExit={handleExit}
+        container={getContainer()}>
+        <MessageBox
+          title={error?.title ?? ''}
+          type="error"
+          description={
+            error?.message || i18n.t('Failed Network, Please retry.')
+          }>
+          <Divider size={40} />
+          <Divider size={10} />
+          {/* eslint-disable-next-line jsx-id-attribute-enforcement/missing-ids */}
+          <Button
+            id={`widget-add-custom-token-${
+              error?.type === 'network-error' ? 'retry' : 'add-another'
+            }-btn`}
+            variant="contained"
+            size="large"
+            type="primary"
+            fullWidth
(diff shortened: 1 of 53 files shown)


nikaaru:
image

Please use a default image for tokens without images.

nikaaru:
image

When searching with query `arsd`, the code throws:
`TypeError: Cannot read properties of undefined (reading 'color')`

nikaaru:
image

The import button's styling doesn't match the Figma design specs.

nikaaru:
image
ss

nikaaru:
image

The token image is currently distorted due to setting both width and height to 100%. This forces the image to stretch to fill its container without maintaining its aspect ratio.

widget/embedded/src/hooks/useSearchCustomTokens/useSearchCustomTokens.ts
```diff
@@ -0,0 +1,74 @@
...
+import { debounce } from '../../utils/common';
+
+import { DEBOUNCE_DELAY } from './useSearchCustomTokens.constants';
+
+export function useSearchCustomTokens() {
+  const abortController = useRef<AbortController | null>(null);
+  const { customTokens } = useAppStore();
+  const [loading, setLoading] = useState(false);
+  const [tokens, setTokens] = useState<Token[]>([]);
+  const [error, setError] = useState<string | null>();
```
nikaaru, line 17: Isn't better to fix state initialization for the error state?
```suggestion
  const [error, setError] = useState<string | null>(null);
```

widget/embedded/src/hooks/useSearchCustomTokens/useSearchCustomTokens.ts
```diff
@@ -0,0 +1,74 @@
...
+    try {
+      let { tokens } = await httpService().searchCustomTokens(
+        { query, blockchain },
+        { signal: abortController.current.signal }
+      );
+
+      const customTokensSet = new Set(
+        customTokens().map((token) => createAssetKey(token))
+      );
+      tokens = tokens.filter(
```
nikaaru, line 35:
```suggestion
     const filteredTokens = tokens.filter(
```

widget/embedded/src/hooks/useSearchCustomTokens/useSearchCustomTokens.ts
```diff
@@ -0,0 +1,74 @@
...
+
+import { httpService } from '../../services/httpService';
+import { useAppStore } from '../../store/AppStore';
+import { createAssetKey } from '../../store/utils/wallets';
+import { debounce } from '../../utils/common';
+
+import { DEBOUNCE_DELAY } from './useSearchCustomTokens.constants';
+
+export function useSearchCustomTokens() {
+  const abortController = useRef<AbortController | null>(null);
```
nikaaru, line 13: NIT:
```suggestion
  const abortControllerRef = useRef<AbortController | null>(null);
```

widget/embedded/src/hooks/useSearchCustomTokens/useSearchCustomTokens.ts
```diff
@@ -0,0 +1,74 @@
...
+import { useCallback, useRef, useState } from 'react';
+
+import { httpService } from '../../services/httpService';
+import { useAppStore } from '../../store/AppStore';
+import { createAssetKey } from '../../store/utils/wallets';
+import { debounce } from '../../utils/common';
+
+import { DEBOUNCE_DELAY } from './useSearchCustomTokens.constants';
+
+export function useSearchCustomTokens() {
```
nikaaru, line 12: Don't you think it would be better to add interfaces for result of  `useSearchCustomTokens`?
```suggestion

interface SearchCustomTokenHookResult {
  fetch: (query: string, blockchain?: string) => void;
  cancel: () => void;
  loading: boolean;
  tokens: Token[];
  error: string | null;
}
export function useSearchCustomTokens(): SearchCustomTokenHookResult {
```

Ikari-Shinji-re:
> image
> 
> The token image is currently distorted due to setting both width and height to 100%. This forces the image to stretch to fill its container without maintaining its aspect ratio.

The Image component is responsible for displaying images at the desired size while maintaining their aspect ratio, ensuring they fit within the container without stretching. Modifying its behavior could potentially break some images.

widget/embedded/src/components/ImportCustomToken/ImportCustomToken.tsx
```diff
@@ -0,0 +1,121 @@
...
+          </Button>
+        </MessageBox>
+      </WatermarkedModal>
+      {blockchain && token && (
+        <CustomTokenModal
+          blockchain={blockchain}
+          token={token}
+          onSubmitClick={handleSubmit}
+          onClose={() => setShowImportModal(false)}
+          open={showImportModal}
```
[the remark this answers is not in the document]
Ikari-Shinji-re: We need a state to manage both opening and closing the modal. If we follow your suggestion for opening it, we won’t have a proper state and handler for closing it. Instead, we’d have to rely on resetting the custom token state, which causes flickering and a poor user experience by resetting the modal state before its animation completes.

nikaaru:
image

Please also include the default image in the confirmed custom token modal.

nikaaru (approved):
Thanks for the changes! Approved.

widget/ui/src/components/ChainToken/ChainToken.tsx
```diff
@@ -5,22 +5,23 @@ import React from 'react';
...
-    chainImage,
-    chianImageId,
-    size,
-    useAsPlaceholder,
-    loading,
-  } = props;
+  const { chianImageId, size, useAsPlaceholder, loading } = props;
+
+  const tokenImage =
+    props.tokenImage === '' ? DEFAULT_TOKEN_IMAGE_SRC : props.tokenImage;
```
yeager-eren, line 22: either use destructure or directly access using props. you can consider tokenImage and chainImage be destructured like what you did at line19

widget/ui/src/containers/Notifications/Notifications.types.ts
```diff
@@ -9,8 +9,8 @@ type Notification = {
 
 export type PropTypes = {
   list: Notification[];
-  getBlockchainImage: (blockchain: string) => string;
-  getTokenImage: (token: Asset) => string;
+  getBlockchainImage: (blockchain: string) => string | undefined;
```
yeager-eren, line 12: Other places you've made the prop optional using ?, I would suggest do the samething here as well.

PR #1023, opened 2025-02-08
fix: update balance after transaction with max amount

RyukTheCoder:
# Summary

After a transaction gets executed, balance for that wallet address will be fetched for source and destination wallet address and blockchain, then new balances will be set in store and already existing balances will be updated, but old balances are not getting removed. This results in that if a transaction get executed with max amount for an asset, there will be no balance related to that asset in the response of fetch balance so old balance for that asset will not get removed from store.

As a result of this change, the old balances will be removed initially, followed by the application of updates to address the described issue.

Fixes # 2231

Create a transaction with max amount of source asset. After executing the transaction successfully, you should observe that balances get updated properly.

# Checklist:

### DIFF
--- a/widget/embedded/src/store/slices/wallets.ts
+++ b/widget/embedded/src/store/slices/wallets.ts
@@ -442,6 +442,11 @@ export const createWalletsSlice: StateCreator<
           return;
         }
 
+        // Remove old balances for current wallet and blockchain
+        get().removeBalancesForWallet(walletType, {
+          chains: [wallet.blockChain],
+        });
+
         const balancesForWallet = createBalanceStateForNewAccount(wallet, get);
 
         nextAggregatedBalances = updateAggregatedBalanceStateForNewAccount(


widget/embedded/src/store/slices/wallets.ts
```diff
@@ -434,6 +436,16 @@ export const createWalletsSlice: StateCreator<
...
+          !addressesToFetch.find(
+            (address) =>
+              extractBalanceKey(balanceKey).walletAddress === address.address &&
+              extractBalanceKey(balanceKey).blockchain === address.blockchain
+          )
+      );
+
       let nextBalances: BalanceState = {};
       let nextAggregatedBalances: AggregatedBalanceState =
         get()._aggregatedBalances;
```
yeager-eren, line 451: How about adding this to the forEach:

```js
        get().removeBalancesForWallet(walletType, {
          chains: [wallet.blockChain],
        });

```

RyukTheCoder: Tnx, applied your suggestion.

yeager-eren (approved):
lgtm

PR #1024, opened 2025-02-08
fix: throw error on connect only if all namespaces failed

RyukTheCoder:
# Summary

Attempting to connect two namespaces which one of them is not available in Phantom account results in a success status but the success status modal will not close automatically because an error is thrown when any namespace fails to connect.
In this change, an error will be thrown only if all the namespaces fail to connect.

Fixes # 2242

This change can be tested by selecting an account containing only one of the supported namespaces and try to connect to both of them and observing that success modal will close automatically.

# Checklist:


yeager-eren:
No need to fix. namespace modal will be redesigned soon.

PR #1026, opened 2025-02-11
feat: add base chain to phantom evm supported chains

mikasackermn:
# Summary

Phantom Wallet has added support for the Base chain, which means we can now include Base in the list of supported chains.


RyukTheCoder (approved):
LGTM

PR #1029, opened 2025-02-15
fix: remove connected wallet on namespace disconnect

RyukTheCoder:
# Summary

There was a bug in disconnecting a namespace for a wallet which that caused incorrect state for wallet. Consider you connect to a wallet on both namespaces and after that one namespace gets disconnected (for example n result of switching to an account which does not support one of namespaces). In this scenario, old balances for disconnected namespace gets cleared but it won't be removed from connected wallets. In this PR, after disconnecting a namespace, the connected wallets in wallet slice related to that namespace get cleared instead of just balances being cleared.

Fixes # (issue)

You can test this change by repeating this scenario:
1. Connect to an account on Phantom with both EVM and  Solana namespaces.
2. Select an EVM and a Solana token and navigate to confirm wallets page.
3. Switch your account on Phantom to an account supporting only one namespace.
4. You should see that wallet related to disconnected namespace gets disconnected properly.

# Checklist:

### DIFF
--- a/widget/embedded/src/store/slices/wallets.ts
+++ b/widget/embedded/src/store/slices/wallets.ts
@@ -108,10 +108,15 @@ export interface WalletsSlice {
     accounts: Wallet[],
     namespace?: Namespace
   ) => Promise<void>;
+  disconnectNamespaces: (walletType: string, namespaces: Namespace[]) => void;
   /**
    * Disconnect a wallet and clean up balances after that.
    */
   disconnectWallet: (walletType: string) => void;
+  _changeSelectedWalletIfNeededOnRemove: (
+    walletType: string,
+    options?: { namespaces?: Namespace[] }
+  ) => void;
   clearConnectedWallet: () => void;
   fetchBalances: (
     accounts: Wallet[],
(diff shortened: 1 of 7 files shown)


yeager-eren:
My last commit includes:
- A hook to separate handling our wallet management events for both legacy and hub which can change `embedded` store.
- some logics of disconnectWallet exctracted to be used in a new method called disconnectNamespace

widget/embedded/src/store/slices/wallets.ts
```diff
@@ -401,33 +434,34 @@ export const createWalletsSlice: StateCreator<
...
+
+      get()._changeSelectedWalletIfNeededOnRemove(walletType);
+    }
+  },
+  /*
+   * If we are disconnecting a wallet that has `selected` for some blockchains,
+   * For those blockchains we will fallback to first connected wallet
+   * which means selected wallet will change.
+   */
+  _changeSelectedWalletIfNeededOnRemove: (walletType) => {
```
RyukTheCoder, line 454: First, this should be called before filtering connected wallets. In these changes. this method is being called after filtering connected wallets so no connected wallet with target wallet type will be found and selected wallet will not get changed properly.
Second, this method is getting called on `disconnectWallet` and `disconnectNamespaces`. It works fine on `disconnectWallet` but when it is getting called in `disconnectNamespace` it should only change selected wallet for blockchains which are related to target namespaces. Imagine you've connected Phantom on Solana and EVM namespaces. After that, you disconnect only Solana namespace. In this scenario, selected wallet for Solana blockchain should change but it should not change for Ethereum blockchain.

RyukTheCoder:
There 2 issues in events we emit using eventEmitter.
1. When we connect a wallet on hub (like Phantom) the event related to connecting wallets (`WalletEventTypes.CONNECT`) is emitted per each namespace but, only one event will be emitted on connecting wallets in legacy.
2. The event related to disconnecting wallets (`WalletEventTypes.DISCONNECT`) is not emitted. The reason is that it is emitted in `disconnectWallet` action in `wallets` slice if there be any connected wallet related to walletType in the parameters but, considering that connected wallets are getting clear in `disconnectNamespaces`, it will never emit that event. Also due to these changes, calling `disconnectWallet` on PROVIDER_DISCONNECT actually has no point.

yeager-eren:
> There 2 issues in events we emit using eventEmitter.
> 
>     1. When we connect a wallet on hub (like Phantom) the event related to connecting wallets (`WalletEventTypes.CONNECT`) is emitted per each namespace but, only one event will be emitted on connecting wallets in legacy.
> 
>     2. The event related to disconnecting wallets (`WalletEventTypes.DISCONNECT`) is not emitted. The reason is that it is emitted in `disconnectWallet` action in `wallets` slice if there be any connected wallet related to walletType in the parameters but, considering that connected wallets are getting clear in `disconnectNamespaces`, it will never emit that event. Also due to these changes, calling `disconnectWallet` on PROVIDER_DISCONNECT actually has no point.

1. Not sure, see `wallets/react/src/hub/utils.ts:194`, on connecting namesapce we are emitting `ACCOUNTS`
2. it will be emit on disconnect as you can see here`widget/embedded/src/containers/Wallets/useUpdates.ts:98`, am I missing something here?

PR #1031, opened 2025-02-16
fix: update balance only if account is still connected

RyukTheCoder:
# Summary

Currently, if we connect a wallet, balances for the addresses related to the wallet will be fetched and after being resolved, balances in store will get updated. The problem is that if we disconnect the wallet while the request for fetching balances is in pending state, balances will get updated after request being resolved anyway.
In this PR, we are making sure  

- [ ] Test A
- [ ] Test B

# Checklist:


yeager-eren (approved):
lgtm

PR #1037, opened 2025-02-18
perf: improve getConnectedWalletsDetails query by memozing

yeager-eren:
# Summary

Memoize getConnectedWalletsDetails result to improve accessing.

This is a deprecated method used by our dApp, to not affecting the performance of our client, we need to keep accessing the connected wallets and its balances through getConnectedWalletsDetails performant.

I only wrote a middleware to keep track of time when a change applied to the slice, then uses that as the key for revalidating the cache of memoized result. 

Fixes RF-2268

build and link `embedded` to our dApp. then make a production build using `yarn build` (on app-v2) and try to serve artifacts with tools like `serve` or `miniserve`. Slowdown your CPU then compare with last dApp deployment for hub.

# Checklist:


RyukTheCoder (approved):
LGTM

PR #1038, opened 2025-02-19
feat: improve widget UX on small devices with dynamic height

Ikari-Shinji-re:
# Summary

Some mobile users are having trouble seeing the call-to-action buttons at the bottom of certain pages. To resolve this, we plan to make the widget height dynamic on devices with a screen width of 768px or less. The widget will have a minimum height (matching its height on the home page when no token is selected) and a maximum height of 700px. The height will adjust based on the user's viewport size and the widget's offset from the top.

You can test these changes by decreasing the browser window width and adjusting the viewport height. Additionally, you can add elements above the widget in the app to check the effect of the offset on the widget's height.

- [ ] Test A
- [ ] Test B

# Checklist:

### DIFF
--- a/widget/embedded/src/components/Layout/Layout.tsx
+++ b/widget/embedded/src/components/Layout/Layout.tsx
@@ -4,12 +4,13 @@ import type { PropsWithChildren } from 'react';
 
 import { useManager } from '@rango-dev/queue-manager-react';
 import { BottomLogo, Divider, Header } from '@rango-dev/ui';
-import React, { useEffect, useRef, useState } from 'react';
+import React, { useEffect, useLayoutEffect, useRef, useState } from 'react';
 
 import { WIDGET_UI_ID } from '../../constants';
 import { useIframe } from '../../hooks/useIframe';
 import { isAppLoadedIntoIframe } from '../../hooks/useIframe/useIframe.helpers';
 import { useNavigateBack } from '../../hooks/useNavigateBack';
+import useScreenDetect from '../../hooks/useScreenDetect';
 import { useTheme } from '../../hooks/useTheme';
 import { useAppStore } from '../../store/AppStore';
 import { tabManager, useUiStore } from '../../store/ui';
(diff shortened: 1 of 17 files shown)


widget/embedded/src/components/Layout/Layout.tsx
```diff
@@ -99,6 +104,37 @@ function Layout(props: PropsWithChildren<PropTypes>) {
     setOpenRefreshModal(fetchStatus === 'failed');
   }, [fetchStatus]);
 
+  useLayoutEffect(() => {
+    const handler = () => {
```
nikaaru, line 108: Please refactor the height calculation logic into a helper function for clarity and reusability. For example:

```
function calculateHeight(containerTop: number, viewportHeight: number, offset: number = 0) {
  const suggestedHeight = viewportHeight - containerTop - offset;
  return Math.min(Math.max(suggestedHeight, WIDGET_MIN_HEIGHT), WIDGET_MAX_HEIGHT);
}
```

Then, update your `useLayoutEffect` to use this helper.

NIT: Also, you can create a **`debounced`** version of the resize handler to improve performance. This ensures that during rapid window resizing, the handler only runs once every 200ms, reducing unnecessary re-calculations:

```
// Create a debounced version of the handler.
const debouncedHandler = debounce(handler, 200);
window.addEventListener('resize', debouncedHandler);
```

Finally, pass the offset as a prop for flexibility and ensure that all dependencies are correctly managed.

Thank you!

nikaaru (commented):
Thanks for this PR :ok_hand: 
I'll take a look and provide some feedback.

widget/embedded/src/components/Layout/Layout.tsx
Ikari-Shinji-re: After addressing the other comments, I don't think there's a need to create a helper method. Also, I believe debounce isn't necessary here—this scenario doesn't occur frequently in real use cases, and adding a debounce might make the interaction feel laggy to the user.

nikaaru (approved):
LGTM, thanks for the changes!

PR #1058, opened 2025-03-01
feat: adjust the number of blockchains in the token selector based on widget height changes

Ikari-Shinji-re:
# Summary

After deciding to make the widget height dynamic for mobile users based on their device viewport, we encountered an issue on the token selector page where the token list might appear too small, affecting the UX. To resolve this, I adjusted the number of blockchains in the blockchain section according to the page height changes.

Decrease your browser width below 640px, go to the token selector page, and adjust the browser height.

- [ ] Test A
- [ ] Test B

# Checklist:

### DIFF
--- a/widget/embedded/src/components/BlockchainsSection/BlockchainsSection.tsx
+++ b/widget/embedded/src/components/BlockchainsSection/BlockchainsSection.tsx
@@ -1,4 +1,3 @@
-/* eslint-disable @typescript-eslint/no-magic-numbers */
 import type { PropTypes } from './BlockchainsSection.types';
 
 import { i18n } from '@lingui/core';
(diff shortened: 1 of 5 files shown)


widget/embedded/src/components/BlockchainsSection/BlockchainsSection.tsx
```diff
@@ -41,16 +48,25 @@ export function BlockchainsSection(props: PropTypes) {
...
+          </Typography>
+        </>
+      )}
       <Divider size={12} />
       <Blockchains>
         {fetchStatus === 'loading' &&
-          Array.from(Array(NUMBER_OF_LOADING), (e) => (
-            <Skeleton key={e} variant="rounded" height={50} />
-          ))}
+          Array.from(
```
nikaaru, line 62: the `e` parameter in the Array.from map function is undefined since Array(n) doesn't have elements. Better to use:

```jsx
{fetchStatus === 'loading' &&
  Array.from(
    { length: showCompactTokenSelector ? NUMBER_OF_LOADING_COMPACT_MODE : NUMBER_OF_LOADING },
    (_, index) => <Skeleton key={index} variant="rounded" height={50} />
  )}
```
  
This uses index for the key and the standard { length: n } pattern which is clearer.

Ikari-Shinji-re: I just moved this code. It was already like this before. But I’ll fix it.

nikaaru (approved):
LGTM

PR #1063, opened 2025-03-05
fix: handle namespace before action failure

RyukTheCoder:
# Summary

Previously, failures in "before" action hooks were not handled properly, preventing "or" action hooks from executing. This caused the UTXO namespace to get stuck in a "connecting" state when connecting Phantom, especially if BTC was turned off in Phantom.

The issue stemmed from changeAccountSubscriber, a "before" hook for the "connect" action, which checked for instance. If the check failed, it threw an error, blocking the execution of "or" and "after" hooks.

This PR introduces the following fixes:

Wraps "before" hooks in a try/catch to ensure errors don't prevent "or" hooks from running.
Adds intoConnectionFinished to the "or" hooks for the "connect" action.
These changes ensure that the connection process proceeds correctly, even if a "before" hook encounters an error.

This change can be tested by turning "BTC" off in Phantom and trying to connect to "UTXO" namespace.

# Checklist:

### DIFF
--- a/wallets/core/src/hub/namespaces/namespace.ts
+++ b/wallets/core/src/hub/namespaces/namespace.ts
@@ -19,9 +19,13 @@ import { generateStoreId, isAsync } from '../helpers.js';
 
 import {
   ACTION_NOT_FOUND_ERROR,
+  BEFORE_ACTION_FAILED_ERROR,
   NO_STORE_FOUND_ERROR,
   OR_ELSE_ACTION_FAILED_ERROR,
 } from './errors.js';
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+type Params = any[];
 
 /**
  *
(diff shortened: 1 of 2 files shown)


wallets/core/src/hub/namespaces/namespace.ts
```diff
@@ -316,8 +318,13 @@ class Namespace<T extends Actions<T>> {
     return result;
   }
 
-  #tryRunAction<K extends keyof T>(actionName: K, params: any[]): unknown {
-    this.#tryRunBeforeHooks(actionName);
+  #tryRunAction<K extends keyof T>(actionName: K, params: Params): unknown {
```
yeager-eren, line 321: First we've talked with each other I confirmed that we need to run `or` operators on `before` failure, but now I think `or` should only handle `action` errors only, and hooks are out of scope for the `or` operators. so the correct change here will be something like this:

```javascript
try{
this.#tryRunBeforeHooks(actionName);
} finally{
      this.#tryRunAfterHooks(actionName);
j}
```

Since we are promising `after` will be run at the end of operation, so when it fails, we need to call that anyways.

RyukTheCoder: So how should we handle `before` errors?

yeager-eren (approved):
lgtm

PR #1075, opened 2025-03-15
fix: batching state updates for fetchBalances

yeager-eren:
# Summary

Reduce state updates (calling `set`) by batching them into one update.

In `fetchBalances` we were updating `balances` and `aggregatedBalances` twice. each of them update the UI synchronously which means it was a render blocking operation. Also calling `fetchBalances` multiple times (during renders and connecting wallets) causes CPU be blocked for a noticeable time. 

In this PR, I made the remove operation into a util, then on fetching balances to batch the update, used that util.
Another change is in our middleware I changed it in someway to update the upcoming change and `lastUpdatedTime` in one update.

Fixes # (issue)

The bug reported on Parcel, so you need to link this PR and then try to test. Please consider to test `fetchingBalances` scenarios which means on connecting/disconnecting/switching wallets.

# Checklist:

### DIFF
--- a/widget/embedded/src/store/slices/wallets.ts
+++ b/widget/embedded/src/store/slices/wallets.ts
@@ -18,13 +18,12 @@ import { memoizedResult } from '../../utils/common';
 import { isAccountAndWalletMatched } from '../../utils/wallets';
 import { keepLastUpdated } from '../middlewares/keepLastUpdated';
 import {
+  computeNextBalancesWithNewPrices,
+  computeNextStateAfterWalletBalanceRemoval,
   createAssetKey,
-  createBalanceKey,
   createBalanceStateForNewAccount,
   extractAssetFromBalanceKey,
-  removeBalanceFromAggregatedBalance,
   updateAggregatedBalanceStateForNewAccount,
-  updateBalancesWithNewPrices,
 } from '../utils/wallets';
 
 type WalletAddress = string;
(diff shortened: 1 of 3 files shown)


widget/embedded/src/store/slices/wallets.ts
```diff
@@ -674,9 +620,21 @@ export const createWalletsSlice = keepLastUpdated<AppStoreState, WalletsSlice>(
...
+          const { _balances, _aggregatedBalances } =
+            nextStateAfterRemoveBalancesForWallet(
+              partialCurrentState,
+              walletType,
+              {
+                chains: [wallet.blockChain],
+              }
+            );
+          nextAggregatedBalances = _aggregatedBalances;
+          nextBalances = _balances;
```
RyukTheCoder, line 637: We also have a `updateBalancesWithNewPrices` which updates balances in some way but, it does in a different way by updating the `nextBalances` object itself. Here, `nextStateAfterRemoveBalancesForWallet` receives an object and returns a new object and we update the old object with the reference of the new created object. Maybe its better to have a consistent behavior and pick one of these two approaches. Also I think the method name can be improved by something like `getNextStateAfterRemovingBalancesForWallet`.

RyukTheCoder (changes requested):
Tnx, I just added a comment. Also I tested fetching balance on switch account it did not work properly. Check it please.

widget/embedded/src/store/slices/wallets.ts
yeager-eren: I want to show the purpose which is calculating the next state. so `get` doesn't make sense in this case. I changed it `computeNextStateAfterWalletBalanceRemoval`.

also changed `updateBalancesWithNewPrices`.

RyukTheCoder: Thanks, maybe we can improve `updateBalancesWithNewPrices` name to something like `computeNextBalancesWithNewPrices`.

RyukTheCoder (approved):
LGTM

PR #1077, opened 2025-03-15
fix: fetch custom tokens balances after wallets balances

Ikari-Shinji-re:
# Summary

Currently, there's a bug where if a user imports a custom token and then connects a wallet that supports the blockchain of that token, the custom token’s balance might not be displayed. This issue occurs when the request to fetch wallet balances resolves after the request for custom token balances, causing the custom token balance to be removed from the store once the wallet details request completes.

Fixes # (issue)

- [ ] Test A
- [ ] Test B

# Checklist:

### DIFF
--- a/widget/embedded/src/hooks/useSubscribeToWidgetEvents/useSubscribeToWidgetEvents.ts
+++ b/widget/embedded/src/hooks/useSubscribeToWidgetEvents/useSubscribeToWidgetEvents.ts
@@ -1,4 +1,5 @@
-import type { RouteEventData, StepEventData } from '../..';
+import type { ConnectedWallet, RouteEventData, StepEventData } from '../..';
+import type { Asset } from 'rango-types';
 
 import {
   isApprovalTX,
@@ -15,11 +16,21 @@ import { useNotificationStore } from '../../store/notification';
 
 export function useSubscribeToWidgetEvents() {
   const setNotification = useNotificationStore.use.setNotification();
-  const { connectedWallets, fetchBalances } = useAppStore();
+  const {
+    connectedWallets,
+    fetchBalances,
+    customTokens: getCustomTokens,
+  } = useAppStore();
 
   useEffect(() => {
     const handleStepEvent = (widgetEvent: StepEventData) => {
       const { event, step, route } = widgetEvent;
+
+      /**
+       * Determine if we should refetch balances:
+       * - When a transaction is sent (TX_SENT) and it's not an approval transaction.
+       * - When a step succeeds.
+       */
       const shouldRefetchBalance =
         (event.type === StepEventType.TX_EXECUTION &&
           event.status === StepExecutionEventStatus.TX_SENT &&
@@ -27,6 +38,12 @@ export function useSubscribeToWidgetEvents() {
         event.type === StepEventType.SUCCEEDED;
 
       if (shouldRefetchBalance) {
+        const fetchBalanceAccounts: ConnectedWallet[] = [];
+
+        /**
+         * Identify the wallet corresponding to the `from` blockchain and match it with a connected wallet.
+         * If found, add it to the balance fetch list.
+         */
         const fromWallet = route.wallets[step?.fromBlockchain];
         if (fromWallet) {
           const fromAccount = connectedWallets.find(
@@ -37,10 +54,14 @@ export function useSubscribeToWidgetEvents() {
               connectedWallet.chain === step?.fromBlockchain
           );
           if (fromAccount) {
-            void fetchBalances([fromAccount]);
+            fetchBalanceAccounts.push(fromAccount);
           }
         }
 
+        /**
+         * If the transaction is cross-chain (`fromBlockchain !== toBlockchain`),
+         * find and add the `to` blockchain wallet to the balance fetch list.
+         */
         if (step?.fromBlockchain !== step?.toBlockchain) {
           const toWallet = route.wallets[step?.toBlockchain];
           if (toWallet) {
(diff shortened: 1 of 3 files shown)


widget/embedded/src/store/slices/wallets.ts
```diff
@@ -724,6 +720,11 @@ export const createWalletsSlice = keepLastUpdated<AppStoreState, WalletsSlice>(
           `We couldn't fetch your account balances. Seem there is no information on blockchain for them yet.`
         );
       }
+
+      void get().fetchCustomTokensBalance({
```
RyukTheCoder, line 724: Two problems come to my mind about this approach:
1. Considering that `fetchBalances` is called inside of its own body to handle retry for failedWallets. This means that `fetchCustomTokensBalance` will be called twice for the same tokens and wallets.
2. If  `fetchBalances` fails, we will not try to call `fetchCustomTokensBalance`

RyukTheCoder (changes requested):

RyukTheCoder (approved):
LGTM

widget/embedded/src/store/slices/wallets.ts
```diff
@@ -626,6 +626,19 @@ export const createWalletsSlice = keepLastUpdated<AppStoreState, WalletsSlice>(
...
     clearConnectedWallet: () => set({ connectedWallets: [] }),
     fetchBalances: async (accounts, options) => {
+      await get().fetchMainTokensBalances(accounts);
+      /*
+       * - If `options.customTokens` is not provided, fetch all custom tokens.
+       * - If `options.customTokens` is provided and not empty, fetch only those tokens.
+       */
+      if (!options?.customTokens || options.customTokens.length > 0) {
+        void get().fetchCustomTokensBalances({
+          tokens: options?.customTokens ?? get().customTokens(),
```
yeager-eren, line 636: you can remove outside if block and do:
```js
tokens: options?.customTokens?.length ? options.customToken :  get().customTokens()
```

widget/embedded/src/store/slices/wallets.ts
```diff
@@ -122,10 +122,14 @@ export interface WalletsSlice {
   ) => void;
   clearConnectedWallet: () => void;
   fetchBalances: (
+    accounts: Wallet[],
+    options?: { customTokens: Asset[] }
```
yeager-eren, line 126:
```suggestion
    options?: { customTokens?: Asset[] }
```

widget/embedded/src/hooks/useSubscribeToWidgetEvents/useSubscribeToWidgetEvents.ts
```diff
@@ -15,18 +16,48 @@ import { useNotificationStore } from '../../store/notification';
...
+       * - When a step succeeds.
+       */
       const shouldRefetchBalance =
         (event.type === StepEventType.TX_EXECUTION &&
           event.status === StepExecutionEventStatus.TX_SENT &&
           !isApprovalTX(step)) ||
         event.type === StepEventType.SUCCEEDED;
 
       if (shouldRefetchBalance) {
+        // Define the from and to tokens involved in the step
```
yeager-eren, line 41: `fromToken`, `toToken` and `customTokens` not related to be exist here. you can move it to top of: getCustomTokens().forEach((token) => {

widget/embedded/src/store/slices/wallets.ts
```diff
@@ -657,7 +670,7 @@ export const createWalletsSlice = keepLastUpdated<AppStoreState, WalletsSlice>(
               address: wallet.address,
             }));
           if (failedWallets.length > 0) {
-            void get().fetchBalances(failedWallets, {
+            await get().fetchMainTokensBalances(failedWallets, {
```
yeager-eren, line 673: Why do you need to await here? previously if there is any failed wallets, it detached from the current function execution and will retry the failed wallets in parallel. but now it block the first execution (which may has some failed wallets) and wait for the retry result and block the updates for first execution.
the second request (for failed wallets) may be failed, in this case the updates for the first request won't be applied to store.

widget/embedded/src/store/slices/wallets.ts
Ikari-Shinji-re: In some situations, a request for custom token balances should not be made (e.g., if neither the "from" nor "to" token is a custom token). This check ensures that requests are only made when necessary.

widget/embedded/src/store/slices/wallets.ts
Ikari-Shinji-re: I use await in this line to ensure that when we fetch the custom token balances, all the API requests for wallet details are resolved first, preventing them from affecting the custom token balances. I will review this again to ensure no bugs occur with this change.

widget/embedded/src/store/slices/wallets.ts
Ikari-Shinji-re: I added an extra line to the comments to make it clearer.

widget/embedded/src/store/slices/wallets.ts
Ikari-Shinji-re: I moved the retry block to the end of the `fetchMainTokensBalance` function to prevent the issue you mentioned.

widget/embedded/src/store/slices/wallets.ts
yeager-eren: You mean `[]` shouldn't request the custom token balances? If yes, can you claridy between undefined and empty array?

yeager-eren: The code is just for demonstrate purposes. Just wanted to mention if we don't want to balances for custom tokens shouldn't be fetched, we can do it explicitly, instead of assumption around data types.

widget/embedded/src/store/slices/wallets.ts
yeager-eren: It's more a workaround to me, but if it needs significant change, please add a comment and describe the issue and why it should be kept at the end of the method (to avoid breaking the app in future)

widget/embedded/src/store/slices/wallets.ts
RyukTheCoder: Applied with minor changes in https://github.com/rango-exchange/rango-client/pull/1077/commits/0aab9d3175fb232439f207dd4c1dc8e69f2a91b3

widget/embedded/src/store/slices/wallets.ts
RyukTheCoder: Comment added https://github.com/rango-exchange/rango-client/pull/1077/commits/0aab9d3175fb232439f207dd4c1dc8e69f2a91b3

yeager-eren (approved):
lgtm

PR #1088, opened 2025-04-08
fix: wrap the error to make sure it is rethrown in the switch network

arlert-armin:
### Fix: prevent infinite loop on network switch cancellation in Trust Wallet

There was an issue with Trust Wallet where, when a user canceled the network switch request, 
the error was observed in the `or` branch of the `connect` action—specifically in `cleanAccountSubscriber`. 

However, the error wasn't rethrown because `cleanAccountSubscriber` only rethrows errors 
that are instances of `Error`. Since the cancellation error wasn't always a proper `Error` object, 
it got swallowed silently.

As a result, the task remained unblocked in the queue manager, but since no account was returned 
from `connect`, the queue manager kept retrying to switch networks, causing an infinite loop.

This PR wraps the caught error in a proper `Error` object to ensure it gets rethrown 
and handled correctly, preventing the loop.

- [ ] Test A
- [ ] Test B

# Checklist:

### DIFF
--- a/wallets/core/src/namespaces/evm/utils.ts
+++ b/wallets/core/src/namespaces/evm/utils.ts
@@ -48,6 +48,10 @@ export async function switchOrAddNetwork(
       await suggestNetwork(instance, chain);
     }
 
+    if (error instanceof Error) {
+      throw error;
+    }
+
     /*
      * Wrap the error to ensure it's an instance of Error.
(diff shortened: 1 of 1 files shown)


RyukTheCoder:
This error refinement should be done at wallet

RyukTheCoder:
This error refinement should be done at provider or hub level.

PR #1089, opened 2025-04-08
fix: optional namespace removal to prevent error for wallets without autoconnect

arlert-armin:
### Fix: prevent error on disconnect for wallets without auto-connect support

Wallets that do not support auto-connect are not stored in localStorage 
as part of the last connected wallets list. 

There was an error during disconnect, specifically when trying to filter 
the last connected wallets by the current wallet. Since there was no 
entry in localStorage for these wallets, the object was `undefined`, 
leading to a runtime error.

This PR makes the access to the last connected wallets optional, 
preventing the error when the value is missing.

Fixes # (issue)

- [ ] Test A
- [ ] Test B

# Checklist:

### DIFF
--- a/wallets/react/src/hub/lastConnectedWallets.ts
+++ b/wallets/react/src/hub/lastConnectedWallets.ts
@@ -119,12 +119,12 @@ export class LastConnectedWalletsFromStorage {
     const storageState = persistor.getItem(this.#storageKey) || {};
 
     const currentProviderNamespaces = storageState[providerId];
-    const newProviderNamespaces = currentProviderNamespaces.filter(
+    const newProviderNamespaces = currentProviderNamespaces?.filter(
       (namespace) => !namespaceIds.includes(namespace.namespace)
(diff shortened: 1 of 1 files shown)


wallets/react/src/hub/lastConnectedWallets.ts
```diff
@@ -119,7 +119,7 @@ export class LastConnectedWalletsFromStorage {
     const storageState = persistor.getItem(this.#storageKey) || {};
 
     const currentProviderNamespaces = storageState[providerId];
-    const newProviderNamespaces = currentProviderNamespaces.filter(
+    const newProviderNamespaces = currentProviderNamespaces?.filter(
```
RyukTheCoder, line 122: Also maybe you should change the line 127 to this:
```
if (newProviderNamespaces?.length > 0) {
```

RyukTheCoder (approved):
LGTM

PR #1097, opened 2025-04-15
fix: fix long token names in widget home page

Ikari-Shinji-re:
# Summary

Currently, on the homepage, token symbols with too many characters (such as some custom tokens) are not displayed properly, and in some cases, the alert for custom tokens may not appear. In this pull request, we've updated some styles to truncate long token symbols and ensure they are displayed correctly.

You could test with these sample tokens:

- 0x8d546c8e41d6c7a25855d0d9a6e0ef5d3b66808b
- 0xdb8f3858d6750f5772b9c77770807d29ec0395cb

# Checklist:

### DIFF
--- a/widget/ui/src/containers/SwapInput/TokenSection.tsx
+++ b/widget/ui/src/containers/SwapInput/TokenSection.tsx
@@ -8,6 +8,7 @@ import {
   ChainToken,
   Divider,
   Skeleton,
+  Tooltip,
   Typography,
 } from '../../components/index.js';
 
@@ -15,6 +16,9 @@ import {
   chainNameStyles,
   Container,
   skeletonStyles,
+  SymbolTooltipContent,
+  SymbolTooltipStyles,
+  Title,
   TitleContainer,
   tokenChainStyles,
   TokenSectionContainer,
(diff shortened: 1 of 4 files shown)


widget/ui/src/containers/SwapInput/TokenSection.tsx
```diff
@@ -54,11 +61,28 @@ export function TokenSection(props: TokenSectionProps) {
           ) : (
             <>
               <TitleContainer>
-                <Typography variant="title" size="medium">
-                  {error || (!loading && !tokenSymbol)
-                    ? i18n.t('Select Token')
-                    : tokenSymbol}
-                </Typography>
+                {tokenSelected ? (
+                  <Tooltip
```
mikasackermn, line 65: NIT:
I think it's better to show the tooltip only for long tokens

Ikari-Shinji-re: Since the token section's width changes with the widget size, we can't determine tooltip visibility based on the symbol's character count alone. While we could observe the `TokenSymbol` container for overflow to decide when to show the tooltip, it might not be worth the added complexity.

mikasackermn: Got it, thanks for the explanation

mikasackermn (approved):
LGTM

PR #1101, opened 2025-04-16
Fix: Resetting Installed State on Playground Wallet Filter

arlert-armin:
### Problem

When filtering wallets in the **Playground** to show only **one Hub wallet**, and there are **more than two Hub wallets available**, the `installed` state of the filtered wallet gets reset to `false`.

This was caused by how `getHub` was implemented in `useHubRefs`:

```ts
function getHub(): Hub {
  const hubProviders = hub.current?.getAll();

  if (
    !hub.current ||
    !hubProviders ||
    !checkProviderListsEquality(Array.from(hubProviders.values()), providers)
  ) {
    return createHub();
  }
  return hub.current;
}
```
If the current hub's providers didn’t match the expected list (`providers`), `getHub` recreated the hub using `createHub()`. This added each provider with default values, where `installed: false`. Since the new hub wasn't re-initialized afterward, the installed state remained `false`.

This issue did not happen when there was only one Hub wallet available, because the equality check always passed, so the hub wasn't recreated.

**✅ First Attempted Fix**

To fix this initially, I:

1.  Extracted the `runOnInit` method.
2.  Created a `hubProviderIds` string using `useMemo`, and used it in the dependency array of a `useEffect`:

    ```typescript
    /*
     * Extract and stringify the sorted list of Hub provider IDs.
     * This is used as a dependency in useEffect to detect changes in Hub providers.
     */
    const hubProviderIds = useMemo(
      () =>
        Array.from(getHub().getAll().values())
          .map((provider) => provider.id)
          .sort()
          .toString(),
      [getHub]
    );

    // Reinitialize Hub providers whenever the list of Hub provider IDs changes.
    useEffect(() => {
      runOnInit();
    }, [hubProviderIds]);
    ```

This ensured that the hub would be properly re-initialized whenever Hub providers changed.

**⚠️ Second Issue: Stale Closure in Subscription**

A second bug appeared in the event subscription logic.

We were doing:

```typescript
const { getStore, getHub } = useHubRefs(params.providers);
```

**The problem:** getHub was created using the initial params.providers, and that list became outdated. As a result, every event caused the hub to be recreated again, resetting the installed state repeatedly.

✅ **Final Fix**

To resolve this properly:

* I returned the actual `hubRef` from `useHubRefs`.
* In the subscription callback, I used `hubRef.current` if available, instead of calling `getHub()`:

```typescript
getStore()
  .subscribe((event) => {
    if (dataRef.current.onUpdateState) {
      try {
        mapHubEventsToLegacy(
          hubRef.current || getHub(), // Prefer the current ref
          event,
          dataRef.current.onUpdateState,
          dataRef.current.allVersionedProviders,
          dataRef.current.allBlockChains
        );
      } catch (e) {
        console.error(e);
      }
    }
    rerender((currentRender) => currentRender + 1);
  })
  .flushEvents();
```

- [ ] Test A
- [ ] Test B

# Checklist:

### DIFF
--- a/wallets/react/src/hub/useHubAdapter.ts
+++ b/wallets/react/src/hub/useHubAdapter.ts
@@ -54,10 +54,18 @@ export function useHubAdapter(params: UseAdapterParams): ProviderContext {
     };
   }, [params]);
 
+  const providersHash = params.providers
+    .map((provider) => provider.info.name)
+    .sort()
+    .toString();
+
   // Initialize instances
   useEffect(() => {
+    const hub = getHub();
+    const store = getStore();
+
     const runOnInit = () => {
-      getHub().init();
+      hub.init();
 
       rerender((currentRender) => currentRender + 1);
     };
@@ -77,21 +85,12 @@ export function useHubAdapter(params: UseAdapterParams): ProviderContext {
       }
     };
 
-    // Try to run, maybe it's ready.
-    runOnInit();
-
-    /*
-     * Try again when the page has been completely loaded.
-     * Some of wallets, take some time to be fully injected and loaded.
-     */
-    document.addEventListener('readystatechange', initHubWhenPageIsReady);
-
-    getStore()
+    store
       .subscribe((event) => {
         if (dataRef.current.onUpdateState) {
           try {
             mapHubEventsToLegacy(
-              getHub(),
+              hub,
               event,
               dataRef.current.onUpdateState,
               dataRef.current.allVersionedProviders,
@@ -104,7 +103,16 @@ export function useHubAdapter(params: UseAdapterParams): ProviderContext {
         rerender((currentRender) => currentRender + 1);
       })
       .flushEvents();
-  }, []);
+
+    // Try to run, maybe it's ready.
+    runOnInit();
+
+    /*
+     * Try again when the page has been completely loaded.
+     * Some of wallets, take some time to be fully injected and loaded.
+     */
+    document.addEventListener('readystatechange', initHubWhenPageIsReady);
+  }, [providersHash]);
 
   useAutoConnect({
     autoConnect: params.autoConnect,


RyukTheCoder:
I think it is not a good idea to change the interface of `useHubRefs` and accessing the hub ref directly. We just need to make sure that on each change on the providers list, below steps take place sequentially:
create store => create hub => subscribe to store => init hub
So I think maybe we can go with something like the below code containing only small changes:

```
  const providersHash = params.providers
    .map((provider) => provider.info.name)
    .sort()
    .toString();

  useEffect(() => {
    const hub = getHub();
    const store = getStore();

    const runOnInit = () => {
      hub.init();

      rerender((currentRender) => currentRender + 1);
    };

    // Then will call init whenever page is ready.
    const initHubWhenPageIsReady = (event: Event) => {
      // Then will call init whenever page is ready.
      if (
        event.target &&
        (event.target as Document).readyState === 'complete'
      ) {
        runOnInit();
        document.removeEventListener(
          'readystatechange',
          initHubWhenPageIsReady
        );
      }
    };

    store
      .subscribe((event) => {
        if (dataRef.current.onUpdateState) {
          try {
            mapHubEventsToLegacy(
              hub,
              event,
              dataRef.current.onUpdateState,
              dataRef.current.allVersionedProviders,
              dataRef.current.allBlockChains
            );
          } catch (e) {
            console.error(e);
          }
        }
        rerender((currentRender) => currentRender + 1);
      })
      .flushEvents();

    // Try to run, maybe it's ready.
    runOnInit();

    /*
     * Try again when the page has been completely loaded.
     * Some of wallets, take some time to be fully injected and loaded.
     */
    document.addEventListener('readystatechange', initHubWhenPageIsReady);
  }, [providersHash]);
```

arlert-armin:
> I think it is not a good idea to change the interface of `useHubRefs` and accessing the hub ref directly. We just need to make sure that on each change on the providers list, below steps take place sequentially: create store => create hub => subscribe to store => init hub So I think maybe we can go with something like the below code containing only small changes:
> 
> ```
>   const providersHash = params.providers
>     .map((provider) => provider.info.name)
>     .sort()
>     .toString();
> 
>   useEffect(() => {
>     const hub = getHub();
>     const store = getStore();
> 
>     const runOnInit = () => {
>       hub.init();
> 
>       rerender((currentRender) => currentRender + 1);
>     };
> 
>     // Then will call init whenever page is ready.
>     const initHubWhenPageIsReady = (event: Event) => {
>       // Then will call init whenever page is ready.
>       if (
>         event.target &&
>         (event.target as Document).readyState === 'complete'
>       ) {
>         runOnInit();
>         document.removeEventListener(
>           'readystatechange',
>           initHubWhenPageIsReady
>         );
>       }
>     };
> 
>     store
>       .subscribe((event) => {
>         if (dataRef.current.onUpdateState) {
>           try {
>             mapHubEventsToLegacy(
>               hub,
>               event,
>               dataRef.current.onUpdateState,
>               dataRef.current.allVersionedProviders,
>               dataRef.current.allBlockChains
>             );
>           } catch (e) {
>             console.error(e);
>           }
>         }
>         rerender((currentRender) => currentRender + 1);
>       })
>       .flushEvents();
> 
>     // Try to run, maybe it's ready.
>     runOnInit();
> 
>     /*
>      * Try again when the page has been completely loaded.
>      * Some of wallets, take some time to be fully injected and loaded.
>      */
>     document.addEventListener('readystatechange', initHubWhenPageIsReady);
>   }, [providersHash]);
> ```

Thanks for your suggestion, I've applied and it worked.

RyukTheCoder (approved):
LGTM

PR #1107, opened 2025-04-21
fix: error rethrow in or cases

arlert-armin:
# Summary

Some wallet providers, such as Trust Wallet, do not throw errors using the standard `Error` object. This behavior can lead to issues when attempting to catch and rethrow these errors consistently.

This update refactors the `tryRunOrOperators` function to capture the responses returned by `or` actions. If a response is an instance of `Error`, it is rethrown directly. If not, it is wrapped in a new `Error` object. This approach ensures that all `or` actions consistently return proper `Error` instances.

To support this change, the only existing `or` action, `cleanAccountSubscribers`, has been updated accordingly.

Fixes # (issue)

- [ ] Test A
- [ ] Test B

# Checklist:

### DIFF
--- a/wallets/core/src/hub/namespaces/namespace.test.ts
+++ b/wallets/core/src/hub/namespaces/namespace.test.ts
@@ -5,10 +5,60 @@ import { createStore } from '../mod.js';
 import { OR_ELSE_ACTION_FAILED_ERROR } from './errors.js';
 import { Namespace } from './namespace.js';
 
+interface CustomMatchers<R = unknown> {
+  toThrowErrorWithCause: (expectedError: unknown, expectedCause: unknown) => R;
+}
+
+declare module 'vitest' {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any, @typescript-eslint/no-empty-object-type
+  interface Assertion<T = any> extends CustomMatchers<T> {}
+  // eslint-disable-next-line @typescript-eslint/no-empty-object-type
+  interface AsymmetricMatchersContaining extends CustomMatchers {}
+}
+
 interface TestNamespaceActions {
   connect: () => void;
   disconnect: () => void;
 }
+
+expect.extend({
+  toThrowErrorWithCause(actualFunction, expectedError, expectedCause) {
+    try {
+      actualFunction();
+    } catch (error) {
+      const isExactError = error === expectedError;
+      const hasExpectedCause =
+        error instanceof Error && error.cause === expectedCause;
+
+      if (isExactError) {
+        return {
+          pass: true,
+          message: () => `Expected function to throw the exact error.`,
+        };
+      }
+
+      if (hasExpectedCause) {
+        return {
+          pass: true,
+          message: () =>
+            `Expected function to throw an error with the exact cause.`,
+        };
+      }
+
+      return {
+        pass: false,
+        message: () =>
+          `Expected function to throw the exact error with the expected cause, but got a different error.`,
+      };
+    }
+
+    return {
+      pass: false,
+      message: () =>
+        `Expected function to throw an Error, but it did not throw.`,
+    };
+  },
+});
 
 describe('check initializing Namespace', () => {
   const NAMESPACE = 'evm';
@@ -260,7 +310,7 @@ describe('check actions with hooks and operators', () => {
       configs: {},
     });
 
-    ns.or_else('connect', (_ctx: any, err: any) => {
+    ns.or_else('connect', (_ctx: unknown, err: unknown) => {
       return err instanceof Error;
     });
 
@@ -301,14 +351,17 @@ describe('check actions with hooks and operators', () => {
     expect(orActionFirst).toBeCalledTimes(2);
     expect(orActionSecond).toBeCalledTimes(2);
   });
-  test('throw error if `or_else` itself failed to run.', () => {
+  test('throw a wrapped error if `or_else` handler fails to run.', () => {
     const orActionFirst = vi.fn((_ctx, _err) => 1);
+    const orActionSecondError = 'This is actually a bad situation';
     const orActionSecond = vi.fn((_ctx, _err) => {
-      throw new Error('This is actually a bad situation');
+      // eslint-disable-next-line @typescript-eslint/only-throw-error
+      throw orActionSecondError;
     });
+    const connectActionError = new Error('Oops!');
 
     const connectAction = vi.fn(() => {
-      throw new Error('Oops!');
+      throw connectActionError;
     });
 
     const actions = new Map();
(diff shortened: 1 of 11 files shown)


wallets/core/src/namespaces/evm/actions.ts
```diff
@@ -111,9 +111,7 @@ export function changeAccountSubscriber(
         evmInstance.removeListener('accountsChanged', eventCallback);
       }
 
-      if (err instanceof Error) {
-        throw err;
-      }
+      return err;
```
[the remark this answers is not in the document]
yeager-eren: feel free to rename and change the code however you want, my code is just for demonstration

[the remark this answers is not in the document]
yeager-eren: also you need to update all the hub providers to use this new or action.

wallets/core/src/hub/namespaces/namespace.ts
```diff
@@ -410,28 +410,34 @@ class Namespace<T extends Actions<T>> {
     return result;
   }
 
-  #tryRunOrOperators<K extends keyof T>(
-    actionName: K,
-    actionError: unknown
-  ): unknown {
+  #tryRunOrOperators<K extends keyof T>(actionName: K, actionError: unknown) {
```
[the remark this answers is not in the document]
arlert-armin: Reverted the changes, rethrown the error, fixed the test cases, and added a new test case.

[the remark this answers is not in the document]
arlert-armin: Also set the action error as cause of or error.

wallets/core/src/hub/namespaces/namespace.test.ts
```diff
@@ -304,9 +304,9 @@ describe('check actions with hooks and operators', () => {
   test('throw error if `or_else` itself failed to run.', () => {
     const orActionFirst = vi.fn((_ctx, _err) => 1);
     const orActionSecond = vi.fn((_ctx, _err) => {
-      throw new Error('This is actually a bad situation');
+      // eslint-disable-next-line @typescript-eslint/only-throw-error
+      throw 'This is actually a bad situation';
```
yeager-eren, line 308: Why you've changed this?

wallets/core/src/hub/namespaces/namespace.test.ts
```diff
@@ -330,4 +330,31 @@ describe('check actions with hooks and operators', () => {
     expect(connectAction).toBeCalledTimes(1);
     expect(orActionSecond).toBeCalledTimes(1);
   });
+  test('throw exact error if `or_else` itself failed to run and error is instanceof Error.', () => {
```
yeager-eren, line 333: please also check `cause`.

wallets/core/src/namespaces/common/or.ts
```diff
@@ -0,0 +1,32 @@
...
+/**
+ * Standardizes an unknown error into an Error object and throws it.
+ * If the input is already an Error, it's thrown directly.
+ * Otherwise, a new Error is created with the input's message or string representation.
+ *
+ * @param context - The context.
+ * @param e - The unknown error object.
+ * @throws {Error} - The standardized Error object.
+ */
+export function standardizeAndThrow(context: Context, e: unknown): never {
```
yeager-eren, line 12: `standardizeAndThrowError` would be better.

arlert-armin: Done

wallets/core/src/hub/namespaces/namespace.test.ts
arlert-armin: Done.

wallets/core/src/hub/namespaces/namespace.test.ts
arlert-armin: Previously, the code checked if a thrown error was specifically OR_ELSE_ACTION_FAILED_ERROR('connect'). However, this scenario will no longer occur with the current changes when the error is an instance of Error. To address this and pass the existing test, I've changed the thrown error to be a string. Additionally, I've added a new test case to ensure that when an instance of an error is thrown, the exact error is propagated.

yeager-eren: So here ensure you will update the `expect`s to check these assumptions.

wallets/core/src/hub/namespaces/namespace.ts
```diff
@@ -422,17 +422,21 @@ class Namespace<T extends Actions<T>> {
...
+          orActionError.cause = actionError;
+          throw orActionError;
         }
         const errorMessage = OR_ELSE_ACTION_FAILED_ERROR(
           `${actionName.toString()} for ${this.namespaceId} namespace.`
         );
-        throw new Error(errorMessage, {
-          cause: actionError,
-        });
+        let wrappedActionError = actionError as Error;
```
yeager-eren, line 433: It took me some time to wrap my head around to what's going on here, I would suggest a code flow like this one:
```
        const errorMessage = OR_ELSE_ACTION_FAILED_ERROR(
          `${actionName.toString()} for ${this.namespaceId} namespace.`
        );
        const defaultError = new Error(errorMessage);

        if(actionError instanceof Error){
          actionError.cause = orActionError;
          defaultError.cause = actionError;
          throw defaultError;
        }else{
          defaultError.cause = new Error(actionError, {cause: orActionError});
          throw defaultError;
        }
```

you can change it, don't mean should be exactly this.

yeager-eren: or ```js

        const errorMessage = OR_ELSE_ACTION_FAILED_ERROR(
          `${actionName.toString()} for ${this.namespaceId} namespace.`
        );
        const defaultError = new Error(errorMessage);

        if(actionError instanceof Error){
          actionError.cause = orActionError;
          defaultError.cause = actionError;
        }else{
          defaultError.cause = new Error(actionError, {cause: orActionError});
        }
        
        throw defaultError;
```

Comment (arlert-armin) [reply wallets/core/src/hub/namespaces/namespace.ts:433]: In my approach, we ensure that the thrown actionError is a standard Error object. If it is not, we wrap it in an Error object to standardize it. This allows us to assign the orActionError as the cause of the error. Finally, we create a default error with a default message and assign the actionError as its cause.
At the first level, we will have an error with a meaningful message, and at the second level — in the cause of the error — we will have access to the string representation of the actionError. Its cause, in turn, will give us access to the orActionError.

In your suggestion, if the actionError is not an instance of Error, we replace the default error with a new error that wraps the actionError inside an Error object. However, this approach would cause us to lose the default error message, and the resulting error might not be meaningful.
If you are fine with that, I will go ahead and make the change.

File: wallets/core/src/hub/namespaces/namespace.test.ts [hunk 7]
```diff
@@ -322,12 +325,107 @@ describe('check actions with hooks and operators', () => {
...
+    const ns = new Namespace<{
+      connect: () => void;
+    }>(NAMESPACE, PROVIDER_ID, {
+      actions: actions,
+    });
+
+    ns.or_else('connect', orActionFirst);
+    ns.or_else('connect', orActionSecond);
+    // eslint-disable-next-line @typescript-eslint/no-magic-numbers
+    expect.assertions(6);
```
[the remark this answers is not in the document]
arlert-armin: Since I am catching the error in a try-catch block, we need to make sure that an error is actually thrown. Checking the assertions guarantees that the assertions inside the catch block are executed, which confirms that the try block has thrown an error.
I haven’t found a better way to ensure that an error has occurred while still having access to the error object. I would be happy to hear if you know of an alternative.

wallets/core/src/hub/namespaces/namespace.ts
arlert-armin: I've rearranged the chains, and it's much simpler now.

wallets/core/src/hub/namespaces/namespace.test.ts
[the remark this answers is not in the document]
yeager-eren: `assertion()` could be useful for async operations mostly. here you need to something like this:  
```js
    expect(() => ns.run('connect')).toThrowError(
      expect.objectContaining({
        message: OR_ELSE_ACTION_FAILED_ERROR(
          `connect for ${ns.namespaceId} namespace.`
        ),
        cause: expect.objectContaining({
          message: orActionSecondError,
          cause: connectActionError,
        }),
      })
    );

```

or if you are going to have strict check, you can use [.extend](https://vitest.dev/api/expect.html#expect-extend) to write a matched with exactly what you need.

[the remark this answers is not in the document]
arlert-armin: Thanks for your suggestion. I've used both a custom matcher and your way to handle the errors. And also there was a test case that wrote on my previous assumptions, which I removed.

yeager-eren (approved):
lgtm. thanks for this.

PR #1108, opened 2025-04-27
fix: avoid hub recreation

arlert-armin:
# Summary

Previously, in the `getHub` method, we checked if:
- the hub does not exist, **or**
- there is no hub provider,  
to decide whether to create a new hub.

However, there was a **third condition**: if the requested providers from the config did not match the available providers on the hub, we would also recreate the hub.

After investigating, we found the reason for this: sometimes, in the playground, the filter was changed, and the requested provider was not available on the hub anymore. Recreating the hub prevented errors indicating that a requested provider didn’t exist.  
**However, this approach reset the state of providers.**

In our first attempt to fix this, we tried re-initiating the providers:  
[Previous PR #1101](https://github.com/rango-exchange/rango-client/pull/1101)

But with that approach, the **connected state** was getting reset.

---

## Changes in this PR

- **Removed** the third condition in `getHub`.  
  Now, the hub will only be created if it doesn't exist.
- In the `providers` function of the adapter:
  - **First**, add any missing providers to the hub.
  - **Then**, filter the available providers on the hub to match the requested ones.

---

## Why?

This ensures:
- We no longer unnecessarily reset the connected state.
- We dynamically manage missing providers without needing to recreate the hub.
- A more stable and predictable provider state.

- [ ] Test A
- [ ] Test B

# Checklist:

### DIFF
--- a/wallets/core/src/hub/hub.ts
+++ b/wallets/core/src/hub/hub.ts
@@ -80,6 +80,18 @@ export class Hub {
     this.#providers.set(id, provider);
     return this;
   }
+  remove(id: string) {
+    const providerToRemove = this.#providers.get(id);
+
+    if (!providerToRemove) {
+      throw new Error(`Provider not found: No provider exists with ID "${id}"`);
+    }
+
+    this.#options.store?.getState().providers.removeProvider(id);
+    this.#providers.delete(id);
+
+    return this;
+  }
 
   get(providerId: string): Provider | undefined {
     return this.#providers.get(providerId);
@@ -96,6 +108,7 @@ export class Hub {
     output.forEach((result) => {
       const namespaces: NamespaceState[] = [];
       result.namespaces.forEach((b) => {
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
         const [getNamespaceState] = b as ReturnType<Namespace<any>['state']>;
 
         namespaces.push(getNamespaceState());
--- a/wallets/core/src/hub/provider/provider.test.ts
+++ b/wallets/core/src/hub/provider/provider.test.ts
@@ -16,6 +16,7 @@ describe('check providers', () => {
     evm: NamespaceBuilder<EvmActions>;
     solana: NamespaceBuilder<SolanaActions>;
   };
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
   let namespacesMap: Map<any, any>;
   let store: Store;
 
@@ -39,6 +40,7 @@ describe('check providers', () => {
     return () => {
       // eslint-disable-next-line @typescript-eslint/ban-ts-comment
       // @ts-ignore-next-line
+      // eslint-disable-next-line @typescript-eslint/no-unused-expressions
       (store = undefined), (namespaces = undefined);
     };
   });
@@ -52,7 +54,6 @@ describe('check providers', () => {
 
     expect(allNamespaces.size).toBe(2);
   });
-
   test("throw an error if store hasn't set and try to access .state() and .info()", () => {
     const provider = new Provider('garbage', namespacesMap, {
       info: garbageWalletInfo,
--- a/wallets/core/src/hub/provider/provider.ts
+++ b/wallets/core/src/hub/provider/provider.ts
@@ -296,10 +296,12 @@ export class Provider {
 
     this.#namespaces.forEach((namespace) => {
       if (hookName === 'after') {
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
         namespace.after(actionName as any, cb, {
           context,
         });
       } else if (hookName === 'before') {
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
         namespace.before(actionName as any, cb, {
           context,
         });
--- a/wallets/react/src/hub/useHubRefs.ts
+++ b/wallets/react/src/hub/useHubRefs.ts
@@ -3,7 +3,7 @@ import type { Provider, Store } from '@rango-dev/wallets-core';
 import { createStore, Hub } from '@rango-dev/wallets-core';
 import { useRef } from 'react';
 
-import { checkProviderListsEquality } from './utils.js';
+import { synchronizeHubWithConfigProviders } from './utils.js';
 
 export function useHubRefs(providers: Provider[]) {
   const store = useRef<Store | null>(null);
@@ -38,14 +38,11 @@ export function useHubRefs(providers: Provider[]) {
   function getHub(): Hub {
     const hubProviders = hub.current?.getAll();
 
-    if (
-      !hub.current ||
-      !hubProviders ||
-      // If hub does not contain a provider, it should be added
-      !checkProviderListsEquality(Array.from(hubProviders.values()), providers)
-    ) {
+    if (!hub.current || !hubProviders) {
       return createHub();
     }
+
+    synchronizeHubWithConfigProviders(hub.current, providers);
     return hub.current;
   }
 
--- a/wallets/react/src/hub/utils.ts
+++ b/wallets/react/src/hub/utils.ts
@@ -319,18 +319,35 @@ export function transformHubResultToLegacyResult(
   };
 }
 
-export function checkProviderListsEquality(
-  providerList1: Provider[],
-  providerList2: Provider[]
+/**
+ * Synchronizes providers in the hub with the configuration providers.
+ * - Registers and initializes any configuration providers not yet in the hub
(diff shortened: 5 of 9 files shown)


wallets/react/src/hub/useHubAdapter.ts
```diff
@@ -335,16 +331,22 @@ export function useHubAdapter(params: UseAdapterParams): ProviderContext {
       };
     },
     providers() {
+      ensureProvidersRegistered(getHub, params.providers);
```
RyukTheCoder, line 334: I think it's better that adding providers to hub take place in `getHub` method of `useHubRefs`. This way other methods of adapter will have access to updated hub as well.

RyukTheCoder (changes requested):
Tnx, I just added a comment.

wallets/react/src/hub/useHubAdapter.ts
arlert-armin: Good point, Thanks, Applied.

RyukTheCoder (approved):
LGTM

wallets/react/src/hub/useHubRefs.ts
```diff
@@ -38,14 +38,11 @@ export function useHubRefs(providers: Provider[]) {
...
-      !hub.current ||
-      !hubProviders ||
-      // If hub does not contain a provider, it should be added
-      !checkProviderListsEquality(Array.from(hubProviders.values()), providers)
-    ) {
+    if (!hub.current || !hubProviders) {
       return createHub();
     }
+
+    ensureProvidersRegistered(hub.current, providers);
```
yeager-eren, line 45: the name is misleading, the name is like it's just a check but it's not. it should describe the action happens by calling it. The actions are adding missing providers and also running `init` on them. You can use `registerMissingProvidersAndInit` or something like this.

wallets/react/src/hub/utils.ts
```diff
@@ -331,18 +331,15 @@ export function transformHubResultToLegacyResult(
...
+export function ensureProvidersRegistered(hub: Hub, providers: Provider[]) {
+  const existingProviders = hub.getAll();
+
+  const missingProviders = providers.filter(
+    (provider) => !existingProviders.get(provider.id)
+  );
+
+  missingProviders.forEach((provider) => {
+    const newProvider = hub.add(provider.id, provider);
+    newProvider.init();
```
yeager-eren, line 343: I think the code is not what exactly you were going to do, `hub.add()` returns `this` which means it will be Hub itself, it seems you are considering it is returning Provider itself. the problem here by this code `init` will be called for all the providers, not just the new one that you're adding.

wallets/react/src/hub/utils.ts
```diff
@@ -331,18 +331,15 @@ export function transformHubResultToLegacyResult(
...
-  const providerIds2 = providerList2
-    .map((provider) => provider.id)
-    .sort()
-    .toString();
-
-  return providerIds1 === providerIds2;
+export function ensureProvidersRegistered(hub: Hub, providers: Provider[]) {
+  const existingProviders = hub.getAll();
+
+  const missingProviders = providers.filter(
```
yeager-eren, line 337: You only `add` the missing ones, why you don't remove the no-longer-exist ones? and then you can remove the filter you've added for `providers` in useHubAdapter

wallets/react/src/hub/useHubRefs.ts
yeager-eren: At the end of my review, I asked for another action to be added in this function, so if you added that, my suggestion for the name is not valid anymore, you will need to choose another name yourself to represent add,remove and init functionality.

wallets/react/src/hub/utils.ts
arlert-armin: Fixed, called init on the added one only.

wallets/react/src/hub/useHubRefs.ts
arlert-armin: Regarding your first comment, thanks for your suggestion; I used yours. About the second one, I will answer the comment below.

wallets/react/src/hub/utils.ts
yeager-eren: The correct behavior is when you are removing a provider, its state should be gone as well. If playground or any other apps have their own requirement, they should handle those cases by themselves, not adding these use cases into a library.

wallets/react/src/hub/utils.ts
yeager-eren: Why not just `provider.init()`?

wallets/react/src/hub/utils.ts
arlert-armin: I've made the necessary changes:

- Added `remove` to the hub, `destroy` to the provider and  `removeProvider` in `providers`  to properly handle cleanup.
- Updated the logic to remove non-existing providers.
- Finally, removed the filtering from the adapter's `providers` method, as it's no longer needed.

wallets/core/src/hub/hub.ts
```diff
@@ -80,6 +80,18 @@ export class Hub {
     this.#providers.set(id, provider);
     return this;
   }
+  remove(id: string) {
```
yeager-eren, line 83: Cover your new methods (`remove` and `destroy`) with test.

wallets/core/src/hub/provider/provider.ts
```diff
@@ -164,6 +164,13 @@ export class Provider {
     this.#setupStore();
     return this;
   }
+  destroy(): void {
+    const store = this.#store;
+    if (!store) {
+      return;
+    }
+    store.getState().providers.removeProvider(this.id);
```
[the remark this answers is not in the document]
yeager-eren: The point here is this is not the `provider`'s responsibility to run `init` and `destroy`, it's the responsibility of topmost layer (in our case `hub`), to remove and call this method.

wallets/core/src/hub/provider/provider.ts
```diff
@@ -296,11 +303,11 @@ export class Provider {
 
     this.#namespaces.forEach((namespace) => {
       if (hookName === 'after') {
-        namespace.after(actionName as any, cb, {
+        namespace.after(actionName as never, cb, {
```
yeager-eren, line 306: `never` is not correct. `never` means code never reaches here, the reason we asserted as any is action name is a union of string, and here are accepting all of them, but in function signature we specified as `string`, so typescript is yelling. any will pass this.

yeager-eren: Same applies to the other file you've changed.

wallets/core/src/hub/hub.ts
arlert-armin: Done.

wallets/core/src/hub/provider/provider.ts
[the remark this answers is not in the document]
arlert-armin: Done.

wallets/core/src/hub/provider/provider.ts
arlert-armin: Done.

wallets/core/src/hub/provider/provider.ts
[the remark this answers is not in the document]
yeager-eren: To be more clear, 
`store.getState().providers.removeProvider(this.id);`
Should be removed from here. it should move up to upper layer (hub).

wallets/core/src/hub/provider/provider.test.ts
```diff
@@ -52,7 +54,22 @@ describe('check providers', () => {
...
+  test('Provider gets destoryed correctly', () => {
+    const providerId = 'garbage';
+    const provider = new Provider(providerId, namespacesMap, {
+      info: garbageWalletInfo,
+    });
+    const store = createStore();
+    provider.store(store);
 
+    expect(store.getState().providers.list[providerId]).toBeDefined();
+    expect(Object.entries(store.getState().providers.list).length).toBe(1);
```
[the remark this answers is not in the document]
arlert-armin: Done.

wallets/core/src/hub/provider/provider.ts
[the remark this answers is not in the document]
arlert-armin: I've removed the entire `destroy` method from here since it didn't have any other functionality. 

The provider removal (`store.getState().providers.removeProvider(this.id)`) has been moved to the hub layer, where it will:
1. Remove the provider from the store
2. Remove it from the local map

I've also:
- Removed the test from provider tests
- Added corresponding tests to `hubTests`

PR #1114, opened 2025-04-30
fix: adjust the warning alert position for custom tokens on the homepage

Ikari-Shinji-re:
# Summary

A large gap appears between the token symbol and the warning alert when there are few characters; this spacing should be removed : 

Fixes # (issue)

Use max-width instead of width for the token symbol element.

Import custom tokens with symbols of varying lengths, including both short and long character counts:
29nZcMWKbq6kPr6icM9tLrYWQor2eACWr1ifrVU1pump
0xe95a392047c46359807e3fbb263492298bcd35f9

# Checklist:


arlert-armin (approved):
LGTM

PR #1132, opened 2025-05-11
Add `has` Method to Namespace Proxy and Fix Hub Wallet `canEagerConnect` Check

arlert-armin:
**Summary:**  
This PR introduces a `has` method to the namespace proxy, allowing us to check for the existence of a property. It also resolves the issue where hub wallets without `canEagerConnect` were incorrectly added to `last-connected-wallets`.

**Changes:**

- Added a `has` method to the namespace proxy:
  - Returns `true` if the requested property:
    - Is one of the allowed methods,
    - Has a value of type `number` or `string`, or
    - Corresponds to an existing action on the namespace.

- Added a `hasAction` method to the `Namespace` class to support this logic.

- Updated `hubAdapter` to use `'canEagerConnect' in namespace` instead of directly checking the value:
  - This ensures the `has` method on the proxy is invoked.
  - Prevents hub wallets without `canEagerConnect` from being added to `last-connected-wallets`.

**Why:**  
Previously, we lacked a clean way to check if a namespace had a specific action or property. This led to incorrect behavior in `hubAdapter`, especially with hub-hosted wallets that didn’t support `canEagerConnect`. This change fixes that logic and improves code readability and robustness.

**Fixes:**
- Bug: Namespace added to `last-connected-wallets` even when wallet on hub doesn't support `canEagerConnect`.

- [ ] Test A
- [ ] Test B

# Checklist:

### DIFF
--- a/wallets/core/src/builders/namespace.ts
+++ b/wallets/core/src/builders/namespace.ts
@@ -19,7 +19,17 @@ export const allowedMethods = [
   'or_else',
   'store',
 ] as const;
-const allowedPublicValues = ['string', 'number'] as const;
+/**
+ * List of value types that are considered safe to return directly when
+ * accessing properties on a proxied `Namespace`.
+ *
+ * This is useful for allowing public values like `version` to be accessed.
+ * If a property's value is of one of these types (e.g.,`'string'`, `'number'`),
+ * it will be returned as-is.
+ *
+ * @const {Array<'string' | 'number'>} allowedPublicValues
+ */
+const allowedPublicValues = ['string', 'number'];
 
 export class NamespaceBuilder<T extends Actions<T>> {
   #id: string;
@@ -194,9 +204,14 @@ export class NamespaceBuilder<T extends Actions<T>> {
 
     const api = new Proxy(namespace, {
       has: (_, property) => {
-        if (
-          allowedMethods.includes(property as (typeof allowedMethods)[number])
-        ) {
+        if (typeof property !== 'string') {
+          throw new Error(
+            'You can use string as your property on Namespace instance.'
+          );
+        }
+        // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+        // @ts-ignore-next-line
+        if (allowedMethods.includes(property)) {
           return true;
         }
         // eslint-disable-next-line @typescript-eslint/ban-ts-comment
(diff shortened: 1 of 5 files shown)


wallets/core/src/builders/namespace.ts
```diff
@@ -208,12 +224,11 @@ export class NamespaceBuilder<T extends Actions<T>> {
           return targetValue.bind(namespace);
         }
 
-        /*
-         * This is useful accessing values like `version`, If we don't do this, we should whitelist
-         * All the values as well, So it can be confusing for someone that only wants to add a public value to `Namespace`
-         */
-        const allowedPublicValues = ['string', 'number'];
-        if (allowedPublicValues.includes(typeof targetValue)) {
+        if (
```
yeager-eren, line 227: Why you removed the comment? it's useful.

yeager-eren: Also why you asserted the type? this is usually reduce the type safety.
target value can be any value, but here you are forcing ts to consider it as only string or number.

if you are trying to fix the ts error, you could remove `as const`

wallets/core/src/builders/namespace.ts
```diff
@@ -192,6 +193,21 @@ export class NamespaceBuilder<T extends Actions<T>> {
     this.#addHooksFromActionBuilders(namespace);
 
     const api = new Proxy(namespace, {
+      has: (_, property) => {
```
yeager-eren, line 196: To remove the inline assertions, you can use what we did in `get`:

```
        if (typeof property !== 'string') {
          throw new Error(
            'You can use string as your property on Namespace instance.'
          );
        }
        // eslint-disable-next-line @typescript-eslint/ban-ts-comment
        // @ts-ignore-next-line
        const targetValue = namespace[property];

```

wallets/core/src/builders/namespace.ts
```diff
@@ -19,6 +19,7 @@ export const allowedMethods = [
   'or_else',
   'store',
 ] as const;
+const allowedPublicValues = ['string', 'number'] as const;
```
yeager-eren, line 22: Don't forget to add jsdoc.

yeager-eren (changes requested):

arlert-armin (commented):
I've fixed things you've mentioned and added new tests.

yeager-eren (approved):
lgtm. thanks.

PR #1134, opened 2025-05-12
fix: display error message from response instead of default fallback

mikasackermn:
# Summary

Show the error message returned in the response instead of using the default one

Fixes # (issue)

1. Connect to the Phantom wallet.
2. Make a swap on the Sui blockchain.
3. At the wallet selection step, add a custom Sui destination and then click Confirm.

# Checklist:

### DIFF
--- a/widget/embedded/src/hooks/useConfirmSwap/useConfirmSwap.ts
+++ b/widget/embedded/src/hooks/useConfirmSwap/useConfirmSwap.ts
@@ -72,7 +72,7 @@ export function useConfirmSwap(): ConfirmSwap {
         const { result } = response;
 
         if (!result) {
-          throw new Error('Error fetching updated route');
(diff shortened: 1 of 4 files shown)


widget/embedded/src/hooks/useConfirmSwap/useConfirmSwap.ts
```diff
@@ -72,7 +72,7 @@ export function useConfirmSwap(): ConfirmSwap {
         const { result } = response;
 
         if (!result) {
-          throw new Error('Error fetching updated route');
+          throw new Error(response.error ?? 'Error fetching updated route');
```
Ikari-Shinji-re, line 75:
```suggestion
          throw new Error(response.error ?? 'Error fetching updated quote');
```

Ikari-Shinji-re (approved):

PR #1142, opened 2025-05-19
fix: run connect namespaces sequentially when connect of adapter called separately

RyukTheCoder:
# Summary

There was an issue when calling connect on multiple namespaces of a provider simultaneously. The root cause was that the useHubAdapter hook exposes a connect method that can be invoked concurrently for different wallet namespaces. Some wallets are unable to handle parallel connection requests, leading to failures during the connection process.

This PR introduces a new custom React hook that manages a controlled, sequential execution queue for asynchronous tasks. This hook ensures that namespace connections are handled one at a time, preventing race conditions or request collisions for wallets that do not support concurrent operations.

Tested by connecting all namespaces on Trust wallet and observing that all of that gets connected properly.

# Checklist:

### DIFF
--- a/wallets/react/src/hub/helpers.ts
+++ b/wallets/react/src/hub/helpers.ts
@@ -77,3 +77,59 @@ export function isConnectResultSolana(
 ): result is Accounts {
   return Array.isArray(result);
 }
+
+type QueueItem<T> = {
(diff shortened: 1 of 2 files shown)


arlert-armin (approved):
LGTM

PR #1165, opened 2025-06-02
fix: incorrect rate display for values smaller than 1

mikasackermn:
# Summary

The rate for values smaller than 1 was displayed with 14 decimal places, which meant it added extra zeros for numbers with fewer than 14 decimals.

Fixes # (issue)

This issue was resolved by removing the unnecessary trailing zeros.

# Checklist:


RyukTheCoder (approved):
LGTM

PR #1167, opened 2025-06-03
fix: resolve error on get wallet info in playground

RyukTheCoder:
# Summary

This PR addresses an issue encountered in retrieving hub wallet information within the functional settings on the playground. The problem stemmed from calling info method on hub wallets, triggering a 'You need to set your store using .store method first.' error due to the store not being configured for the hub. Directly accessing provider information led to this error. The error was caught by checking information on the legacy provider, but the `getWalletInfo` method was not implemented for the slush wallet in the legacy provider, causing the playground to crash. This PR resolves this issue by obtaining the necessary information through the getWalletInfo method provided by the `useWallets` hook.

Tested by checking wallets list in functional settings on playground.

# Checklist:

### DIFF
--- a/widget/playground/src/containers/FunctionalLayout/FunctionalLayout.Wallets.tsx
+++ b/widget/playground/src/containers/FunctionalLayout/FunctionalLayout.Wallets.tsx
@@ -14,7 +14,6 @@ import React from 'react';
 
 import { MultiSelect } from '../../components/MultiSelect';
 import { useConfigStore } from '../../store/config';
-import { useMetaStore } from '../../store/meta';
 
 import {
   connectButtonStyles,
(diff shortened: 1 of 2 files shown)


arlert-armin:
Great work on this PR! I wanted to add some context on why I found it didn't fully resolve the broader issue I've been tackling in the Wallets Playground.

This PR successfully fixed the immediate **"store not set" error** by correctly fetching wallet information via the `useWallets` hook. However, the problem it uncovered was that if I then filtered wallets, it inadvertently **prevented other necessary providers from being added to the adapter's complete list**.

The Wallets Playground's side list needs *all* available wallets to be displayed. If some were filtered out at the adapter level, the system would still throw an error because these providers weren't found when trying to render them. So, while this PR addressed a specific crash by changing the info source, it didn't account for the comprehensive display requirement of all wallets in the sidebar.

---

I've since implemented a **temporary fix in PR #1168**  to unblock our current release. That PR specifically addresses a crash involving Slush Wallet, where its legacy `getWalletInfo` wasn't implemented.

For a more robust, long-term solution, I'm proposing that we revisit the core wallet architecture. The idea is to allow a provider's info to be returned independently, even if its store isn't yet set. Since the store is designed to be optional, its info *should* ideally be accessible on its own. This could provide a more stable way to ensure all wallets are correctly represented and their info is accessible, without these ongoing filtering complications.

arlert-armin:
I'm closing this pr since I've opened #1169  on this issue.

PR #1169, opened 2025-06-08
fix: return provider info with optional store

arlert-armin:
# Summary

This PR addresses an issue where the search input on the History page retained its last value upon re-entry, even after navigating away without explicitly clearing it.

Fixes # (issue)
This PR resolves the issue by disabling autocomplete for the search input field on the History page. 

- [ ] Test A
- [ ] Test B

# Checklist:


widget/playground/src/containers/FunctionalLayout/FunctionalLayout.utils.ts
```diff
@@ -46,9 +46,9 @@ export function getWalletsList(
       const namespacesProperty = providerProperties?.find(
         (property) => property.name === 'namespaces'
       );
-      const supportedChainsNames = namespacesProperty?.value.data
-        .flatMap((obj) => obj.chains)
-        .map((chain) => chain.name);
+      const supportedChainsNames = namespacesProperty?.value.data.flatMap(
```
RyukTheCoder: With this change, `supportedChainsNames` is no longer list of supported chains by the provider and it is now list of ids of namespaces. Considering that it is only used to pass to `getCategoryNetworks`, I think it does not currently result in any problem but it is adding an assumption indicating that having the id of namespaces is enough for getting categories and this assumption is not communicated through the code.

RyukTheCoder (changes requested):
Just added a comment.

widget/playground/src/containers/FunctionalLayout/FunctionalLayout.utils.ts
arlert-armin: It was already working since it only required the blockchain type in [getCategoryNetworks](https://github.com/rango-exchange/rango-client/blob/83214525647db795ab0d04e2ef4a5a88b39087bc/widget/playground/src/containers/FunctionalLayout/FunctionalLayout.utils.ts#L57), However, I've updated it to get the supported chains from the provider, as it was originally.

RyukTheCoder (approved):
LGTM

PR #1181, opened 2025-06-10
fix: abort all namespace connections if user rejects one connect request

RyukTheCoder:
# Summary

Currently, if user select multiple namespace to connect to and rejects the connect request in wallet for one namespace, widget will try to connect to other remaining namespaces resulting in multiple connect requests on wallet which can be annoying for the user. In this PR, this behavior is changed by checking for the code of error related to failed connect request and if it is equal to 4001 (which is documented [here](https://eips.ethereum.org/EIPS/eip-1193#provider-errors) for Ethereum wallets and also is accepted by [Solana wallets](https://docs.phantom.com/solana/errors)) the connection for remaining namespaces will be aborted.

Fixes # 2399

Tested by trying to connect to all namespaces on Phantom and Trust wallet and rejecting the connect request in wallet and observing that remaining connect errors fail without triggering any connect request to the wallet.

# Checklist:

### DIFF
--- a/wallets/provider-trustwallet/src/namespaces/solana.ts
+++ b/wallets/provider-trustwallet/src/namespaces/solana.ts
@@ -8,7 +8,10 @@ import {
 import { actions, builders } from '@rango-dev/wallets-core/namespaces/solana';
 
 import { WALLET_ID } from '../constants.js';
-import { solanaTrustWallet } from '../utils.js';
+import {
+  solanaTrustWallet,
+  standardizeTrustWalletInAppBrowserError,
+} from '../utils.js';
 
 const [changeAccountSubscriber, changeAccountCleanup] =
   actions.changeAccountSubscriber(solanaTrustWallet);
@@ -18,6 +21,7 @@ const connect = builders
   .action(actions.connect(solanaTrustWallet))
   .before(changeAccountSubscriber)
   .or(changeAccountCleanup)
+  .or(standardizeTrustWalletInAppBrowserError)
   .or(standardizeAndThrowError)
   .build();
 
--- a/wallets/react/src/hub/helpers.ts
+++ b/wallets/react/src/hub/helpers.ts
@@ -90,9 +90,29 @@ type QueueItem<T> = {
  * predictable task execution order.
  *
  */
-export function createQueue() {
+export function createQueue(options?: {
+  onError?: (
+    error: unknown,
+    actions: {
+      removeCurrentKeyFromQueue: () => void;
+    }
+  ) => void;
+}) {
   const processingKeys = new Set<string>();
-  const queue: QueueItem<unknown>[] = [];
+  let queue: QueueItem<unknown>[] = [];
+
+  const removeKeyFromQueue = (
+    key: string,
+    result: Result<unknown, unknown>
+  ) => {
+    queue = queue.filter((q) => {
+      if (q.key !== key) {
+        return true;
+      }
+      q.resolve(result);
+      return false;
+    });
+  };
 
   const processQueue = async () => {
     const currentItem = queue.find((item) => !processingKeys.has(item.key));
@@ -107,6 +127,13 @@ export function createQueue() {
       const result = await task();
       resolve(new Ok(result));
     } catch (error) {
+      if (options?.onError) {
+        options.onError(error, {
+          removeCurrentKeyFromQueue: () =>
+            removeKeyFromQueue(key, new Err(error)),
+        });
+      }
+
       resolve(new Err(error));
     } finally {
       const indexOfCurrentItem = queue.findIndex((item) => item.key === key);
--- a/wallets/core/src/hub/mod.ts
+++ b/wallets/core/src/hub/mod.ts
@@ -1,4 +1,8 @@
-export type { Subscriber, SubscriberCleanUp } from './namespaces/mod.js';
+export type {
+  Subscriber,
+  SubscriberCleanUp,
+  Context,
+} from './namespaces/mod.js';
 export { Namespace } from './namespaces/mod.js';
 
 export { Provider } from './provider/mod.js';
--- a/wallets/core/src/mod.ts
+++ b/wallets/core/src/mod.ts
@@ -6,6 +6,7 @@ export type {
   CommonNamespaceKeys,
   Subscriber,
   SubscriberCleanUp,
+  Context,
 } from './hub/mod.js';
 export {
   Hub,
--- a/wallets/core/src/namespaces/evm/utils.ts
+++ b/wallets/core/src/namespaces/evm/utils.ts
@@ -50,3 +50,17 @@ export async function switchOrAddNetwork(
     throw switchError;
   }
 }
+
+const EIP_1193_USER_REJECTION_ERROR_CODE = 4001;
+export function isUserRejectionError(error: unknown): boolean {
+  // EIP-1193 user rejection error can be in error.code or error.cause.code
+  if (typeof error === 'object' && error !== null) {
+    const code = (error as { code?: number }).code;
+    const causeCode = (error as { cause?: { code?: number } }).cause?.code;
+    return (
+      code === EIP_1193_USER_REJECTION_ERROR_CODE ||
+      causeCode === EIP_1193_USER_REJECTION_ERROR_CODE
+    );
+  }
+  return false;
+}
--- a/wallets/provider-trustwallet/src/namespaces/evm.ts
+++ b/wallets/provider-trustwallet/src/namespaces/evm.ts
@@ -8,7 +8,10 @@ import {
 import { actions, builders } from '@rango-dev/wallets-core/namespaces/evm';
 
 import { WALLET_ID } from '../constants.js';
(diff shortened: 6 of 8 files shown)


wallets/core/src/errors/provider.ts
```diff
@@ -0,0 +1,54 @@
...
+    case 'UNKNOWN_ERROR':
+      return 'User rejected the request.';
+    case 'UNEXPECTED_ERROR':
+      return 'User rejected the request.';
+    case 'DISCONNECTED':
+      return 'The Provider is disconnected from all chains.';
+    case 'CHAIN_DISCONNECTED':
+      return 'The Provider is not connected to the requested chain.';
+
+    default:
```
yeager-eren: You are returning `string | void`, what you are expecting here?

wallets/react/src/hub/helpers.ts
```diff
@@ -107,6 +109,16 @@ export function createQueue() {
       const result = await task();
       resolve(new Ok(result));
     } catch (error) {
+      if (options?.shouldClearQueueForKeyOnError?.(error)) {
```
yeager-eren: Keep the flow simple and explicit by avoiding adding specific use cases to the main function.

you can have a `onError` here which has some hooks and don't care about how it's going to handle the situation (e.g. clearing queue in your case).

pseudo code:

```
onError: (error: uknown, actions: { removeCurrentItemFromQueue } ): void
```

and:

```
let queue: QueueItem<unknown>[] = [];
const actions = {
    removeCurrentItemFromQueue: () => {
       queue = queue.filter...
    }
}

// ...

catch(error){
  if(options.onError) options.onError(error);
  resolve(new Err(error));
}
```

and in wallets/react/src/hub/useHubAdapter.ts

```
  if (error && typeof error === 'object' &&  (error as ProviderError).type === 'USER_REJECTION') {
        actions.removeCurrentItemFromQueue()
      }
```

wallets/provider-trustwallet/src/namespaces/solana.ts
```diff
@@ -18,6 +19,7 @@ const connect = builders
   .action(actions.connect(solanaTrustWallet))
   .before(changeAccountSubscriber)
   .or(changeAccountCleanup)
+  .or(evmOR.standardizeAndThrowEIP1193Error) // Trust wallet errors are inspired by EIP1193
```
yeager-eren: Adding doc link is more helpful, the util name is communicating same thing as comment.

wallets/core/src/errors/provider.ts
```diff
@@ -0,0 +1,54 @@
...
+/**
+ * Custom error class to represent provider-related errors.
+ * Extends the native JavaScript Error class and adds a custom 'type' field
+ * to differentiate between various error types specific to the provider context.
+ *
+ * @param type - A specific type of provider error, represented by the ProviderErrorType enum.
+ * @param message - Optional custom error message; if omitted, a default message based on the error type is used.
+ * @param options - Optional ErrorOptions to control error behavior (e.g., cause).
+ */
+export class ProviderError extends Error {
```
yeager-eren: You're implementation is different than [spec](https://eips.ethereum.org/EIPS/eip-1193#errors). see these as [ref1](https://github.com/thirdweb-dev/web3-onboard/blob/57b00e5946da88de164a7692775830283a026a4c/packages/common/src/types.ts#L378),[ref2](https://github.com/thirdweb-dev/web3-onboard/blob/57b00e5946da88de164a7692775830283a026a4c/packages/common/src/errors.ts#L3)

would be great to support newer options like `cause`. [see this](https://github.com/wevm/viem/blob/65b25a9aaeaefb798cba2647d87c642a76405a52/src/errors/rpc.ts#L87).

wallets/core/src/errors/provider.ts
```diff
@@ -0,0 +1,54 @@
+// This code is inspired by EIP-1193 provider errors https://eips.ethereum.org/EIPS/eip-1193#provider-errors
+export type ProviderErrorType =
```
yeager-eren: Use enum instead defining erros twice:

```
enum ...Codes{
USER_REJECTION = 4001
}
```

yeager-eren: You can rely on `code` number only, why you are getting an error code from wallet, making it to a string then in `getDefaultErrorMessage` making a message based on that string. you can use the `code` number without needing this extra work.

wallets/core/src/errors/provider.ts
yeager-eren: make sure the name is exactly what you see in the spec as well.

wallets/core/src/errors/provider.ts
```diff
@@ -0,0 +1,54 @@
+// This code is inspired by EIP-1193 provider errors https://eips.ethereum.org/EIPS/eip-1193#provider-errors
```
yeager-eren: this error can be defined under evm itself, so you can move it to `namespaces/evm/erros.ts` instead of adding a new entry point.

wallets/core/src/errors/provider.ts
RyukTheCoder: it's `string | undefined`. Added the return type.

wallets/react/src/hub/helpers.ts
RyukTheCoder: Tnx, applied in https://github.com/rango-exchange/rango-client/pull/1181/commits/27a6a5a2341d8128cbd67a7b28fc8d7c848d27bd

wallets/provider-trustwallet/src/namespaces/solana.ts
RyukTheCoder: I don't get how they are communicating the same thing. The comment is explaining why standardizeAndThrowEIP1193Error util is used to handle Solana errors. Didn't found any docs indicating that.

wallets/core/src/errors/provider.ts
RyukTheCoder: I think you misunderstood the purpose of this code. This code is not implemented to handle EVM errors. It is implemented to have a standard way to handle errors among all namespaces. `ProviderError` will be our standard error type to throw everywhere we encounter an error related to any provider. This class extends Error and adds a type property to it to indicate the cause of the error and we can take action based on this property everywhere. It's not a namespace specific error to handle things like `code`, `data` and `cause`. It meant to be the interface of the hub package to handle errors.
Handling data and code should take place on namespace or provider level like the references you have provided as it is implemented in `standardizeAndThrowEIP1193Error` on Evm namespace. Errors on other namespaces may be different from Evm errors and do not contain these `code` and `data` properties and handling those errors will take place on the specific providers and namespace but at the end they will throw a standard `ProviderError`.
This code is just inspired by EIP-1193 to have a thorough list of provider errors.

yeager-eren: There are two separate concerns,
1. Wallet/Provider should follow `eip 1193` to throw error, if they not, we will do it on behalf of them to ensure a consistence error throwing. After that, all the libs (including us) can have their own logic around that in error handling cases.
2. Throwing an standard error from Hub (includes Provider)


The first concern should be exactly what described in the spec and I also commented. 

The second concern doesn't add any value here for us yet, we will use plain Error everywhere needed. I don't see any use case that differentiate the Error would help to solve any thing.

If I'm missing something here, please let me know.

wallets/provider-trustwallet/src/namespaces/solana.ts
yeager-eren: When you are using an util called `standardizeAndThrowEIP1193Error` it means the provider/wallet doesn't throwing the standard error and you are trying to fix that.

If there are no links you can remove the comment or you can describe why the actions exists in the first place in `evm/or.ts` (if needed.)

wallets/core/src/errors/provider.ts
yeager-eren: Also want to mention that, If other namespaces are following the EVM spec, it doesn't change the fact that the origin come from EVM so it not a universal spec. the right place for that is still `evm` which can be used by other namespace as well, probably they do for backward compat or lack of that standard in other namespaces.

wallets/provider-trustwallet/src/namespaces/solana.ts
RyukTheCoder: Removed it as we decided to remove standard ProviderError.

wallets/core/src/errors/provider.ts
RyukTheCoder: Removed it as we decided that we do not need `ProviderError` and only handle the use case with a util.

wallets/core/src/errors/provider.ts
RyukTheCoder: Removed it as we decided that we do not need ProviderError and only handle the use case with a util.

wallets/core/src/errors/provider.ts
RyukTheCoder: Removed it as we decided that we do not need ProviderError and only handle the use case with a util.

yeager-eren (approved):
neat code. thanks. lgtm.

PR #1182, opened 2025-06-11
fix: handle long names and numbers

mikasackermn:
# Summary

Implement a unified truncation strategy for long cryptocurrency names, wallet addresses, and numeric values to ensure proper display across all wallet interface components on mobile viewports.

Fixes # (issue)

- [x] Truncation implemented for token names and amounts in the swap widget
- [x] Route selection component now gracefully handles long token pair names
- [x] Modal dialogs (e.g., token selection, confirmations) display truncated content correctly
- [x] Confirm Swap and Swap Details pages support truncation for long wallet addresses and transaction IDs
- [x] History page displays truncated token names and amounts in transaction lists
- [x] Input fields properly format and truncate large numeric values for better readability

# Checklist:

### DIFF
--- a/widget/embedded/src/components/CustomTokenModal/CustomTokenModal.styles.ts
+++ b/widget/embedded/src/components/CustomTokenModal/CustomTokenModal.styles.ts
@@ -1,4 +1,4 @@
-import { Button, darkTheme, styled } from '@rango-dev/ui';
+import { Button, darkTheme, styled, Typography } from '@rango-dev/ui';
 
 export const StyledLink = styled('a', {
   textDecoration: 'none',
@@ -58,4 +58,17 @@ export const Container = styled('div', {
 
 export const StyledButton = styled(Button, {
   minHeight: '$40',
+});
+
+export const TokenSymbolText = styled(Typography, {
+  overflow: 'hidden',
+  textOverflow: 'ellipsis',
+  whiteSpace: 'nowrap',
+  maxWidth: '96px',
+});
+
+export const FlexCenteredContainer = styled('div', {
+  display: 'flex',
+  justifyContent: 'center',
+  alignItems: 'center',
 });
--- a/widget/embedded/src/components/CustomTokenModal/CustomTokenModal.tsx
+++ b/widget/embedded/src/components/CustomTokenModal/CustomTokenModal.tsx
@@ -1,7 +1,14 @@
 import type { PropTypes } from './CustomTokenModal.types';
 
 import { i18n } from '@lingui/core';
-import { Divider, ExternalLinkIcon, Image, Typography } from '@rango-dev/ui';
+import {
+  Divider,
+  ExternalLinkIcon,
+  Image,
+  InfoIcon,
+  Tooltip,
+  Typography,
+} from '@rango-dev/ui';
 import React from 'react';
 
 import { DEFAULT_TOKEN_IMAGE_SRC } from '../../constants/customTokens';
@@ -10,8 +17,15 @@ import { WatermarkedModal } from '../common/WatermarkedModal';
 
 import { CUSTOM_TOKEN_LEARN_MORE_LINK } from './CustomTokenModal.constants';
 import { generateExplorerLink } from './CustomTokenModal.helpers';
-import { Container, StyledButton, StyledLink } from './CustomTokenModal.styles';
+import {
+  Container,
+  FlexCenteredContainer,
+  StyledButton,
+  StyledLink,
+  TokenSymbolText,
+} from './CustomTokenModal.styles';
 
+const MAX_SYMBOL_LENGTH_THRESHOLD = 9;
 export function CustomTokenModal(props: PropTypes) {
   const { open, onClose, token, onExit, onSubmitClick, blockchain } = props;
 
@@ -35,9 +49,17 @@ export function CustomTokenModal(props: PropTypes) {
           type="circular"
         />
         <Divider size={4} />
-        <Typography variant="title" size="medium">
-          {token.symbol}
-        </Typography>
+        <FlexCenteredContainer>
+          <TokenSymbolText variant="title" size="medium">
+            {token.symbol}
+          </TokenSymbolText>
+          {token.symbol.length > MAX_SYMBOL_LENGTH_THRESHOLD && (
+            <Tooltip content={token.symbol} container={getContainer()}>
+              <InfoIcon size={12} color="gray" />
+            </Tooltip>
+          )}
+        </FlexCenteredContainer>
+
         <Typography variant="body" size="small" className="_blockchain-name">
           {blockchain.displayName}
         </Typography>
--- a/widget/embedded/src/components/Quote/Quote.tsx
+++ b/widget/embedded/src/components/Quote/Quote.tsx
@@ -13,6 +13,7 @@ import {
   StepDetails,
   TokenAmount,
   Typography,
+  useIsTruncated,
 } from '@rango-dev/ui';
 import BigNumber from 'bignumber.js';
 import React, { useRef, useState } from 'react';
@@ -57,11 +58,9 @@ import { getTotalFeeInUsd, getUsdFeeOfStep } from '../../utils/swap';
 
 import {
   AllRoutesButton,
+  AmountText,
   BasicInfoOutput,
-  basicInfoStyles,
-  ContainerInfoOutput,
   Content,
-  FrameIcon,
   HorizontalSeparator,
   Line,
   QuoteContainer,
@@ -70,6 +69,8 @@ import {
   summaryHeaderStyles,
   summaryStyles,
   TagContainer,
+  TokenNameText,
+  UsdValueText,
 } from './Quote.styles';
 import { QuoteCostDetails } from './QuoteCostDetails';
 import { QuoteSummary } from './QuoteSummary';
@@ -97,11 +98,6 @@ export function Quote(props: QuoteProps) {
   const userSlippage = customSlippage || slippage;
   const [expanded, setExpanded] = useState(props.expanded);
   const quoteRef = useRef<HTMLButtonElement | null>(null);
-  const roundedInput = numberToString(
-    input.value,
-    TOKEN_AMOUNT_MIN_DECIMALS,
-    TOKEN_AMOUNT_MAX_DECIMALS
-  );
   const roundedOutput = numberToString(
     output.value,
     TOKEN_AMOUNT_MIN_DECIMALS,
@@ -241,6 +237,7 @@ export function Quote(props: QuoteProps) {
                         <Typography
                           size="xsmall"
                           variant="body"
+                          className="from-amount-text"
                           color="neutral900">
                           {i18n.t({
                             id: 'Yours: {amount} {symbol}',
@@ -328,6 +325,15 @@ export function Quote(props: QuoteProps) {
   const feeWarning = totalFee.gte(new BigNumber(GAS_FEE_MAX));
   const timeWarning =
     totalDurationSeconds / SECONDS_IN_MINUTE >= ROUTE_TIME_MAX;
+  const inputValueRef = useRef<HTMLSpanElement | null>(null);
+  const isInputValueTruncated = useIsTruncated(input.value, inputValueRef);
+  const outputValueRef = useRef<HTMLSpanElement | null>(null);
+  const isOutputValueTruncated = useIsTruncated(output.value, outputValueRef);
+  const usdValueRef = useRef<HTMLSpanElement | null>(null);
+  const isUsdValueTruncated = useIsTruncated(
+    roundedOutputUsdValue,
+    usdValueRef
+  );
 
   const lastStep = steps[numberOfSteps - 1];
   const firstStep = steps[0];
@@ -428,31 +434,52 @@ export function Quote(props: QuoteProps) {
           )}
         </div>
         {type === 'basic' && (
-          <div className={basicInfoStyles()}>
-            <FrameIcon>
-              <InfoIcon size={12} color="gray" />
-            </FrameIcon>
-            <ContainerInfoOutput>
-              <BasicInfoOutput size="small" variant="body">
-                {`${roundedInput} ${firstStep.from.token.displayName} = `}
-              </BasicInfoOutput>
+          <BasicInfoOutput>
+            <AmountText ref={inputValueRef} size="small" variant="body">
+              {input.value}
+            </AmountText>
+            {isInputValueTruncated && (
+              <NumericTooltip
+                content={input.value}
+                container={container}
+                open={!input.value ? false : undefined}>
+                <InfoIcon size={12} color="gray" />
+              </NumericTooltip>
+            )}
+            <TokenNameText size="small" variant="body">
+              {steps[0]?.from.token.displayName}
+            </TokenNameText>
+            <Typography size="small" variant="body">
+              =
+            </Typography>
+            <AmountText ref={outputValueRef} size="small" variant="body">
+              {output.value}
+            </AmountText>
+            {isOutputValueTruncated && (
               <NumericTooltip
                 content={output.value}
                 container={container}
                 open={!output.value ? false : undefined}>
-                <BasicInfoOutput size="small" variant="body">
-                  &nbsp;
-                  {`${roundedOutput} ${lastStep.to.token.displayName}`}
-                </BasicInfoOutput>
+                <InfoIcon size={12} color="gray" />
               </NumericTooltip>
-            </ContainerInfoOutput>
-            <NumericTooltip content={output.usdValue} container={container}>
-              <Divider size={2} direction="horizontal" />
-              <Typography color="$neutral600" size="xsmall" variant="body">
-                {`($${roundedOutputUsdValue})`}
-              </Typography>
-            </NumericTooltip>
-          </div>
+            )}
+            <TokenNameText size="small" variant="body">
+              {lastStep?.to.token.displayName}
+            </TokenNameText>
+            <Divider size={2} direction="horizontal" />
+            <UsdValueText
+              ref={usdValueRef}
+              color="$neutral600"
+              size="xsmall"
+              variant="body">
+              {`($${roundedOutputUsdValue})`}
+            </UsdValueText>
+            {isUsdValueTruncated && (
+              <NumericTooltip content={output.usdValue} container={container}>
+                <InfoIcon size={12} color="gray" />
+              </NumericTooltip>
+            )}
+          </BasicInfoOutput>
         )}
         {type === 'list-item' && (
           <TokenAmount
--- a/widget/embedded/src/components/QuoteWarningsAndErrors/HighValueLossWarningModal.tsx
+++ b/widget/embedded/src/components/QuoteWarningsAndErrors/HighValueLossWarningModal.tsx
@@ -16,8 +16,6 @@ import {
   GAS_FEE_MIN_DECIMALS,
   PERCENTAGE_CHANGE_MAX_DECIMALS,
   PERCENTAGE_CHANGE_MIN_DECIMALS,
-  USD_VALUE_MAX_DECIMALS,
-  USD_VALUE_MIN_DECIMALS,
 } from '../../constants/routing';
 import { getContainer } from '../../utils/common';
 import { numberToString } from '../../utils/numbers';
@@ -41,11 +39,7 @@ export function HighValueLossWarningModal(props: Props) {
   const highValueLossData = [
     {
       title: i18n.t('Swapping'),
-      value: numberToString(
-        warning.inputUsdValue,
-        USD_VALUE_MIN_DECIMALS,
-        USD_VALUE_MAX_DECIMALS
-      ),
+      value: numberToString(warning.inputUsdValue),
     },
     {
       title: i18n.t('Gas cost'),
@@ -57,11 +51,7 @@ export function HighValueLossWarningModal(props: Props) {
     },
     {
       title: i18n.t('Receiving'),
-      value: numberToString(
-        warning.outputUsdValue,
-        USD_VALUE_MIN_DECIMALS,
-        USD_VALUE_MAX_DECIMALS
-      ),
+      value: numberToString(warning.outputUsdValue),
     },
     {
       title: i18n.t('Price impact'),
--- a/widget/embedded/src/components/QuoteWarningsAndErrors/QuoteWarningsAndErrors.styles.ts
+++ b/widget/embedded/src/components/QuoteWarningsAndErrors/QuoteWarningsAndErrors.styles.ts
@@ -1,4 +1,4 @@
-import { Button, darkTheme, styled } from '@rango-dev/ui';
+import { Button, darkTheme, styled, Typography } from '@rango-dev/ui';
 
 export const Alerts = styled('div', {
   width: '100%',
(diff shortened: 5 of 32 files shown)


widget/ui/src/components/Alert/Alert.styles.ts
```diff
@@ -95,6 +95,9 @@ export const Main = styled('div', {
   display: 'flex',
   alignItems: 'center',
   alignSelf: 'stretch',
```
RyukTheCoder, line 97: The way you handled it caused every word to break and I'm not sure if it is the desired approach. Maybe we just want to break the "token name" not every word. For example "liquidity" is broken in this example.

image

widget/embedded/src/components/CustomTokenModal/CustomTokenModal.styles.ts
```diff
@@ -59,3 +59,16 @@ export const Container = styled('div', {
 export const StyledButton = styled(Button, {
   minHeight: '$40',
 });
+
+export const TokenSymbolText = styled(Typography, {
+  overflow: 'hidden',
+  textOverflow: 'ellipsis',
+  letterSpacing: 0.4,
+  whiteSpace: 'nowrap',
+  maxWidth: '97px',
```
RyukTheCoder: `maxWidth` is set `96px` in design and I don't see any changes to `letterSpacing` which is changed by you.

widget/ui/src/components/TokenAmount/TokenAmount.styles.ts
```diff
@@ -31,3 +32,14 @@ export const usdValueStyles = css({
...
 };
+
+export const TokenNameText = styled(Typography, {
+  maxWidth: '64px',
+});
+
+export const textTruncate = css({
+  overflow: 'hidden',
+  textOverflow: 'ellipsis',
+  letterSpacing: 0.4,
```
RyukTheCoder: I don't see `letterSpacing` change in design files.

RyukTheCoder:
Also cases should be handled based on the design files:

1. letter-spacing should be set to 1px in home page token input.
2. Should set max-width for blockchain name and usd price and quote items in home page.
image
image
image

RyukTheCoder (changes requested):
I just added some comments.

widget/ui/src/components/Alert/Alert.styles.ts
mikasackermn: This solution was suggested by the UI team, considering that we display the message inside a toast or alert and don't have direct access to the token name or amount.
Do you have any alternative suggestions?

widget/embedded/src/components/QuoteWarningsAndErrors/QuoteWarningsAndErrors.styles.ts
```diff
@@ -27,6 +27,21 @@ export const Item = styled('div', {
...
+  display: 'flex',
+  justifyContent: 'center',
+  alignItems: 'center',
+  gap: '$2',
+});
+
+export const ValueTypography = styled(Typography, {
+  overflow: 'hidden',
+  textOverflow: 'ellipsis',
+  letterSpacing: 0.4,
```
RyukTheCoder: I think letterSpacing is not changed in the design files. Please check all of your changes on letterSpacing.

widget/embedded/src/components/QuoteWarningsAndErrors/HighValueLossWarningModal.tsx
```diff
@@ -41,11 +41,7 @@ export function HighValueLossWarningModal(props: Props) {
   const highValueLossData = [
     {
       title: i18n.t('Swapping'),
-      value: numberToString(
```
RyukTheCoder, line 44: Why these Min and Max values are removed? and if they should get removed maybe they should be removed for receiving value as well.

widget/ui/src/components/PriceImpact/PriceImpact.styles.ts
```diff
@@ -13,6 +13,7 @@ export const ValueTypography = styled('div', {
   overflow: 'hidden',
   textOverflow: 'ellipsis',
   '& ._typography': {
+    maxWidth: '71px',
```
RyukTheCoder: Can you explain how you got to this 71px value?

widget/embedded/src/components/QuoteWarningsAndErrors/QuoteWarningsAndErrors.styles.ts
mikasackermn: You're right; I reviewed all my changes to `letterSpacing` and removed them.

widget/embedded/src/components/QuoteWarningsAndErrors/HighValueLossWarningModal.tsx
mikasackermn: Since we need to show the real value, I'm not sure which part you're suggesting to remove.

widget/ui/src/components/PriceImpact/PriceImpact.styles.ts
mikasackermn: On the swap confirmation page, the max width for the price impact is 71 pixels. Also, in the "To" input, the price section is truncated just like in the "From" section. It has the exact same width, and the 71px size doesn't cause any issues.

widget/embedded/src/components/QuoteWarningsAndErrors/HighValueLossWarningModal.tsx
RyukTheCoder: You have changed "Swapping" value in `highValueLossData` to prevent setting min and max decimals for it. First I want to know why this is changed. And after that, there is another similar property in `highValueLossData` titled "Receiving" and the value of that is `outputUsdValue` but you did not do the same for that. If Swapping value should change, why Receiving value should not change?

widget/ui/src/components/PriceImpact/PriceImpact.styles.ts
RyukTheCoder: I don't see a 71px max-width for this typography in the design files. I see a typography with 77px width in the design files. Please double check this.

mikasackermn: I checked, and you’re right. I changed it to 77px.

widget/embedded/src/components/QuoteWarningsAndErrors/HighValueLossWarningModal.tsx
mikasackermn: I changed `Receiving` to be the same as `Swapping`. We need to show the real value for both `Receiving` and `Swapping`

widget/ui/src/hooks/useIsTruncated.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+export function useIsTruncated(
+  content: React.ReactNode,
+  ref: RefObject<HTMLElement>
+): boolean {
+  const [isTruncated, setIsTruncated] = useState(false);
+
+  useEffect(() => {
+    const element = ref.current;
+    console.log({ element });
```
RyukTheCoder: Remove console log

widget/ui/src/hooks/useIsTruncated.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+    if (!element) {
+      return;
+    }
+    console.log({ content });
```
RyukTheCoder: Remove console log

widget/ui/src/hooks/useIsTruncated.tsx
```diff
@@ -0,0 +1,24 @@
+import { type RefObject, useEffect, useState } from 'react';
+
+export function useIsTruncated(
```
RyukTheCoder, line 8: Please add comments and explain what this hook does and what are the props. Also this hook only sets `isTruncated` to true. If it should contain a dynamic logic, it also should set `isTruncated` to false when needed.

widget/ui/src/containers/SwapInput/SwapInput.tsx
```diff
@@ -157,17 +158,19 @@ export function SwapInput(props: SwapInputPropTypes) {
...
-                    <UsdPrice variant="body" size="medium">
-                      {displayUsdValue}
-                    </UsdPrice>
-                  </ValueTypography>
-                </NumericTooltip>
+                <ValueTypography hasWarning={!!price.error}>
+                  <UsdPrice variant="body" size="medium">
+                    {displayUsdValue}
+                  </UsdPrice>
+                  {displayUsdValue.length > USD_VALUE_TOOLTIP_THRESHOLD && (
```
RyukTheCoder: This gets truncated even when `percentageChange` is not provided and there exists extra available space for this to grow.

RyukTheCoder: Also it seems that this is only added here to be displayed when `percentageChange` does not exist as a prop and if not, `PriceImpact` component will be rendered. In the `PriceImpact` it seems that it checks again for `percentageChange` and I think we can use that directly here. Also truncation is handled in both places. Please check if we can reuse this logic here.

widget/embedded/src/components/CustomTokenModal/CustomTokenModal.tsx
```diff
@@ -35,9 +49,17 @@ export function CustomTokenModal(props: PropTypes) {
...
         />
         <Divider size={4} />
-        <Typography variant="title" size="medium">
-          {token.symbol}
-        </Typography>
+        <FlexCenteredContainer>
+          <TokenSymbolText variant="title" size="medium">
+            {token.symbol}
+          </TokenSymbolText>
+          {token.symbol.length > VALUE_LENGTH_THRESHOLD && (
```
RyukTheCoder: It is defined as `VALUE_LENGTH_THRESHOLD` but symbol length is being checked with it.

widget/embedded/src/constants/routing.ts
```diff
@@ -15,3 +15,5 @@ export const PERCENTAGE_CHANGE_MAX_DECIMALS = 2;
 
 export const BALANCE_MIN_DECIMALS = 8;
 export const BALANCE_MAX_DECIMALS = 8;
+
+export const VALUE_LENGTH_THRESHOLD = 12;
```
RyukTheCoder: I do not see this being imported anywhere.

widget/ui/src/components/StepDetails/StepDetails.tsx
```diff
@@ -149,11 +152,23 @@ const StepDetailsComponent = forwardRef<HTMLDivElement, StepDetailsProps>(
                 content={step.from.price.realValue}
                 container={tooltipContainer}>
                 <Divider direction="horizontal" size={4} />
-                <Typography
+                <ValueTypography
                   size="small"
                   color="$neutral700"
-                  variant="body">{`${step.from.price.value} ${step.from.token.displayName}`}</Typography>
+                  variant="body">{`${step.from.price.realValue}`}</ValueTypography>
+                <Divider direction="horizontal" size={2} />
```
RyukTheCoder: These dividers are rendered regardless of `InfoIcon`  being rendered or not.

widget/ui/src/containers/SwapInput/SwapInput.tsx
RyukTheCoder: image

widget/embedded/src/components/Quote/Quote.tsx
```diff
@@ -429,29 +436,54 @@ export function Quote(props: QuoteProps) {
...
+                {isOutputValueTruncated && (
+                  <NumericTooltip
+                    content={output.value}
+                    container={container}
+                    open={!output.value ? false : undefined}>
+                    <InfoIcon size={12} color="gray" />
+                  </NumericTooltip>
+                )}
+                <TokenNameText size="small" variant="body">
+                  {steps[steps.length - 1]?.to.token.displayName}
```
RyukTheCoder:
```suggestion
                  {lastStep.to.token.displayName}
```

widget/ui/src/components/FullExpandedQuote/FullExpandedQuote.TokenSection.tsx
```diff
@@ -32,12 +38,17 @@ export function TokenSection(props: TokenSectionPropTypes) {
           container={tooltipProps?.container}
           side="bottom"
           open={tooltipProps?.open}>
-          <Typography
+          <AmountText
             size="xsmall"
             variant="body"
-            className={isInternalSwap ? tokenLabelStyles() : ''}>
+            className={`${textTruncate()} ${
+              isInternalSwap ? tokenLabelStyles() : ''
```
RyukTheCoder: `tokenLabelStyles` is an empty object. I do not get what it does.

widget/ui/src/components/TokenAmount/TokenAmount.tsx
```diff
@@ -1,79 +1,189 @@
...
   return (
     <Container
       direction={props.direction}
       centerAlign={props.centerAlign}
       id={props.id}>
-      <div className={tokenAmountStyles()}>
+      <div
+        className={tokenAmountStyles()}
+        ref={leftBoxRef}
+        style={{
```
RyukTheCoder: We do not use inline styles. Please check for it in all usages in this file.

widget/ui/src/components/TokenAmount/TokenAmount.tsx
```diff
@@ -1,79 +1,189 @@
...
+  const isRealValueTruncated = useIsTruncated(
+    props.price.realValue,
+    realValueRef
+  );
+  const realUSdValueRef = useRef<HTMLSpanElement | null>(null);
+  const isRealUsedValueTruncated = useIsTruncated(
+    props.price.realUsdValue,
+    realUSdValueRef
+  );
+  const leftBoxRef = useRef<HTMLDivElement | null>(null);
```
RyukTheCoder: `leftBox`  and `isLeftAtMax` are not clear names

widget/ui/src/components/TokenAmount/TokenAmount.tsx
```diff
@@ -1,79 +1,189 @@
...
+  const [isLeftAtMax, setIsLeftAtMax] = useState(false);
+
+  useLayoutEffect(() => {
```
RyukTheCoder: Please explain what this does. Maybe it can be custom hook.

widget/ui/src/components/TokenAmount/TokenAmount.tsx
```diff
@@ -1,79 +1,189 @@
...
-              </Typography>
-            </NumericTooltip>
+          <div style={{ minWidth: 0 }} className={centeredFlexBox()}>
+            <Typography
+              className={textTruncate()}
+              ref={realValueRef}
+              size="medium"
+              variant="title"
+              style={{
+                fontWeight: 600,
```
RyukTheCoder: `fontWeight` should be specified for typography using `size` and `variant`.

widget/ui/src/hooks/useIsTruncated.tsx
mikasackermn: Removed console log

widget/ui/src/hooks/useIsTruncated.tsx
mikasackermn: I’ve added comments to explain the hook and its props, and also updated the logic so that isTruncated is set to false when the content fully fits inside the element.

widget/embedded/src/components/CustomTokenModal/CustomTokenModal.tsx
mikasackermn: I renamed it to MAX_SYMBOL_LENGTH_THRESHOLD.

widget/embedded/src/constants/routing.ts
mikasackermn: It was deleted

widget/ui/src/components/StepDetails/StepDetails.tsx
mikasackermn: I moved the divider related to the `InfoIcon` inside the condition.

widget/ui/src/components/FullExpandedQuote/FullExpandedQuote.TokenSection.tsx
mikasackermn: In `TokenInfo`, a style is defined for it.

widget/ui/src/components/TokenAmount/TokenAmount.tsx
mikasackermn: I’ve completely revised this file and aligned it with the Next branch.

widget/ui/src/components/TokenAmount/TokenAmount.tsx
mikasackermn: I've changed the names to` isMaxTokenAmount` and `tokenAmountRef`

widget/ui/src/components/TokenAmount/TokenAmount.tsx
mikasackermn: In fact, it takes the size of the div we set and compares it with the max width we provided. If the occupied size is equal to or larger than the max width, it returns `true`.

widget/ui/src/components/TokenAmount/TokenAmount.tsx
mikasackermn: This was already there before

widget/ui/src/containers/SwapInput/SwapInput.tsx
mikasackermn: I checked, and we can use `priceImpact` directly here without having duplicate logic.
I removed the max width for `priceImpact` so it can grow as much as space allows.
But in two places I gave it a max width:

- In the amount token, which was checked with the `UI` and needs a max width of 88.
- In the full expanded view, where we also needed to set a max width.

widget/ui/src/components/FullExpandedQuote/FullExpandedQuote.TokenSection.tsx
RyukTheCoder: So we can have a className here.
```suggestion
              isInternalSwap ? "internal_swap_amount_typography" : ''
```

and update the usage like this:
```
-  [`& .${tokenLabelStyles}`]: {
+  '& .internal_swap_amount_typography': {
```

widget/ui/src/containers/SwapInput/SwapInput.tsx
```diff
@@ -144,32 +135,17 @@ export function SwapInput(props: SwapInputPropTypes) {
...
-              )}
+              <PriceImpact
+                size="medium"
+                tooltipProps={{
+                  container: props.tooltipContainer,
+                  side: 'bottom',
+                }}
+                outputUsdValue={price.usdValue}
+                realOutputUsdValue={price.realUsdValue}
+                {...props}
```
RyukTheCoder: I don't get why all the props should be passed to `PriceImpact`

widget/ui/src/components/PriceImpact/PriceImpact.tsx
```diff
@@ -21,25 +24,32 @@ export function PriceImpact(props: PriceImpactPropTypes) {
...
-              : undefined
-          }
-          side={tooltipProps?.side}>
-          <ValueTypography>
-            <Typography size={size} variant="body" color={outputColor}>
-              {outputUsdValue === '0' ? '0.00' : `~$${outputUsdValue}`}
-            </Typography>
-          </ValueTypography>
-        </NumericTooltip>
+        <ValueTypography className={textTruncate()}>
```
RyukTheCoder, line 35: `ValueTypography ` is a div so it is not a Typography

widget/ui/src/hooks/useIsTruncated.tsx
```diff
@@ -0,0 +1,27 @@
+import { type RefObject, useEffect, useState } from 'react';
+
+/*
+ * This hook checks whether the text (or content) inside a React element fits within its width.
+ * In other words, it determines if the content is truncated or fully visible.
+ */
+
+export function useIsTruncated(
+  content: React.ReactNode,
```
RyukTheCoder: The useEffect runs on every render when content changes, but content is a React node that could be a new object reference each time, causing unnecessary re-runs.

RyukTheCoder: The hook doesn't respond to container size changes, only content changes. Is it needed?

widget/ui/src/components/TokenAmount/TokenAmount.tsx
RyukTheCoder: It is hard for me to read and understand

widget/ui/src/components/TokenAmount/TokenAmount.tsx
```diff
@@ -1,79 +1,154 @@
...
-              </Typography>
-              <Divider direction="horizontal" size={8} />
-              <Typography
-                size="medium"
-                variant="title"
-                style={{ fontWeight: 400 }}>
-                {props.token.displayName}
-              </Typography>
-            </NumericTooltip>
+          <div className={`${flexShrinkFix()} ${flexRowAligned()}`}>
```
RyukTheCoder: Is `flexRowAligned` defined like this to be reused? It is not used anywhere else

widget/ui/src/containers/SwapInput/SwapInput.tsx
RyukTheCoder: It causes some errors in console

Image

RyukTheCoder (changes requested):

widget/ui/src/components/FullExpandedQuote/FullExpandedQuote.TokenSection.tsx
mikasackermn: We were using `tokenLabelStyles` in two places. Since the second one needed its own scope, I added a dedicated class name for it.

widget/ui/src/components/PriceImpact/PriceImpact.tsx
mikasackermn: I only passed the props that were needed. I just had to add a condition so that if `percentageChange` exists in props, it gets passed; otherwise, it sends undefined. Without this condition, we were getting a typing error.

widget/ui/src/hooks/useIsTruncated.tsx
mikasackermn: I changed the content type to string. Since the widget has a fixed size, it only needs to respond to content changes.

widget/ui/src/components/TokenAmount/TokenAmount.tsx
mikasackermn: I refactored the wrapper div into a styled component `TokenInfoRow` to keep the code cleaner and more consistent.

widget/ui/src/components/FullExpandedQuote/FullExpandedQuote.TokenSection.tsx
RyukTheCoder: When I search for `tokenLabelStyles` it is just exported and not used anywhere.

mikasackermn: removed `tokenLabelStyles`

PR #1185, opened 2025-06-16
feat: migrate ledger provider to use hub

RyukTheCoder:
# Summary

Ledger PR is migrated to use hub. All the changes are related to a typical provider migration except that `derivationPath` is added to parameters of connect action of Evm and Solana actions and also derivationPath is passed to it in `useHubAdapter`.

Tested by connecting to Ledger provider.

# Checklist:

### DIFF
--- a/queue-manager/rango-preset/src/actions/executeTransaction.ts
+++ b/queue-manager/rango-preset/src/actions/executeTransaction.ts
@@ -17,6 +17,7 @@ import {
   isNetworkMatchedForTransaction,
   isRequiredWalletConnected,
   isWalletNull,
+  resetNetworkStatus,
   signTransaction,
   updateNetworkStatus,
 } from '../helpers';
@@ -103,12 +104,15 @@ export async function executeTransaction(
     requestBlock(blockedFor);
     return;
   }
-  // Update network to mark it as network changed successfully.
-  updateNetworkStatus(actions, {
-    message: '',
-    details: 'The network has been successfully changed.',
-    status: PendingSwapNetworkStatus.NetworkChanged,
-  });
+
+  if (currentStep.networkStatus === PendingSwapNetworkStatus.NetworkChanged) {
+    // Considering that network is matched now, if currently network status of the current step is equal to `NetworkChanged`, we need to update the network status to mark it as network changed successfully.
+    updateNetworkStatus(actions, {
+      message: '',
+      details: 'The network has been successfully changed.',
+      status: PendingSwapNetworkStatus.NetworkChanged,
+    });
+  }
 
   /*
    *For avoiding conflict by making too many requests to wallet, we need to make sure
@@ -125,6 +129,8 @@ export async function executeTransaction(
     requestBlock(blockedFor);
     return;
   }
+
+  resetNetworkStatus(actions);
 
   // All the conditions are met. We can safely send the tx to wallet for sign.
   await signTransaction(actions);


wallets/core/src/namespaces/evm/types.ts
```diff
@@ -8,7 +8,10 @@ import type {
 export interface EvmActions
   extends AutoImplementedActionsByRecommended,
     CommonActions {
-  connect: (chain?: Chain | ChainId) => Promise<AccountsWithActiveChain>;
+  connect: (
+    chain?: Chain | ChainId,
+    derivationPath?: string
```
yeager-eren: `derivationPath` is only for some sort of wallets, and it's not a universal property among them. so it should be `options.derivationPath`

wallets/core/src/namespaces/solana/types.ts
```diff
@@ -7,7 +7,7 @@ import type {
 export interface SolanaActions
   extends AutoImplementedActionsByRecommended,
     CommonActions {
-  connect: () => Promise<Accounts>;
+  connect: (network?: unknown, derivationPath?: string) => Promise<Accounts>;
```
yeager-eren: solana doesn't have network. would be something like `connect: (options) => ....`

wallets/provider-ledger/src/namespaces/evm.ts
```diff
@@ -0,0 +1,59 @@
...
+  standardizeAndThrowLedgerError,
+} from '../utils.js';
+
+const connect = builders
+  .connect()
+  .action(async function (_context, chain, derivationPath) {
+    if (!derivationPath) {
+      throw new Error('Derivation Path can not be empty.');
+    }
+
```
yeager-eren, line 27: we can use `chain` value to be ETH, if not, and error can be thrown.

wallets/provider-ledger/src/namespaces/evm.ts
```diff
@@ -0,0 +1,59 @@
...
+    const result = await getEthereumAccounts();
+
+    const formatAccounts = result.accounts.map(
+      (account) =>
+        CAIP.AccountId.format({
+          address: account,
+          chainId: {
+            namespace: CAIP_NAMESPACE,
+            reference: ETHEREUM_CHAIN_ID,
```
yeager-eren: that would be better to not hardcode `ETHEREUM_CHAIN_ID` here and use use `result.chainId`. If we added support for other chains in future, this code works.

yeager-eren: also consider samething for L46.

yeager-eren: even same thing applies to solana's implementation.

yeager-eren (changes requested):

wallets/core/src/namespaces/evm/types.ts
RyukTheCoder: Applied

wallets/core/src/namespaces/solana/types.ts
RyukTheCoder: `network` is added here to make the interface of connect action compatible with other namespaces. If I remove it here I should an `if` statement in `useHubAdapter` to check for the type and only pass the network if the selected namespace for connect is equal to `EVM`. Also considering that in the future we need to add support for other SVM chains, maybe it is better to keep it as it is.

wallets/provider-ledger/src/namespaces/evm.ts
RyukTheCoder: Considering that `connect` gets called with `undefined` network in most places, adding this check will make problems in connecting ledger. Also considering that the goal of this PR is just to migrate the ledger provider and there is no checking on chain in the legacy code, I disagree with your comment and if you think adding this check is necessary we can investigate it more and add such check in another PR.

wallets/provider-ledger/src/namespaces/evm.ts
RyukTheCoder: Used `result.chainId` here and on L46. Also a similar approach applied for Solana namespace.

wallets/provider-ledger/src/namespaces/evm.ts
yeager-eren: good point. no need to change.

wallets/core/src/namespaces/solana/types.ts
yeager-eren: we should face it anyways, let's try to do it. we can work on it together if it's complicated.
the reason is all the VM's (e.g. SVM) won't work same as each other, depends on the case they can have differences, like Move VMs.
Besides that, we have other chains like TON, Bitcoin, etc which doesn't have any network, so we should act on this anyways.

RyukTheCoder: Removed `network` from here and added a type guard to handle it in `useHubAdapter`.

yeager-eren (approved):
thanks. lgtm

PR #1188, opened 2025-06-21
Fix/rf 2490 recommended slippage

mikasackermn:
# Summary

Previously, we always stored the recommended slippage value in the `slippage` field, even if it didn't match any of the predefined options (e.g., 0.5, 1, 3). This caused a UI bug where the slippage input and options in the Setting page would appear empty.

Fixes # (issue)

With this fix, if the recommended slippage matches one of the preset options, it is stored in the `slippage` field. Otherwise, it is saved as `customSlippage`.

- [ ] Test A
- [ ] Test B

# Checklist:


widget/embedded/src/pages/Home.tsx
```diff
@@ -137,8 +138,12 @@ export function Home() {
 
   const onChangeSlippage = (slippage: number | null) => {
     if (slippage) {
-      setSlippage(slippage);
```
arlert-armin, line 140: Using an early return would reduce nesting.

widget/embedded/src/pages/Home.tsx
```diff
@@ -137,8 +138,12 @@ export function Home() {
...
-      setCustomSlippage(null);
+      if (SLIPPAGES.includes(slippage)) {
+        setSlippage(slippage);
+        setCustomSlippage(null);
+        return;
+      }
+      setCustomSlippage(slippage);
     }
   };
```
arlert-armin: You can do it also for the first condition:
```suggestion
      return;
      }
      if (SLIPPAGES.includes(slippage)) {
        setSlippage(slippage);
        setCustomSlippage(null);
        return;
      }
      setCustomSlippage(slippage);
  };
```

PR #1190, opened 2025-06-22
feat: add status to textField

mikasackermn:
# Summary

Add support for a status prop in the text field that changes the border color based on the validation state (default, error, warning, success).

Fixes # (issue)

- [ ] Test A
- [ ] Test B

# Checklist:

### DIFF
--- a/widget/embedded/src/components/Slippage/Slippage.styles.ts
+++ b/widget/embedded/src/components/Slippage/Slippage.styles.ts
@@ -32,26 +32,3 @@ export const SlippageChip = styled(Chip, {
   width: '61px',
   flexShrink: 0,
 });
-
-export const SlippageTextFieldContainer = styled('div', {
-  outlineWidth: 1,
-  outlineStyle: 'solid',
-  borderRadius: '$xs',
-  flex: 1,
-  variants: {
-    status: {
-      safe: {
-        outlineColor: '$secondary500',
-      },
-      error: {
-        outlineColor: '$error500',
-      },
-      warning: {
-        outlineColor: '$warning500',
-      },
-      empty: {
-        outlineWidth: 0,
-      },
-    },
-  },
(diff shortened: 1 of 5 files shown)


widget/ui/src/components/TextField/TextField.styles.ts
```diff
@@ -5,6 +5,8 @@ export const InputContainer = styled('div', {
   display: 'flex',
   justifyContent: 'center',
   alignItems: 'center',
+  outlineWidth: 1,
```
RyukTheCoder: I checked the design files and saw that `border` is used but you've used `outline` which made height of input be more comparing to other inputs .

widget/embedded/src/components/Slippage/Slippage.styles.ts
```diff
@@ -10,6 +10,7 @@ export const SlippageChipsContainer = styled('div', {
   flexWrap: 'wrap',
   gap: '$10',
   [`& ${TextField}`]: {
+    height: '$32',
```
RyukTheCoder:
```suggestion
    boxSizing: 'content-box',
```
I think it's better to handle element taking extra space using `boxSizing` instead of adding a magical number for height.

mikasackermn: I was setting a fixed height on the input to prevent layout shifts when displaying success or error states. This was because the border was conditionally applied with different thicknesses (e.g., 1px), which caused the input’s height to increase.
Now, I’ve replaced that workaround with a cleaner solution: using a consistent border of `1px solid transparent` and only changing the `border-color` based on the input state. This way, the input always keeps the same dimensions, and we avoid unnecessary height changes without hardcoding the height.

RyukTheCoder (approved):
LGTM

PR #1192, opened 2025-06-23
feat: add disconnect wallet button to detached modal

RyukTheCoder:
# Summary

Added "Disconnect wallet" button to "Detached" modal and modified `handleDisconnect` method of `useStatefulConnect` to update the state to `needsNamespace` if it is on `detached`.

Fixes # (issue)

Tested by connecting Phantom wallet on some namespaces and clicking on "Disconnect wallet" button.

# Checklist:


widget/embedded/src/components/StatefulConnectModal/StatefulConnectModal.tsx
```diff
@@ -194,6 +195,9 @@ export function StatefulConnectModal(props: PropTypes) {
       {isOnDetached(getState) && (
         <Detached
           onConfirm={handleDetachedConfirm}
+          onDisconnectWallet={() =>
+            void handleDisconnect(getState().namespace.targetWallet)
```
mikasackermn, line 199: Instead of calling `getState()` multiple times, consider storing its result in a variable like `const state = getState();.` This improves readability and ensures consistency

RyukTheCoder: As you see, `isOnDetached(getState)` is working as a type guard here so if I want to apply your comment, I should update it as well and also I should update its usages and do the same thing for other type guards like `isOnDerivationPath` and `isOnNamespace`. Do you think that it's helpful and should be done in this PR?

mikasackermn: I don’t think it’s necessary to do that

mikasackermn (approved):
LGTM

PR #1194, opened 2025-06-29
fix: convert trust-wallet in-app browser chainId to hex string

arlert-armin:
# Summary
This PR addresses an issue with Trust Wallet's in-app browser, which returns the `chainId` as a number instead of a hexadecimal string. This was causing an error in the `accountId.format` function, which expects the `chainId` in hex format.

To fix this, I've added a conditional check to convert `chainId` to a hex string when it's returned as a number. This ensures compatibility with Trust Wallet without affecting other providers.

Fixes # (issue)
* Normalize `chainId` to hex string if it’s returned as a number.

- [ ] Test A
- [ ] Test B

# Checklist:

### DIFF
--- a/wallets/core/src/namespaces/evm/actions.ts
+++ b/wallets/core/src/namespaces/evm/actions.ts
@@ -16,7 +16,7 @@ import { CAIP_NAMESPACE } from './constants.js';
 import { getAccounts, switchOrAddNetwork } from './utils.js';
 
 export const recommended = [...commonRecommended];
-
+const CHAIN_ID_RADIX = 16;
 export function connect(
   instance: () => ProviderAPI
 ): FunctionWithContext<EvmActions['connect'], Context> {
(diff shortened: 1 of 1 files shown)


wallets/core/src/namespaces/evm/actions.ts
```diff
@@ -33,10 +33,20 @@ export function connect(
       await switchOrAddNetwork(evmInstance, chain);
     }
 
-    const chainId = await evmInstance.request({ method: 'eth_chainId' });
```
arlert-armin, line 36: This line seems to be unnecessary since we are already getting it in the next line in the `getAccounts` function.

RyukTheCoder (approved):
LGTM

wallets/react/src/hub/useHubAdapter.ts
```diff
@@ -242,12 +242,12 @@ export function useHubAdapter(params: UseAdapterParams): ProviderContext {
...
       }
 
-      wallet.getAll().forEach((namespace) => {
+      wallet.getAll().forEach(async (namespace) => {
         const namespaceShouldBeDisconnected =
           !namespaces || namespaces.includes(namespace.namespaceId);
         const namespaceIsConnected = namespace.state()[0]().connected;
         if (namespaceShouldBeDisconnected && namespaceIsConnected) {
-          return namespace.disconnect();
+          return await namespace.disconnect();
```
arlert-armin: We had a problem with the trust-wallet's in-app browser, which only disconnected the first namespace, and the second one didn't get disconnected (due to race conditions or whatever else). In this change, I'm awaiting for disconnect to prevent this error.

PR #1199, opened 2025-07-08
feat: implement retry mechanism for lazy imports

RyukTheCoder:
# Summary

Implemented retry mechanism for lazy imports to avoid errors related failing to fetch dynamically imported module.

Tested by signing some transaction using Phantom and Metamask.

# Checklist:

### DIFF
--- a/wallets/core/src/legacy/types.ts
+++ b/wallets/core/src/legacy/types.ts
@@ -139,6 +139,7 @@ export type ConnectResult = {
 };
 
 export type Providers = { [type in WalletType]?: InstanceType };
+export type NetworkProviderMap = Map<Networks, unknown>;
 
 export enum Events {
   CONNECTED = 'connected',
--- a/wallets/provider-bitget/src/signer.ts
+++ b/wallets/provider-bitget/src/signer.ts
@@ -1,16 +1,23 @@
+import type { LegacyNetworkProviderMap } from '@rango-dev/wallets-core/legacy';
 import type { SignerFactory } from 'rango-types';
 
 import { DefaultTronSigner } from '@rango-dev/signer-tron';
-import { getNetworkInstance, Networks } from '@rango-dev/wallets-shared';
+import {
+  getNetworkInstance,
+  Networks,
+  retryLazyImport,
+} from '@rango-dev/wallets-shared';
 import { DefaultSignerFactory, TransactionType as TxType } from 'rango-types';
 
 export default async function getSigners(
-  provider: any
+  provider: LegacyNetworkProviderMap
 ): Promise<SignerFactory> {
   const ethProvider = getNetworkInstance(provider, Networks.ETHEREUM);
   const tronProvider = getNetworkInstance(provider, Networks.TRON);
   const signers = new DefaultSignerFactory();
-  const { DefaultEvmSigner } = await import('@rango-dev/signer-evm');
+  const { DefaultEvmSigner } = await retryLazyImport(
+    async () => await import('@rango-dev/signer-evm')
+  );
   signers.registerSigner(TxType.EVM, new DefaultEvmSigner(ethProvider));
   signers.registerSigner(TxType.TRON, new DefaultTronSigner(tronProvider));
   return signers;
--- a/wallets/provider-brave/src/signer.ts
+++ b/wallets/provider-brave/src/signer.ts
@@ -1,17 +1,24 @@
+import type { LegacyNetworkProviderMap } from '@rango-dev/wallets-core/legacy';
 import type { SignerFactory } from 'rango-types';
 
-import { getNetworkInstance, Networks } from '@rango-dev/wallets-shared';
+import {
+  getNetworkInstance,
+  Networks,
+  retryLazyImport,
+} from '@rango-dev/wallets-shared';
 import { DefaultSignerFactory, TransactionType as TxType } from 'rango-types';
 
 import { CustomSolanaSigner } from './solana-signer.js';
 
 export default async function getSigners(
-  provider: any
+  provider: LegacyNetworkProviderMap
 ): Promise<SignerFactory> {
   const ethProvider = getNetworkInstance(provider, Networks.ETHEREUM);
   const solProvider = getNetworkInstance(provider, Networks.SOLANA);
   const signers = new DefaultSignerFactory();
-  const { DefaultEvmSigner } = await import('@rango-dev/signer-evm');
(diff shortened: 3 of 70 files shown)


wallets/provider-bitget/src/signer.ts
```diff
@@ -1,16 +1,22 @@
...
-import { getNetworkInstance, Networks } from '@rango-dev/wallets-shared';
+import {
+  getNetworkInstance,
+  Networks,
+  retryLazyImport,
+} from '@rango-dev/wallets-shared';
 import { DefaultSignerFactory, TransactionType as TxType } from 'rango-types';
 
 export default async function getSigners(
-  provider: any
```
arlert-armin, line 8: Let's have a similar way to handle this `any` among all providers.

wallets/provider-brave/src/signer.ts
```diff
@@ -1,17 +1,23 @@
...
+  getNetworkInstance,
+  Networks,
+  retryLazyImport,
+} from '@rango-dev/wallets-shared';
 import { DefaultSignerFactory, TransactionType as TxType } from 'rango-types';
 
 import { CustomSolanaSigner } from './solana-signer.js';
 
 export default async function getSigners(
-  provider: any
```
arlert-armin, line 9: Also in here.

wallets/provider-clover/src/signer.ts
```diff
@@ -1,17 +1,23 @@
...
+  getNetworkInstance,
+  Networks,
+  retryLazyImport,
+} from '@rango-dev/wallets-shared';
 import { DefaultSignerFactory, TransactionType as TxType } from 'rango-types';
 
 import { CustomSolanaSigner } from './solana-signer.js';
 
 export default async function getSigners(
-  provider: any
```
arlert-armin, line 9: And here.

wallets/provider-coin98/src/signer.ts
```diff
@@ -1,17 +1,23 @@
...
+  getNetworkInstance,
+  Networks,
+  retryLazyImport,
+} from '@rango-dev/wallets-shared';
 import { DefaultSignerFactory, TransactionType as TxType } from 'rango-types';
 
 import { CustomSolanaSigner } from './solana-signer.js';
 
 export default async function getSigners(
-  provider: any
```
arlert-armin, line 9: And here.

wallets/provider-coinbase/src/signer.ts
```diff
@@ -1,17 +1,23 @@
...
+  getNetworkInstance,
+  Networks,
+  retryLazyImport,
+} from '@rango-dev/wallets-shared';
 import { DefaultSignerFactory, TransactionType as TxType } from 'rango-types';
 
 import { CustomSolanaSigner } from './solana-signer.js';
 
 export default async function getSigners(
-  provider: any
```
arlert-armin, line 9: And here.

wallets/provider-cosmostation/src/signer.ts
```diff
@@ -1,16 +1,24 @@
...
-import { getNetworkInstance, Networks } from '@rango-dev/wallets-shared';
+import {
+  getNetworkInstance,
+  Networks,
+  retryLazyImport,
+} from '@rango-dev/wallets-shared';
 import { DefaultSignerFactory, TransactionType as TxType } from 'rango-types';
 
 export default async function getSigners(
-  provider: any
```
arlert-armin, line 7: And here.

wallets/provider-metamask/src/signer.ts
```diff
@@ -1,12 +1,16 @@
 import type { SignerFactory } from 'rango-types';
 
+import { retryLazyImport } from '@rango-dev/wallets-shared';
 import { DefaultSignerFactory, TransactionType as TxType } from 'rango-types';
 
 export default async function getSigners(
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
arlert-armin: We can use the `phantom` way of handling this. Alternatively, we could create a universal type in the `core`. However, since we already have a provider type for each namespace in `core`, we can either pipe them or use a single type for all of them. Your call.

RyukTheCoder: Defined `LegacyProviderMap` in `wallets/code` and used it for map providers. Also used `ProviderAPI` from `@rango-dev/wallets-core/namespaces/evm` for EVM instances.

wallets/core/src/legacy/types.ts
```diff
@@ -139,6 +139,7 @@ export type ConnectResult = {
 };
 
 export type Providers = { [type in WalletType]?: InstanceType };
+export type ProviderMap = Map<Networks, unknown>;
```
arlert-armin: The name is not descriptive enough. At first glance, it seems to be a map of providers or something. Please add some context to it.

wallets/provider-xdefi/src/cosmos-signer.ts
```diff
@@ -1,12 +1,17 @@
...
+  retryLazyImport,
+} from '@rango-dev/wallets-shared';
 import { SignerError, SignerErrorCode } from 'rango-types';
 
 /*
  * TODO - replace with real type
  * tslint:disable-next-line: no-any
  */
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
 type CosmosExternalProvider = any;
```
arlert-armin: We can define a ProviderAPI type for cosmos and use its type.

wallets/core/src/legacy/types.ts
RyukTheCoder: Changed to `NetworkProviderMap` https://github.com/rango-exchange/rango-client/pull/1199/commits/5fb1349e0fe9bf19a321bbe7475727deebfe73d6

wallets/provider-xdefi/src/cosmos-signer.ts
RyukTheCoder: I think you are getting distracted from the goal of this PR. I do not think such type refinements should be done in this PR. The goal of this PR is to **implement retry mechanism for lazy imports**. All the type refinements which were done in this PR are only done to handle lint errors inside changed files.

arlert-armin: I get it, but since we are forced to change this, letting this remain like this would create a technical debt.

RyukTheCoder: Tnx, applied.

PR #1216, opened 2025-07-25
chore: refactor changelog script and make it an standalone script

yeager-eren:
# Summary

Generate changelog using a separate script.

This PR includes:
- Upgrade changelog related packages.
- Making a separate script for generating changelog.
- Two types of changelog can be generated: 1. packages changelog (already had) 2. root changelog (manually created)
- We used to rely on `conventional-commits-cli` binary and its cli interface, but now we've switched to their nodejs package and are using stream instead.
Fixes RF-2536

### Flags

`--name`: 
generating changelog for an specific package. 

```shell
node ./scripts/changelog/command.mjs --name @rango-dev/wallets-core
```

`--save`:
By default, we print the output into the stdout. with this flag it will append the changelog to `CHANGELOG.md`.
In root it will write into `./CHANGELOG.md`, in packages it will write into the package's changelog.

```shell
node ./scripts/changelog/command.mjs --save
```

`--all`:
Extract the list of packages, and try to generate changelog for all of the workspace members:

```shell
node ./scripts/changelog/command.mjs --all
```

### Root changelog

From now on, we should update the root's package.json. don't need to tag that using git tags.
The reason behind this is the changelog's title includes a version always, to make that automatically we need to do this.

### TODO

We modified the root's changelog a bit, and it adds a `includes widget-emmbedded@0.0.0`. We can do it in a separate task.

in `scripts/common/changelog.mjs`, we have `generateChangelog`. it should be added here.
`conventional-changelog` has support for templates, it can be done by using writer context and options. [more info.](https://github.com/conventional-changelog/conventional-changelog/tree/master/packages/conventional-changelog-writer#options.

We have got a bug in github release notes, turns out the changelog is always empty because genertaeChangelog function needs to be run before tagging.

Publish script was using the changelog, ensure it's not affected in any ways.

Try to create a commit on a single package, and one where it modiefies different packages at the same time, it should generate changelog correctly.

Try to generate change log for the root.

# Checklist:


scripts/changelog/command.mjs
```diff
@@ -0,0 +1,97 @@
...
+async function generateChangelogForWorkspaceMembers(pkgs) {
+  for (const pkg of pkgs) {
+    const { from, commitsCount } = await getInfoBeforeGeneratingChangelog(pkg);
+
+    if (commitsCount > 0) {
+      console.log('name:', pkg.name);
+      console.log('from:', from || 'start (first release)');
+      console.log('commits:', commitsCount);
+
+      if (save) {
```
arlert-armin: Save is not declared, and also, why aren't we getting it from options like root?

scripts/changelog/command.mjs
```diff
@@ -0,0 +1,97 @@
...
+  const optionDefinitions = [
+    { name: 'name', type: String },
+    { name: 'save', type: Boolean },
+    { name: 'all', type: Boolean },
+  ];
+  const { name, save, all } = commandLineArgs(optionDefinitions);
+
+  if (name && all)
+    throw new Error('One of the --name or --all flag should be given');
+
```
arlert-armin, line 67: The error message is not clear enough; at first glance it seems that giving at least one of them is mandatory. However, we want to indicate that just one of them should be passed at a time, or none. Something like 'Use either --name, --all, or neither, but not both' is more clear.

scripts/common/changelog.mjs
```diff
@@ -1,49 +1,151 @@
...
+ * Create a write stream for the target package's changelog.
+ *
+ * @param {import("./typedefs.mjs").Package} pkg
+ * @returns {WriteStream}
+ */
+export function changelogFileStream(pkg) {
+  const changelogPath = path.join(pkg.location, 'CHANGELOG.md');
+  const file = fs.createWriteStream(changelogPath, {
+    encoding: 'utf-8',
+    flags: 'a',
```
arlert-armin: We were prepending new logs until now, but this line is appending.

scripts/common/changelog.mjs
```diff
@@ -1,49 +1,157 @@
...
+
+/**
+ *
+ * @param {import("./typedefs.mjs").Package} [pkg]
+ */
+export async function generateChangelogAndSave(pkg) {
+  return new Promise((resolve, reject) => {
+    const changelog = generateChangelog(pkg);
+
+    // we only need location for file stream, when pkg is undefined, we will point to root package.json
```
arlert-armin: I'm not sure what this has to do with package.json. I think you wanted to mention the root package.

arlert-armin (changes requested):
We have another issue with the `conventional-recommended-bump` package, which returns an empty object right now.

scripts/changelog/command.mjs
yeager-eren: I missed that. fixed. f9c53ade811c31699b664192fd15495adead3188

scripts/changelog/command.mjs
yeager-eren: this is where user wants to use these flags, so we don't need to worry whether they understand it also functions without them. I'm still not convinced to change it.

scripts/common/changelog.mjs
yeager-eren: Good catch! fixed. 6d5355f4032b0a57eeb6275b27db9a5b00b6f1f1

scripts/common/changelog.mjs
yeager-eren: You are right. changed the description. 25a6a03d9caa8b4bf57c46ca4e9e2d4c0a0de409

scripts/common/github.mjs
```diff
@@ -41,9 +41,11 @@ export async function getGithubReleaseFor(pkg) {
  * @param {import('./typedefs.mjs').Package} pkg
  */
 export async function makeGithubRelease(pkg) {
-  const notes = await generateChangelog(pkg, {
-    saveToFile: false,
-  });
+  const notes = '';
+  for await (chunk of generateChangelog(pkg)) {
```
arlert-armin:
```suggestion
  for await (let chunk of generateChangelog(pkg)) {
```

arlert-armin: There are a few inconsistencies with our current changelog format:

1. **Compare link**
   We don’t use a compare link in our headings — we just put the package name and version.

2. **Includes line**
   We should add the `_includes ..._` line like we currently do, e.g.:

   ```
   _includes `@rango-dev/widget-embedded@0.44.1`_
   ```

3. **Heading style**
   The package name should come first, followed by the version in brackets, and it should be a **second-level heading** (`##`), not first-level (`#`). Example:

   ```
   ## Widget [0.40.0] (2025-06-10)
   _includes `@rango-dev/widget-embedded@0.46.4`_
   ```

Currently, it shows as:

```
# [1.0.0] (2025-08-12)
```

— which doesn’t match the above format.

yeager-eren: here you go, 7b13e62f5ea839819bad68b8f285b6121c376653. 
the output should be matched with our current format.

PR #1217, opened 2025-07-26
chore: add widget examples

nikaaru:
# Summary

This PR adds widget examples to the existing `examples/ `directory, updates `eslint.config.mjs` to include an ignores: `['examples/**']` entry so ESLint skips that folder, and removes examples from the root `package.json` workspaces array to prevent its dependencies from inflating the root `node_modules`.

- [ ] Test A
- [ ] Test B

# Checklist:


.gitignore
```diff
@@ -41,4 +41,7 @@ yarn-error.log*
 storybook-static
 .yalc
 yalc.lock
-.nx
\ No newline at end of file
+.nx
+
+examples/**/node_modules
```
yeager-eren: no need to this. it's a global pattern that we should ignore **/**/node_modules.
I removed it and nothing bad happend.

.gitignore
```diff
@@ -41,4 +41,7 @@ yarn-error.log*
...
+examples/**/build
```
yeager-eren: not sure we can do this or not, for those examples that put build artificats in `/build`, that would be great to make them unified and all of them put the artificats in `dist` instead. in this way you can remove this as well

package.json
```diff
@@ -8,7 +8,6 @@
   },
   "workspaces": {
     "packages": [
-      "examples/*",
```
yeager-eren, line 11: by doing this, `queue-manager-demo` is not a member of workspace so our publish script won't upgrade its dependencis any more, I guess you need to use `latset` vesrion there as well.

.gitignore
nikaaru: I deleted it.

.gitignore
nikaaru: I moved all the build examples to the `dist` folder and deleted both of them.

package.json
nikaaru: I moved all the `@rango-dev` packages to the latest.

nikaaru:
@yeager-eren 

I moved `queue-manager-demo` under `queue-manager` directory ([chore: move queue-manager-demo to queue-manager directory](https://github.com/rango-exchange/rango-client/pull/1217/commits/809963ea11db0ac4694a60ffa5f5ba7be3ed6a61)) and restored the previous **`yarn.lock`**. When I removed `/examples` from the `workspaces`, Yarn recalculated **hoisting** and changed `yarn.lock`, which broke a few builds.

For example: `visx` `TooltipWithBounds` started failing due to mixed `@types/react`, and `@mysten/sui`’s `d.ts` began pointing to an unresolved subpath. Keeping the demo inside the queue-manager workspace restores the original hoisting and keeps `yarn.lock` stable, and avoids those regressions without touching root configs.

yeager-eren (approved):
lgtm

PR #1218, opened 2025-07-27
fix: fix undefined wallet type for transfer transactions

RyukTheCoder:
# Summary

According to that blockchain is not available in transfer transaction, it was resulting in that required wallet for the transfer transaction step not getting found and failing to find the correct wallet and signing transactions. It is fixed in this PR by checking for the source wallet address among the list of wallets in the swap.

Tested by signing transfer transactions using Phantom wallet.

# Checklist:


nikaaru (approved):
LGTM

PR #1222, opened 2025-07-30
chore: simplify the widget height calculation

Ikari-Shinji-re:
# Summary

The dynamic height feature of the widget hasn't worked reliably across all applications. This pull request aims to fix that: 
https://github.com/rango-exchange/rango-client/pull/1210

I discussed the current solution with @nikaaru , and he pointed out that it adds unnecessary complexity. We've now replaced it with a simpler approach in this PR.

Interestingly, we had initially used a similar method during early development, but it caused various issues and introduced implicit dependencies on the host app. This PR revisits that approach for further testing.

It should be tested in different environments, including apps with nested layouts and the iframe version of the widget.

Fixes # (issue)

- [ ] Test A
- [ ] Test B

# Checklist:

### DIFF
--- a/widget/embedded/src/components/Layout/Layout.styles.ts
+++ b/widget/embedded/src/components/Layout/Layout.styles.ts
@@ -25,7 +25,7 @@ export const Container = styled('div', {
       fixed: {
         minHeight: WIDGET_MIN_HEIGHT,
         maxHeight: WIDGET_MAX_HEIGHT,
-        height: WIDGET_MAX_HEIGHT,
+        height: '100%',
       },
     },
     showBanner: {
--- a/widget/app/src/App.tsx
+++ b/widget/app/src/App.tsx
@@ -50,7 +50,14 @@ export function App() {
 
   return (
     <Routes>
-      <Route path="/*" element={<Widget config={configRef.current} />} />
+      <Route
+        path="/*"
+        element={
+          <div style={{ height: '100vh' }}>
+            <Widget config={configRef.current} />
+          </div>
+        }
+      />
     </Routes>
   );
 }
--- a/widget/embedded/src/components/Layout/Layout.tsx
+++ b/widget/embedded/src/components/Layout/Layout.tsx
@@ -4,13 +4,12 @@ import type { PropsWithChildren } from 'react';
 
 import { useManager } from '@rango-dev/queue-manager-react';
 import { BottomLogo, Divider, Header } from '@rango-dev/ui';
-import React, { useEffect, useLayoutEffect, useRef, useState } from 'react';
+import React, { useEffect, useRef, useState } from 'react';
 
 import { WIDGET_UI_ID } from '../../constants';
 import { useIframe } from '../../hooks/useIframe';
 import { isAppLoadedIntoIframe } from '../../hooks/useIframe/useIframe.helpers';
 import { useNavigateBack } from '../../hooks/useNavigateBack';
-import useScreenDetect from '../../hooks/useScreenDetect';
 import { useTheme } from '../../hooks/useTheme';
 import { useAppStore } from '../../store/AppStore';
 import { tabManager, useUiStore } from '../../store/ui';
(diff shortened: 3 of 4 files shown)


widget/embedded/src/components/Layout/Layout.styles.ts
```diff
@@ -23,17 +23,30 @@ export const Container = styled('div', {
...
+      true: {},
+      false: {},
+    },
     showBanner: {
       true: {
         overflow: 'visible',
       },
     },
   },
+  compoundVariants: [
```
nikaaru: What do you think about setting the widget height to `100%` and setting a `minwidth` to ensure it doesn't have side effects for other DAPPs?
That is, should we keep the `minwidth` and `maxwidth`, and change them on different devices?

Something like this:
```ts
      css: {
        '@media screen and (max-height: 799px)': {
          minHeight: '500px',
          maxHeight: '700px'
        },
```

widget/embedded/src/components/Layout/Layout.styles.ts
```diff
@@ -23,17 +23,30 @@ export const Container = styled('div', {
...
+    {
+      height: 'fixed',
+      isIframe: false,
+      css: {
+        '@media screen and (max-height: 799px)': {
+          height: 500,
```
nikaaru: Please move it to the constant file.

```suggestion
          height: 500px,
```

widget/embedded/src/hooks/useScreenDetect.ts
```diff
@@ -35,6 +37,7 @@ const useScreenDetect = () => {
         window.innerWidth <= MIN_EXTRA_LARGE_SCREEN_WIDTH
     );
     setIsExtraLargeScreen(window.innerWidth > MIN_EXTRA_LARGE_SCREEN_WIDTH);
+    setIsLowHeightScreen(window.window.innerHeight < MAX_lOW_HEIGHT_DEVICE);
```
nikaaru:
```suggestion
    setIsLowHeightScreen(window.innerHeight < MAX_lOW_HEIGHT_DEVICE);
```

widget/embedded/src/hooks/useScreenDetect.ts
```diff
@@ -12,13 +12,15 @@ const MIN_TABLET_WIDTH = 480;
 const MIN_NOTEBOOK_WIDTH = 768;
 const MIN_LARGE_SCREEN_WIDTH = 1024;
 const MIN_EXTRA_LARGE_SCREEN_WIDTH = 1200;
+const MAX_lOW_HEIGHT_DEVICE = 800;
```
nikaaru: NIT: It's not a good descriptive name.

```suggestion
const MAX_LOW_SCREEN_HEIGHT = 800;
```

widget/iframe/src/main.ts
```diff
@@ -8,7 +8,7 @@ import {
 import { canParse } from './utils';
 
 // Actual app that will be loaded inside the iframe.
-const WIDGET_URL = 'https://widget.rango.exchange/';
+const WIDGET_URL = 'http://localhost:3002';
```
nikaaru: Why `WIDGET_URL` hardcoded to localhost?

widget/embedded/src/hooks/useScreenDetect.ts
Ikari-Shinji-re: fixed

widget/embedded/src/hooks/useScreenDetect.ts
Ikari-Shinji-re: fixed

widget/embedded/src/components/Layout/Layout.styles.ts
Ikari-Shinji-re: fixed

widget/iframe/src/main.ts
Ikari-Shinji-re: his is only for testing purposes, and I will revert the changes before merging the pull request. I’ve reverted the change and added the widget URL to the iframe modal example.

widget/embedded/src/components/Layout/Layout.styles.ts
Ikari-Shinji-re: Setting the widget height to 100% introduces side effects for apps already using our widget, similar to what we discussed earlier. In those cases, the parent element should have a specific height defined. With the current changes, the only side effect is that the widget no longer adjusts its height dynamically—it simply switches between two values.

nikaaru (approved):
LGTM

widget/iframe/src/main.ts
nikaaru: Reminder!

PR #1231, opened 2025-08-12
feat: implement changeAccountSubscriber builder

arlert-armin:
# Summary

* Implemented **`changeAccountSubscriberBuilder`** in `common/hooks.ts` with support for namespace-specific defaults.
* Updated all providers in **Hub** to use this builder for consistency and easier customization.
* Moved **`changeAccountSubscriber`** and **`changeChainSubscriber`** from `actions` to `hooks`.
* Applied the new builder-based approach across all existing providers, ensuring no functional changes but improving maintainability.

Fixes # (issue)

- [ ] Test A
- [ ] Test B

# Checklist:

### DIFF
--- a/wallets/core/src/namespaces/common/hooks/changeAccountSubscriber.ts
+++ b/wallets/core/src/namespaces/common/hooks/changeAccountSubscriber.ts
@@ -1,0 +1,188 @@
+import type {
+  Actions,
+  SubscriberCleanUp,
+} from '../../../hub/namespaces/types.js';
+import type { Subscriber } from '../../../mod.js';
+import type { AutoImplementedActionsByRecommended } from '../types.js';
+
+export class ChangeAccountSubscriberBuilder<EventType, ProviderAPI> {
+  #getInstance: (() => ProviderAPI) | null = null;
+  #format:
+    | ((instance: ProviderAPI, event: EventType) => Promise<string[]>)
+    | null = null;
+  #validateEventPayload: ((event: EventType) => boolean) | null = null;
+
+  #addEventListener:
+    | ((
+        instance: ProviderAPI,
+        callback: (event: EventType) => void
+      ) => (() => void) | void)
+    | null = null;
+  #removeEventListener:
+    | ((instance: ProviderAPI, callback: (event: EventType) => void) => void)
+    | null = null;
+
+  /**
+   * Sets the function that provides the provider API instance.
+   *
+   * @param operator - Function that returns the provider API instance
+   * @returns The builder instance for method chaining
+   * @example
+   * ```typescript
+   * builder.getInstance(() => window.ethereum)
+   * ```
+   */
+  public getInstance(operator: () => ProviderAPI) {
+    this.#getInstance = operator;
+    return this;
+  }
+
+  /**
+   * Sets the formatter function that converts provider events to account strings.
+   *
+   * @param operator - Function that takes a provider instance and event, returns array of account strings
+   * @returns The builder instance for method chaining
+   * @example
+   * ```typescript
+   * builder.format(async (instance, event) => event.accounts || [])
+   * ```
+   */
+  public format(
+    operator: (instance: ProviderAPI, event: EventType) => Promise<string[]>
+  ) {
+    this.#format = operator;
+    return this;
+  }
+
+  /**
+   * Sets the event payload validation function (optional).
+   *
+   * @param operator - Function that validates if an event payload is valid
+   * @returns The builder instance for method chaining
+   * @example
+   * ```typescript
+   * builder.validateEventPayload(event => !!event)
+   * ```
+   */
+  public validateEventPayload(operator: (event: EventType) => boolean) {
+    this.#validateEventPayload = operator;
+    return this;
+  }
+
+  /**
+   * Sets the event listener attachment function.
+   *
+   * @param operator - Function that attaches an event listener and optionally returns a cleanup function
+   * @returns The builder instance for method chaining
+   * @example
+   * ```typescript
+   * builder.addEventListener((instance, callback) => {
+   *   return instance.on('accountsChanged', callback);
+   * })
+   * ```
+   */
+  public addEventListener(
+    operator: (
+      instance: ProviderAPI,
+      callback: (event: EventType) => void
+    ) => (() => void) | void
+  ) {
+    this.#addEventListener = operator;
+    return this;
+  }
+
+  /**
+   * Sets the event listener removal function.
+   *
+   * @param operator - Function that removes an event listener from the provider
+   * @returns The builder instance for method chaining
+   * @example
+   * ```typescript
+   * builder.removeEventListener((instance, callback) => instance.off('accountsChanged', callback))
+   * ```
+   */
+  public removeEventListener(
+    operator: (
+      instance: ProviderAPI,
+      callback: (event: EventType) => void
+    ) => void
+  ) {
+    this.#removeEventListener = operator;
+    return this;
+  }
+  public build<
+    ActionsType extends Actions<ActionsType> &
+      Actions<AutoImplementedActionsByRecommended>
+  >(): [Subscriber<ActionsType>, SubscriberCleanUp<ActionsType>] {
+    if (this.#getInstance === null) {
+      throw new Error(this.#getErrorMessage('getInstance'));
+    }
+    if (this.#format === null) {
+      throw new Error(this.#getErrorMessage('format'));
+    }
+    if (this.#addEventListener === null) {
+      throw new Error(this.#getErrorMessage('addEventListener'));
+    }
+    if (this.#removeEventListener === null) {
+      throw new Error(this.#getErrorMessage('removeEventListener'));
+    }
+
+    /**
+     * Capture current operator state at build time to ensure immutability.
+     *
+     * This creates a snapshot of all operators, preventing the built subscriber
+     * from being affected by subsequent changes to the builder instance.
+     * Each call to build() gets its own isolated set of operators, allowing
+     * the builder to be reused for creating multiple independent subscribers.
+     */
+    const getInstance = this.#getInstance;
+    const format = this.#format;
+    const addEventListener = this.#addEventListener;
+    const removeEventListener = this.#removeEventListener;
+    const validateEventPayload = this.#validateEventPayload;
+    let subscriber: (event: EventType) => void;
+    let unsubscribe: (() => void) | void;
+    return [
+      async (context) => {
+        const [, setState] = context.state();
+        const instance = getInstance();
+
+        if (!instance) {
+          throw new Error(
+            'Trying to subscribe to your wallet, but seems its instance is not available.'
+          );
+        }
+        subscriber = async (event) => {
+          if (!!validateEventPayload && !validateEventPayload(event)) {
+            context.action('disconnect');
+            return;
+          }
+          setState('accounts', await format(instance, event));
+        };
+        unsubscribe = addEventListener(instance, subscriber);
+      },
+      (_, err) => {
+        /**
+         * Call the cleanup function if addEventListener returned one.
+         * This handles providers that return an unsubscribe function from their event listeners.
+         */
+        if (unsubscribe && typeof unsubscribe === 'function') {
+          unsubscribe();
+        }
+        const instance = getInstance();
+
+        /**
+         * Always call removeEventListener as well to handle the on/off pattern.
+         * This ensures cleanup works regardless of which pattern the provider uses.
+         */
+        removeEventListener(instance, subscriber);
+
+        // subscriber can be passed to `or`, it will get the error and should rethrow error to pass the error to next `or` or throw error.
+        return err;
+      },
+    ];
+  }
+  #getErrorMessage(operatorName: string) {
+    return `Required "${operatorName}" operation has not been set for "changeAccountSubscriber"`;
+  }
+}
--- a/wallets/core/src/namespaces/evm/builders.ts
+++ b/wallets/core/src/namespaces/evm/builders.ts
@@ -1,12 +1,18 @@
-import type { EvmActions } from './types.js';
+import type { EvmActions, ProviderAPI } from './types.js';
+
+import { AccountId } from 'caip';
 
 import { ActionBuilder } from '../../mod.js';
+import { ChangeAccountSubscriberBuilder } from '../common/hooks/changeAccountSubscriber.js';
 import {
   connectAndUpdateStateForMultiNetworks,
   intoConnecting,
   intoConnectionFinished,
 } from '../common/mod.js';
 
+import { CAIP_NAMESPACE } from './constants.js';
+
+// Actions
 export const connect = () =>
   new ActionBuilder<EvmActions, 'connect'>('connect')
     .and(connectAndUpdateStateForMultiNetworks)
@@ -17,3 +23,35 @@ export const canEagerConnect = () =>
   new ActionBuilder<EvmActions, 'canEagerConnect'>('canEagerConnect');
 export const canSwitchNetwork = () =>
   new ActionBuilder<EvmActions, 'canSwitchNetwork'>('canSwitchNetwork');
+
+// Hooks
+export const changeAccountSubscriber = (getInstance: () => ProviderAPI) =>
+  new ChangeAccountSubscriberBuilder<string[], ProviderAPI>()
+    .getInstance(getInstance)
+    .validateEventPayload(
+      (accounts) =>
+        /*
+         * In some wallets, when a user switches to an account not yet connected to the dApp, it returns null.
+         * A null value indicates no access to the account, requiring a disconnect and user reconnection.
+         * This behavior may vary across different wallets, and if so, a different approach may be needed.
+         */
+        accounts && accounts.length !== 0
+    )
+    .format(async (instance, accounts) => {
+      const chainId = await instance.request({ method: 'eth_chainId' });
+      return accounts.map((account) =>
+        AccountId.format({
+          address: account,
+          chainId: {
+            namespace: CAIP_NAMESPACE,
+            reference: chainId,
+          },
+        })
+      );
+    })
+    .addEventListener((instance, callback) => {
+      instance.on('accountsChanged', callback);
+    })
+    .removeEventListener((instance, callback) => {
+      instance.removeListener?.('accountsChanged', callback);
+    });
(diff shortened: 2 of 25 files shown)


wallets/core/src/namespaces/solana/builders.ts
```diff
@@ -1,14 +1,40 @@
...
+         * So null means we don't have access to account and we 0 need to disconnect and let the user connect the account.
+         */
+        !accounts
+    )
+    .setFormat(async (_, accounts) => formatAccountsToCAIP([accounts]))
+    .setAddEventListener((instance, callback) => {
+      instance.on('accountChanged', callback);
+    })
+    .setRemoveEventListener((instance, callback) => {
+      if (!callback && instance) {
```
RyukTheCoder:
```suggestion
      if (callback && instance) {
```

wallets/core/src/namespaces/sui/builders.ts
```diff
@@ -42,3 +49,34 @@ export const connect = (params: ConnectParams) =>
...
+    { accounts: readonly WalletAccount[] },
+    (() => void) | undefined,
+    ProviderAPI,
+    SuiActions
+  >()
+    .setGetInstance(() => getInstanceOrThrow(params.name))
+    .setShouldItDisconnect(
+      (event) =>
+        /*
+         * In Phantom, when user is switching to an account which is not connected to dApp yet, it returns a null.
```
RyukTheCoder: We do not need provider specific comments in the namespace.

wallets/core/src/namespaces/common/hooks/changeAccountSubscriber.ts
```diff
@@ -0,0 +1,112 @@
...
+  ) => void;
+};
+export class ChangeAccountSubscriberBuilder<
+  EventType,
+  AddEventListenerReturnType,
+  ProviderAPI extends SupportedProviderTypes,
+  ActionsType extends Actions<ActionsType> &
+    Actions<AutoImplementedActionsByRecommended>
+> {
+  private config = {} as ChangeAccountSubscriberBuilderConfig<
```
RyukTheCoder:
```suggestion
  private operations = {} as ChangeAccountSubscriberBuilderOperations<
```
I suggest to name this `operations`  as they are atomic units of work.

wallets/core/src/namespaces/common/hooks/changeAccountSubscriber.ts
```diff
@@ -0,0 +1,112 @@
...
+  setRemoveEventListener(
+    removeEventListener: (typeof this.config)['removeEventListener']
+  ) {
+    this.config.removeEventListener = removeEventListener;
+    return this;
+  }
+  build(): [Subscriber<ActionsType>, SubscriberCleanUp<ActionsType>] {
+    for (const configKey in this.config) {
+      if (!this.config[configKey as keyof typeof this.config]) {
+        throw new Error(`"${configKey}" has not been set`);
```
RyukTheCoder:
```suggestion
        throw new Error(`Required "${configKey}" operation has not been set for "changeAccountSubscriber"`);
```
Errors can be more informative to help make debugging easier.

wallets/core/src/namespaces/evm/builders.ts
```diff
@@ -17,3 +23,38 @@ export const canEagerConnect = () =>
   new ActionBuilder<EvmActions, 'canEagerConnect'>('canEagerConnect');
 export const canSwitchNetwork = () =>
   new ActionBuilder<EvmActions, 'canSwitchNetwork'>('canSwitchNetwork');
+
+// Hooks
+export const changeAccountSubscriber = (getInstance: () => ProviderAPI) =>
+  new ChangeAccountSubscriberBuilder<string[], void, ProviderAPI, EvmActions>()
```
RyukTheCoder:
```suggestion
  new ChangeAccountSubscriberBuilder<string[], undefined, ProviderAPI, EvmActions>()
```
Shouldn't we do this?

RyukTheCoder: You can define a `ChangeAccountSubscriberBuilder` in `namespaces/common` and attach all the operations (which are common between evm, solana, etc) except `formatAccounts` and in each namespace just use that one and attach the required `formatAccounts`.

wallets/core/src/namespaces/common/hooks/changeAccountSubscriber.ts
```diff
@@ -0,0 +1,112 @@
...
+  }
+  setRemoveEventListener(
+    removeEventListener: (typeof this.config)['removeEventListener']
+  ) {
+    this.config.removeEventListener = removeEventListener;
+    return this;
+  }
+  build(): [Subscriber<ActionsType>, SubscriberCleanUp<ActionsType>] {
+    for (const configKey in this.config) {
+      if (!this.config[configKey as keyof typeof this.config]) {
```
RyukTheCoder: This check makes every operation required. Shouldn't we leave some operations optional?

wallets/core/src/namespaces/common/hooks/changeAccountSubscriber.ts
```diff
@@ -0,0 +1,112 @@
...
+} from '../types.js';
+
+type ChangeAccountSubscriberBuilderConfig<
+  EventType,
+  ProviderAPI extends SupportedProviderTypes,
+  AddEventListenerReturnType
+> = {
+  getInstance: () => ProviderAPI;
+  format: (instance: ProviderAPI, event: EventType) => Promise<string[]>;
+  shouldItDisconnect: (event: EventType) => boolean;
```
RyukTheCoder: I do not like this `shouldItDisconnect`. What about having something like this:
```
handleEvent(event: EventType, options: {disconnect: () => void}) => boolean;
```
the return value indicates that if we should proceed
```
eventCallback = async (event) => {
const shouldProceed= this.operations.handleEvent(event, {disconnect: () => {context.action('disconnect');}})
if (!shouldProceed) return;
setState('accounts', await this.config.formatAccounts(instance, event));
}
```
usage:
```
.setHandleEvent(event, options) {
  if (!event) {options.disconnect(); return false;}
}
```

wallets/core/src/namespaces/solana/builders.ts
```diff
@@ -1,14 +1,40 @@
...
     .after(intoConnectionFinished);
+
+// Hooks
+export const changeAccountSubscriber = (getInstance: () => ProviderAPI) =>
+  new ChangeAccountSubscriberBuilder<string, void, ProviderAPI, SolanaActions>()
+    .setGetInstance(getInstance)
+    .setShouldItDisconnect(
+      (accounts) =>
+        /*
+         * In Phantom, when user is switching to an account which is not connected to dApp yet, it returns a null.
```
RyukTheCoder: We do not need provider specific comments in the namespace.

wallets/core/src/namespaces/common/hooks/changeAccountSubscriber.ts
arlert-armin: I've made `validateEventArgs` optional.

wallets/core/src/namespaces/common/hooks/changeAccountSubscriber.ts
arlert-armin: Good point done.

wallets/core/src/namespaces/common/hooks/changeAccountSubscriber.ts
arlert-armin: Done and also changed the type to `Operations`. Since it is a local type I think it doesn't need that much context.

wallets/core/src/namespaces/common/hooks/changeAccountSubscriber.ts
arlert-armin: I've changed it to `validateEventArgs` as we discussed.

wallets/core/src/namespaces/evm/builders.ts
arlert-armin: No, we can't. There is no single operation that could be shareable between all namespaces.

arlert-armin: Done.

wallets/core/src/namespaces/solana/builders.ts
arlert-armin: Generalized it.

wallets/core/src/namespaces/solana/builders.ts
arlert-armin: Good catch, Done.

wallets/core/src/namespaces/common/hooks/changeAccountSubscriber.ts
```diff
@@ -0,0 +1,121 @@
...
+        const [, setState] = context.state();
+        const instance = this.operations.getInstance();
+
+        if (!instance) {
+          throw new Error(
+            'Trying to subscribe to your wallet, but seems its instance is not available.'
+          );
+        }
+        eventCallback = async (event) => {
+          if (this.operations.validateEventArgs?.(event)) {
```
RyukTheCoder: We should disconnect if validate returns `false`. Right now you are disconnecting when it returns `true`
```suggestion
          if (!!this.operations.validateEventArgs && !this.operations.validateEventArgs.(event)) {
```

RyukTheCoder (changes requested):

wallets/core/src/namespaces/common/hooks/changeAccountSubscriber.ts
```diff
@@ -0,0 +1,121 @@
...
+} from '../types.js';
+
+type Operations<
+  EventType,
+  ProviderAPI extends SupportedProviderTypes,
+  AddEventListenerReturnType
+> = {
+  getInstance: () => ProviderAPI;
+  format: (instance: ProviderAPI, event: EventType) => Promise<string[]>;
+  validateEventArgs?: (event: EventType) => boolean;
```
RyukTheCoder: Based on its name, it seems that it checks the validity of event arguments passed to a function, But what you are trying to do with this callback is to validate the data or payload related to the event not it arguments of its handler.

wallets/core/src/namespaces/common/hooks/changeAccountSubscriber.ts
arlert-armin: Yes you are right,Applied it, Thanks.

wallets/core/src/namespaces/common/hooks/changeAccountSubscriber.ts
arlert-armin: I've change it to `validateEventPayload`.

wallets/core/src/namespaces/common/hooks/changeAccountSubscriber.ts
```diff
@@ -0,0 +1,124 @@
...
+    EventType,
+    ProviderAPI,
+    AddEventListenerReturnType
+  >;
+  private requiredFields: (keyof typeof this.operations)[] = [
+    'addEventListener',
+    'removeEventListener',
+    'format',
+    'getInstance',
+  ];
```
yeager-eren: There changes needs to be here:
1. don't use `private`, use `#` which is a native feature of js.
2. don't use assertion, it loses type safety.
3. take a look at ActionBuilder, you can store each operation as class properties, like `class ...{ #format; #getInstance; ...}`

wallets/core/src/namespaces/common/hooks/changeAccountSubscriber.ts
```diff
@@ -0,0 +1,124 @@
...
+  addEventListener: (
+    instance: ProviderAPI,
+    callback: (event: EventType) => void
+  ) => AddEventListenerReturnType;
+  removeEventListener: (
+    instance: ProviderAPI,
+    callback: (event: EventType) => void,
+    addEventListenerReturn: AddEventListenerReturnType
+  ) => void;
+};
```
yeager-eren: don't get why you made a generic for this. each method could be a separate. after my other comment, https://github.com/rango-exchange/rango-client/pull/1231/files#r2297988191, you can do this.

wallets/core/src/namespaces/common/hooks/changeAccountSubscriber.ts
```diff
@@ -0,0 +1,124 @@
...
+
+export class ChangeAccountSubscriberBuilder<
+  EventType,
+  AddEventListenerReturnType,
+  ProviderAPI extends SupportedProviderTypes,
+  ActionsType extends Actions<ActionsType> &
+    Actions<AutoImplementedActionsByRecommended>
+> {
```
yeager-eren: I'm not happy with this class signature. some changes need to be made here:

1. Remove `AddEventListenerReturnType` and `ActionsType`, don't need to be handled in class level, instead you can make the `build` a generic. I will explain how you can remove the first one totally in a separate comment. 

2.  Not sure why you needed `SupportedProviderTypes`. 

3. The desired signarutre for this will be `ChangeAccountSubscriberBuilder<ProviderAPI, EventType>`


for the first comment you can do:

```
  build<
    T extends Actions<ActionsType> &
      Actions<AutoImplementedActionsByRecommended>
  >(): [Subscriber<T>, SubscriberCleanUp<T>] 
```

wallets/core/src/namespaces/common/hooks/changeAccountSubscriber.ts
```diff
@@ -0,0 +1,124 @@
...
+          eventCallback
+        );
+      },
+      (_, err) => {
+        const instance = this.operations.getInstance();
+
+        this.operations.removeEventListener(
+          instance,
+          eventCallback,
+          addEventListenerReturn
```
yeager-eren: Usually there are two approches to clean up, by calling the subscriber function, it will return the clean up function (like SUI's standrad events, or react hooks), or there two separate methods but you need to pass the same callback (like `on|off` methods on eip providers).

Switching between these two shouldn't be the concern of clean up function.

I would suggest to change your code to handle it automaticly when the subscriber returns the clean up.

so it would be something like this:

```
// ...
if(addEventListenerReturn) addEventListenerReturn()

this.operations.removeEventListener(
          instance,
          eventCallback
}
// ...
```

yeager-eren: you can also rename addEventListenerReturn to a more meaningful one.

wallets/core/src/namespaces/common/hooks/changeAccountSubscriber.ts
```diff
@@ -0,0 +1,124 @@
+import type {
```
yeager-eren, line 1: Please try to follow the conventions in the codebase, if you take a look at `/builders`, we never used `setBlahBlah`. or all the public methods has `public`. each method has a jsdoc. it should be covered by test.

wallets/core/src/namespaces/solana/builders.ts
```diff
@@ -1,14 +1,45 @@
...
+         * A null value indicates no access to the account, requiring a disconnect and user reconnection.
+         */
+        !accounts
+    )
+    .setFormat(async (_, accounts) => formatAccountsToCAIP([accounts]))
+    .setAddEventListener((instance, callback) => {
+      instance.on('accountChanged', callback);
+    })
+    .setRemoveEventListener((instance, callback) => {
+      if (callback && instance) {
```
yeager-eren: I don't why you can not guarantee these values, in the type they are not optional, but you added a check in callbacks. 

you should either guarantee the value (adding your check to the builder), or types for `instance` and `callback` have a correct type and reflect them as optional.

wallets/core/src/namespaces/common/hooks/changeAccountSubscriber.ts
arlert-armin: Done.

wallets/core/src/namespaces/common/hooks/changeAccountSubscriber.ts
arlert-armin: Removed extra ones.

wallets/core/src/namespaces/common/hooks/changeAccountSubscriber.ts
arlert-armin: Put it in build as you said.

wallets/core/src/namespaces/common/hooks/changeAccountSubscriber.ts
arlert-armin: Done.

wallets/core/src/namespaces/common/hooks/changeAccountSubscriber.ts
arlert-armin: Renamed to unsubscribe and called it by default.

arlert-armin (commented):
I've also wrote tests for it.

wallets/core/src/namespaces/common/hooks/changeAccountSubscriber.ts
yeager-eren: most of them are resolved, but no jsdoc on top of the methods, and no comment on unsubscribe code as we talked (to explain we are running unsubscribe automatilcally and etc.)

wallets/core/src/namespaces/common/hooks/changeAccountSubscriber.ts
yeager-eren: no jsdoc on top of the methods, and no comment on unsubscribe code as we talked (to explain we are running unsubscribe automatilcally and etc.).
also add public before public methods.

wallets/core/src/namespaces/common/hooks/changeAccountSubscriber.ts
```diff
@@ -0,0 +1,115 @@
...
+        removeEventListener(instance, eventCallback);
+
+        // subscriber can be passed to `or`, it will get the error and should rethrow error to pass the error to next `or` or throw error.
+        return err;
+      },
+    ];
+  }
+  #getErrorMessage(operatorName: string) {
+    return `Required "${operatorName}" operation has not been set for "changeAccountSubscriber"`;
+  }
```
yeager-eren: nit:

const subscriber = () => {
// ..
}
const unsubscriber = () => {
// ..
}

return [subsciber, unsubscriber]

yeager-eren: also not sure why you are doing thsi:
const format = this.#format;

it's an arrow function, it still loose `this` access?

arlert-armin: Changed it to subscriber. 
I added a comment to help clear up your question.

wallets/core/src/namespaces/common/hooks/changeAccountSubscriber.ts
arlert-armin: I've added jsdoc and also added public before methods.

PR #1239, opened 2025-08-18
fix: handle dynamic import error

RyukTheCoder:
# Summary

Dynamic import error on `getSigners` method in `queue-manager` was not handled properly. With these changes, the correct error message will be displayed on fail to fetch dynamically imported modules and transaction will fail.

Also retry mechanism for dynamic imports was not working properly so it is removed in this PR.

## Places to check
- wallets/shared/src/helpers.ts
removed retry mechanism and added an error refinement
- queue-manager/rango-preset/src/helpers.ts
wrapped `await getSigners(sourceWallet.walletType);` in try/catch to handle dynamic import errors related to signers
- all signer.ts files
replaced `retryLazyImport` with `dynamicImportWithRefinedError`

Tested by throttling network to offline before attempting to download the bundle related to sign transaction and observing that transaction fails and proper error message gets displayed.

# Checklist:

### DIFF
--- a/queue-manager/rango-preset/src/helpers.ts
+++ b/queue-manager/rango-preset/src/helpers.ts
@@ -37,6 +37,7 @@ import type {
   PendingSwap,
   PendingSwapStep,
   SignerErrorCode,
+  SignerFactory,
   StepStatus,
 } from 'rango-types';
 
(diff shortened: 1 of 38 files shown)


arlert-armin (approved):
LGTM

PR #1253, opened 2025-09-08
fix: fix issue with checking version of providers

RyukTheCoder:
# Summary

There was an issue in checking for version of providers which was causing a crash in every dapp or example which uses `widget-embedded` package after migrating ledger provider to hub. The issue was related to that, the version of provider was being checked by this expression:
```
versionedProvider instanceof Provider
```
Given that `provider-ledger` relies on code splitting, it seems that a different clone of `ProviderBuilder` class is being used in `provider-ledger` which is different from the one we are checking the ledger built provider instance with it. So it causes that the expression always turn to false and result in unwanted flows and crash in each app or example which integrated the widget. In this PR, checking for version of provider is changed to checking for `version` property of the provider object.

Tested by linking the the build widget with new changes in vite example and dapp and observing that no crashes happen and connecting to wallets works properly.

# Checklist:

### DIFF
--- a/widget/embedded/src/utils/providers.ts
+++ b/widget/embedded/src/utils/providers.ts
@@ -1,12 +1,8 @@
 import type { WidgetConfig } from '../types';
+import type { Provider, VersionedProviders } from '@rango-dev/wallets-core';
 import type { LegacyProviderInterface } from '@rango-dev/wallets-core/legacy';
 
-import {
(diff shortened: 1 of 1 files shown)


RyukTheCoder:
The issue was fixed in https://github.com/rango-exchange/rango-client/pull/1255

PR #1255, opened 2025-09-09
fix: fix dynamic import issue in provider ledger

RyukTheCoder:
# Summary

There was an issue in dynamically importing signers in `provider-ledger`. The issue was related to that the paths `./signers/ethereum` and `./signers/solana` were being imported dynamically in the file `wallets/provider-ledger/src/signer.ts` . There was some problems with this approach which are explained below:
1. Most of the code being bundled in the chunks related to these paths was related to two packages '@ledgerhq/hw-app-solana' and '@ledgerhq/hw-app-eth' which were also being imported dynamically in `wallets/provider-ledger/src/utils.ts` and there is no need to create other chunks for `./signers/ethereum` and `./signers/solana` and we can only create new chunks for '@ledgerhq/hw-app-solana' and '@ledgerhq/hw-app-eth'.
2. Dynamically importing  `./signers/ethereum` and `./signers/solana` was making some issues because some other packages like `@rango-dev/wallets-shared` and `@rango-dev/wallets-core` were being used in these two files and was making bundlers to create a new chunk for these two packages to be shared between the chunks related to `./signers/ethereum` and `./signers/solana`. This was creating some issues like creating two different instances of classes like `ProviderBuilder` and `Provider` in the main chunk and the mentioned shared chunk which was creating some problems for checking the `instanceof` for objects which were created using these two classes.

In this PR, explained issues are fixed by removing dynamic imports for `./signers/ethereum` and `./signers/solana` in `wallets/provider-ledger/src/signer.ts` and only dynamically importing '@ledgerhq/hw-app-solana' and '@ledgerhq/hw-app-eth' everywhere they are used

Tested by linking the build result of `@rango-dev/widget-embedded` and `@rango-dev/provider-ledger` in the vite example and rango dapp and observing that everything works properly.

# Checklist:

### DIFF
--- a/wallets/provider-ledger/src/signers/ethereum.ts
+++ b/wallets/provider-ledger/src/signers/ethereum.ts
@@ -2,8 +2,10 @@ import type { TransactionLike } from 'ethers';
 import type { GenericSigner } from 'rango-types';
 import type { EvmTransaction } from 'rango-types/mainApi';
 
-import Eth, { ledgerService } from '@ledgerhq/hw-app-eth';
-import { DEFAULT_ETHEREUM_RPC_URL } from '@rango-dev/wallets-shared';
+import {
+  DEFAULT_ETHEREUM_RPC_URL,
+  dynamicImportWithRefinedError,
+} from '@rango-dev/wallets-shared';
 import { JsonRpcProvider, Transaction } from 'ethers';
 import { SignerError, SignerErrorCode } from 'rango-types';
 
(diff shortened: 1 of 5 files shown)


wallets/provider-ledger/src/signers/ethereum.ts
```diff
@@ -19,7 +21,11 @@ export class EthereumSigner implements GenericSigner<EvmTransaction> {
     try {
       const transport = await transportConnect();
```
yeager-eren: why not ‍‍
```
const ledgerEthApp =   await dynamicImportWithRefinedError(
          async () => await import('@ledgerhq/hw-app-eth')
        );

// then 
const eth = ...

const resolution = ...
```

yeager-eren: same applies to solana.

RyukTheCoder: Good suggestion. Applied.

yeager-eren (approved):
lgtm

PR #1259, opened 2025-09-21
feat: make detached confirm text customizable

RyukTheCoder:
# Summary

Added optional `confirmText` prop to `Detached` component enabling customizing the text of confirm button.

# Checklist:

### DIFF
--- a/widget/embedded/src/components/WalletStatefulConnect/Detached.tsx
+++ b/widget/embedded/src/components/WalletStatefulConnect/Detached.tsx
@@ -25,6 +25,7 @@ export function Detached(props: PropTypes) {
     value,
     selectedNamespaces,
     onConfirm,
+    confirmText = i18n.t('Done'),
     onDisconnectWallet,
     navigateToDerivationPath,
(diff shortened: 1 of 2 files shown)


widget/embedded/src/components/WalletStatefulConnect/Detached.tsx
```diff
@@ -25,6 +25,7 @@ export function Detached(props: PropTypes) {
     value,
     selectedNamespaces,
     onConfirm,
+    confirmText,
```
mikasackermn: Wouldn’t it be better to write it this way instead of adding a conditional?

```
const {
...props,
  confirmText = i18n.t('Done'),
} = props;

return (
  <StyledButton
    id="widget-name-space-confirm-btn"
    type="primary"
    onClick={onConfirm}
  >
    {confirmText}
  </StyledButton>
);
```

RyukTheCoder: Applied

PR #1264, opened 2025-09-28
ci: adding built widget examples

arlert-armin:
# Summary

This PR adds **example builds to the check workflow** to verify that they build correctly along with the rest of the packages.

To achieve this, I created a script that temporarily adds the `examples` to the workspace so they can access the latest changes and all sub-dependencies.

I explored other approaches (e.g. linking with `yarn`/`yalc` or embedding the built widget as a dependency), but those required linking all related sub-dependencies or caused build issues. The current method proved to be the most efficient and reliable.

Additionally:

* Installed dependencies with `skipLibCheck: true` in the `tsconfig`s to avoid build errors caused by `react-scripts` dependencies in rewired examples.
* Built all packages (including examples) to ensure everything compiles correctly.

Fixes # (issue)

---

I verified the workflow by:

* Running the script locally to ensure examples are added correctly.
* Installing dependencies and confirming builds succeed without errors.
* Running the full build process to ensure all packages (including examples) compile successfully.

---

# Checklist:

* [ ] I have performed a self-review of my code
* [ ] I have commented my code, particularly in hard-to-understand areas

### DIFF
--- a/.github/actions/build-examples/action.yml
+++ b/.github/actions/build-examples/action.yml
@@ -1,0 +1,69 @@
+name: 'Build Examples'
+description: 'Links the current widget embedded package to example projects and builds them'
+
+runs:
+  using: 'composite'
+  steps:
+    - name: 🔨 Build Widget Package
+      run: yarn run build:widget
+      shell: bash
+
+    - name: 🔗 Link Core Packages
+      run: |
+        echo "::group::🔗 Creating symbolic links for widget and React dependencies"
+        yarn link --cwd widget/embedded
+        yarn link --cwd node_modules/react
+        yarn link --cwd node_modules/react-dom
+        echo "✅ Packages linked successfully"
+        echo "::endgroup::"
+      shell: bash
+
+    - name: 📦 Setup Example Projects
+      run: |
+        echo "::group::🔍 Scanning and configuring example projects"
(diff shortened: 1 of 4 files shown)


arlert-armin:
> ```json
>   "examples/*",
>   "examples/**",
>   "examples/**/**"
> ```

I've added these two since the pattern was using the package name, and it prevented the hoisting and kept all dependencies in those two node_modules, like react-scripts or the problematic package eslint-scope, but the build failed again.
```json
      "widget-rewired/**",
      "widget-rewired-js/**"

```

nikaaru:
> > ```json
> >   "examples/*",
> >   "examples/**",
> >   "examples/**/**"
> > ```
> 
> I've added these two since the pattern was using the package name, and it prevented the hoisting and kept all dependencies in those two node_modules, like react-scripts or the problematic package eslint-scope, but the build failed again.
> 
> ```json
>       "widget-rewired/**",
>       "widget-rewired-js/**"
> ```

It's probably due to the behavior of yarn version 1.
It doesn't correctly hoist dependencies in MonoRepo.
@yeager-eren  What do you think? What approach should we take?

I believe the correct and principled approach is to use PNPM or YARN version 4; however, it's likely not within the scope of this task, and it would be better to move towards it in the near future.

PR #1268, opened 2025-09-30
feat: add solana support to metamask

arlert-armin:
# Summary

This PR adds Solana support to the MetaMask Provider by implementing the Wallet Standard. The core challenge was a mismatch between the method signatures and behavior of our previous default injected provider instance and the new Wallet Standard-based instance.

To unblock this high-priority integration, the initial implementation includes:
1.  A custom `SolanaSigner` class, as the signer methods (e.g., `signMessage`, `signTransaction`) on the Wallet Standard instance are different from those on our legacy injected provider.
2.  A temporary set of custom actions to handle Solana-specific operations through the new Wallet Standard interface.

This is a stopgap solution. The long-term plan is to deprecate this custom code once Wallet Standard support is fully integrated and standardized into the core, at which point we will use the core actions instead.

Fixes # (issue)

- [ ] Test A
- [ ] Test B

# Checklist:

### DIFF
--- a/wallets/provider-metamask/src/actions/solana.ts
+++ b/wallets/provider-metamask/src/actions/solana.ts
@@ -1,0 +1,42 @@
+import type { WalletStandardSolanaInstance } from '../types.js';
+import type { Context, FunctionWithContext } from '@rango-dev/wallets-core';
+
+import {
+  type SolanaActions,
+  utils,
+} from '@rango-dev/wallets-core/namespaces/solana';
+
+function connect(
+  getInstance: () => WalletStandardSolanaInstance
+): FunctionWithContext<SolanaActions['connect'], Context> {
+  return async () => {
+    const solanaInstance = getInstance();
+    const connectResult = await solanaInstance.features[
+      'standard:connect'
+    ].connect();
+    return utils.formatAccountsToCAIP(
+      connectResult.accounts.map((account) => account.address)
+    );
+  };
+}
+function canEagerConnect(getInstance: () => WalletStandardSolanaInstance) {
+  return async () => {
+    const solanaInstance = getInstance();
+
+    if (!solanaInstance) {
+      throw new Error(
+        'Trying to eagerly connect to your Solana wallet, but it seems that its instance is not available.'
+      );
+    }
+
+    try {
+      const result = await solanaInstance.features['standard:connect'].connect({
+        silent: true,
+      });
+      return !!result.accounts.length;
+    } catch {
+      return false;
+    }
+  };
+}
+export const solanaActions = { connect, canEagerConnect };
--- a/wallets/provider-metamask/src/builders/solana.ts
+++ b/wallets/provider-metamask/src/builders/solana.ts
@@ -1,0 +1,27 @@
+import type { WalletStandardSolanaInstance } from '../types.js';
+import type { StandardEventsChangeProperties } from '@wallet-standard/features';
+
+import { ChangeAccountSubscriberBuilder } from '@rango-dev/wallets-core/namespaces/common';
+import { utils } from '@rango-dev/wallets-core/namespaces/solana';
+
+// Hooks
+const changeAccountSubscriber = (
+  getInstance: () => WalletStandardSolanaInstance
+) =>
+  new ChangeAccountSubscriberBuilder<
+    StandardEventsChangeProperties,
+    WalletStandardSolanaInstance
+  >()
+    .getInstance(getInstance)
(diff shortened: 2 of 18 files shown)


wallets/provider-metamask/src/utils.ts
```diff
@@ -1,22 +1,35 @@
...
+} from './constants.js';
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 export type Provider = Record<string, any>;
 
 export function metamask(): Provider | null {
   const { ethereum } = window;
-
+  const solana = getSolanaWalletInstance();
   if (!ethereum || !isEthereumMetamaskProvider(ethereum)) {
```
RyukTheCoder, line 19: Can't we proceed if one of these is not available?

wallets/provider-metamask/src/utils.ts
```diff
@@ -106,6 +119,16 @@ export function evmMetamask(): EvmProviderApi {
 
   return evmInstance as EvmProviderApi;
 }
+export function solanaMetamask(): WalletStandardSolanaInstance {
+  const instances = metamask();
+  const solanaInstance = instances?.get(LegacyNetworks.SOLANA);
+  if (!solanaInstance) {
+    throw new Error(
+      'Metamask not injected or EVM not enabled. Please check your wallet.'
```
RyukTheCoder: wrong error message

wallets/provider-metamask/src/utils.ts
```diff
@@ -106,6 +119,16 @@ export function evmMetamask(): EvmProviderApi {
...
+    );
+  }
+  return solanaInstance as WalletStandardSolanaInstance;
```
RyukTheCoder, line 125: I think you can improve `Provider` type to avoid assertion.

wallets/provider-metamask/src/actions/solana.ts
```diff
@@ -0,0 +1,42 @@
...
+    );
+  };
+}
+function canEagerConnect(getInstance: () => WalletStandardSolanaInstance) {
+  return async () => {
+    const solanaInstance = getInstance();
+
+    if (!solanaInstance) {
+      throw new Error(
+        'Trying to eagerly connect to your Solana wallet, but seems its instance is not available.'
```
RyukTheCoder:
```suggestion
        'Trying to eagerly connect to your Solana wallet, but it seems that its instance is not available.'
```

wallets/provider-metamask/src/builders/solana.ts
```diff
@@ -0,0 +1,27 @@
...
+    WalletStandardSolanaInstance
+  >()
+    .getInstance(getInstance)
+    .validateEventPayload((accounts) => !!accounts.accounts?.length)
+    .format(async (_, event) =>
+      utils.formatAccountsToCAIP(
+        event.accounts!.map((account) => account.address)
+      )
+    )
+    .addEventListener((instance, callback) => {
```
RyukTheCoder, line 22: Switch account did not work for me.  Did you test it?

RyukTheCoder (changes requested):
Tnx for your contribution, just added some comments.

wallets/provider-metamask/src/utils.ts
arlert-armin: Yes, we can. I removed the Solana check.

wallets/provider-metamask/src/utils.ts
arlert-armin: Thanks. I've changed it.

wallets/provider-metamask/src/utils.ts
arlert-armin: We have some extra properties on EVM types which we've not defined yet, and if I want to define them, it would take much more time than the actual scope of the task. Can we postpone it?

wallets/provider-metamask/src/actions/solana.ts
arlert-armin: Thanks, I've changed it.

wallets/provider-metamask/src/builders/solana.ts
arlert-armin: Yes, I've tested it and it worked.

wallets/provider-metamask/src/utils.ts
arlert-armin: Since the map didn't correctly enforce the type, I couldn't remove the type assertion, but I removed the `Record<string, any>` and used a better typed provider instead, leaving a `type Provider = any` in legacy code.

PR #1289, opened 2025-11-01
feat: integrate xverse wallet

arlert-armin:
# Summary

This PR integrates **Xverse Wallet** into the Hub to enable **BTC transactions**.

Fixes # (issue)

---

The integration has been manually tested to confirm wallet connection, transaction flow, and proper handling of disconnection scenarios.

**Steps to reproduce:**

1. Connect the Xverse wallet from the Hub.
2. Perform a BTC transaction and confirm it succeeds.
3. Disconnect the wallet manually and verify the Hub handles the state gracefully.

* [x] Test wallet connection and authorization flow
* [x] Test BTC transaction execution

---

# Checklist

* [x] I have made corresponding changes to the documentation

### DIFF
--- a/wallets/provider-xverse/src/builders/utxo.ts
+++ b/wallets/provider-xverse/src/builders/utxo.ts
@@ -1,0 +1,58 @@
+import type { XVerseEvent } from '../types.js';
+
+import { ActionBuilder } from '@rango-dev/wallets-core';
+import { ChangeAccountSubscriberBuilder } from '@rango-dev/wallets-core/namespaces/common';
+import {
+  type ProviderAPI,
+  utils,
+  type UtxoActions,
+} from '@rango-dev/wallets-core/namespaces/utxo';
+
+import { XVERSE_ACCESS_DENIED_ERROR_CODE } from '../constants.js';
+
+export const changeAccountSubscriber = (getInstance: () => ProviderAPI) =>
+  new ChangeAccountSubscriberBuilder<XVerseEvent, ProviderAPI>()
+    .getInstance(getInstance)
+    /*
+     * Xverse wallet may call the `changeAccount` event with `empty` value
+     * but we shouldn't disconnect in this case.
+     */
+    .onSwitchAccount((event) => {
+      if (!event.payload?.addresses?.length) {
+        event.preventDefault();
+      }
+    })
+    .format(async (_, event) => {
+      return utils.formatAccountsToCAIP(
+        event.addresses
+          .filter((address) => address.purpose === 'payment')
+          .map((address) => address.address)
+      );
+    })
+    .addEventListener((instance, callback) => {
+      return instance.addListener('accountChange', callback);
+    })
+    .removeEventListener((_, __) => {
+      // it will be removed by the main class
+    });
+
+function canEagerConnect(getInstance: () => ProviderAPI) {
+  return new ActionBuilder<UtxoActions, 'canEagerConnect'>(
+    'canEagerConnect'
+  ).action(async () => {
+    const instance = getInstance();
+    try {
+      const addressesResponse = await instance.request('getAddresses', {
+        purposes: ['payment'],
+      });
+      if (addressesResponse.error?.code === XVERSE_ACCESS_DENIED_ERROR_CODE) {
+        return false;
+      }
+      return !!addressesResponse.result?.addresses?.length;
+    } catch {
+      return false;
+    }
+  });
+}
+
+export const utxoBuilders = { changeAccountSubscriber, canEagerConnect };
--- a/wallets/provider-xverse/src/namespaces/utxo.ts
+++ b/wallets/provider-xverse/src/namespaces/utxo.ts
@@ -1,0 +1,50 @@
+import { NamespaceBuilder } from '@rango-dev/wallets-core';
+import {
+  builders as commonBuilders,
+  standardizeAndThrowError,
+} from '@rango-dev/wallets-core/namespaces/common';
+import {
+  utils,
+  type UtxoActions,
+} from '@rango-dev/wallets-core/namespaces/utxo';
+import { builders } from '@rango-dev/wallets-core/namespaces/utxo';
+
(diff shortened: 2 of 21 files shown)


wallets/provider-xverse/src/provider.ts
```diff
@@ -0,0 +1,22 @@
...
+const buildProvider = () =>
+  new ProviderBuilder(WALLET_ID)
+    .init(function (context) {
+      const [, setState] = context.state();
+      setTimeout(() => {
+        if (unisatInstance()) {
+          setState('installed', true);
+          console.debug('[xverse] instance detected.', context);
+        }
+      }, XVERSE_INJECTION_DELAY_MS);
```
RyukTheCoder, line 17: Please add a todo comment to remove this after we solved the not initiating issue.
Also maybe we can consider a unified approach for checking instances with some internals with a delay and a max retries constant. What do you think? (in another PR)

wallets/provider-xverse/src/builders/utxo.ts
```diff
@@ -0,0 +1,26 @@
+import type { XVerseEvent } from '../types.js';
+
+import { ChangeAccountSubscriberBuilder } from '@rango-dev/wallets-core/namespaces/common';
+import {
+  type ProviderAPI,
+  utils,
+} from '@rango-dev/wallets-core/namespaces/utxo';
+
+export const changeAccountSubscriber = (getInstance: () => ProviderAPI) =>
```
arlert-armin: Actually there's a problem that in switch account it triggers two event one that contains zero addresses and one that contains actual all addresses so if I use it for disconnect it will disconnect each time user changes its account. I will put the recorded jam on the task.

wallets/provider-xverse/src/provider.ts
arlert-armin: Done.
That's one of the key points for my R&D task on hub plugin api.

wallets/provider-xverse/src/namespaces/utxo.ts
```diff
@@ -0,0 +1,52 @@
...
+  .connect()
+  .action(async function () {
+    const accountsResult = await getBitcoinAccounts();
+
+    if (accountsResult.result?.addresses?.length === 0) {
+      throw new Error("Couldn't find any address!");
+    }
+    if (accountsResult.result?.network?.bitcoin?.name !== 'Mainnet') {
+      throw new Error('Please turn off testnet mode!');
+    }
```
arlert-armin: `getBitcoinAccounts` should be only responsible for what its interface is telling having no accounts is not its concern and it should return whatever it had but in connect it would be an issue so handling it here would be more meaningful.

wallets/provider-xverse/src/signers/utxoSigner.ts
```diff
@@ -0,0 +1,56 @@
...
+    });
+
+    // 3. Sign (& broadcast)
+    const signPsbtResult = await this.provider.request('signPsbt', {
+      psbt: psbt.unsignedPsbtBase64,
+      signInputs,
+      broadcast: true,
+    });
+
+    if (signPsbtResult?.error) {
```
arlert-armin: This is the type of the response of  `sats-conenct` lib and status not exists on injected provider which is the one i'm using.

wallets/provider-xverse/src/builders/utxo.ts
RyukTheCoder, line 13: Should disconnect the wallet when you get an event with empty addresses

Image

wallets/provider-xverse/src/namespaces/utxo.ts
RyukTheCoder: Errors should be handled inside `getBitcoinAccounts`

wallets/provider-xverse/src/utils.ts
```diff
@@ -0,0 +1,53 @@
...
+  if (!bitcoinInstance) {
+    throw new Error('Xverse not injected. Please check your wallet.');
+  }
+
+  return bitcoinInstance;
+}
+export async function getBitcoinAccounts(): Promise<XVerseResponse> {
+  const instance = bitcoinXverse();
+  const requestResult = await instance.request('wallet_connect', {
+    addresses: ['payment'],
```
RyukTheCoder, line 45: You can specify network as Mainnet and remove the related error handler

wallets/provider-xverse/src/signers/utxoSigner.ts
RyukTheCoder, line 45: You can check the result and handle errors using `response.status === "success"`.
https://docs.xverse.app/sats-connect/bitcoin-methods/signpsbt

RyukTheCoder (changes requested):

wallets/provider-xverse/src/utils.ts
arlert-armin: Thanks for your suggestion, applied.

RyukTheCoder (approved):
LGTM

wallets/provider-xverse/src/types.ts
```diff
@@ -0,0 +1,37 @@
...
+import type { ProviderAPI } from '@rango-dev/wallets-core/namespaces/utxo';
+
+type XVerseAddress = {
+  address: string;
+  publicKey: string;
+  purpose: 'ordinals' | 'payment' | 'stacks' | 'starknet' | 'spark';
+  addressType: 'p2tr' | 'p2wpkh' | 'p2sh' | 'stacks' | 'starknet' | 'spark';
+  walletType: 'software' | 'hardware';
+};
+export type XVerseResponse = {
```
RyukTheCoder, line 11: Considering that you are requesting for bitcoin mainnet payment addresses, I think we do not need these extra values to be defined for each type.

arlert-armin: I've removed them

RyukTheCoder (approved):
LGTM

PR #1293, opened 2025-11-05
feat: add btc support to okx wallet

arlert-armin:
# Summary
Bitcoin support has been added to OKX wallet.

#### ⚠️ UTXO Chains

The wallet only supports bitcoin.

#### ⚠️ Disconnect

The disconnect function will not work for this wallet, as it attempts to invoke disconnect on the connect method for private key imported wallets.

#### ⚠️ Switch Account

If you switch accounts while still connected to the application, the `accountChanged` event will only emit for the last connected namespace. In result of that, the account for previously connected namespaces will not get updated and requires a disconnect and reconnect to properly connect to the desired account.
Also, when you call connect on a previously connected namespace, it doesn't return the address of the new account the user switched to. To prevent the user from being stuck in such situation, we call `disconnect` on the wallet whenever user disconnects the wallet from the provider. This results in a limitation which is explained in the `Disconnect` section.

Fixes # (issue)

- [ ] Test A
- [ ] Test B

# Checklist:

### DIFF
--- a/wallets/provider-okx/src/utils.ts
+++ b/wallets/provider-okx/src/utils.ts
@@ -1,6 +1,7 @@
-import type { Provider } from './types.js';
+import type { OkxBtcAddress, Provider } from './types.js';
 import type { ProviderAPI as EvmProviderApi } from '@rango-dev/wallets-core/namespaces/evm';
 import type { ProviderAPI as SolanaProviderApi } from '@rango-dev/wallets-core/namespaces/solana';
+import type { ProviderAPI as UtxoProviderApi } from '@rango-dev/wallets-core/namespaces/utxo';
 
 import { LegacyNetworks } from '@rango-dev/wallets-core/legacy';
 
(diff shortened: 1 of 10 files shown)


wallets/provider-okx/src/utils.ts
```diff
@@ -56,3 +59,27 @@ export function solanaOKX(): SolanaProviderApi {
...
+      'OKX Wallet not injected or Utxo not enabled. Please check your wallet.'
+    );
+  }
+
+  return bitcoinInstance as UtxoProviderApi;
+}
+export async function getBitcoinAccounts(): Promise<OkxBtcAddress> {
+  const instance = bitcoinOKX();
+  const requestResult = await instance.connect({
+    namespaces: { btc: { chains: 'btc:mainnet' } },
```
RyukTheCoder: Saw no params for `connect` in docs.
https://web3.okx.com/build/dev-docs/sdks/chains/bitcoin/provider#connect

RyukTheCoder (changes requested):

wallets/provider-okx/src/utils.ts
arlert-armin: I've saw it on the injected instance object and thought its better to use it.

RyukTheCoder: Why you thought its better? What does having it add?

RyukTheCoder (approved):
LGTM

PR #1294, opened 2025-11-08
fix: validate evm address for safepal connect

arlert-armin:
# Summary

SafePal wallet sometimes returns non-EVM addresses (e.g., Solana addresses) during connection, which incorrectly sets invalid addresses for the EVM network. This PR adds validation to check that all returned addresses are valid EVM addresses before completing the connection.

Fixes BUG-SafePal Connect Sets Solana Address for EVM

**Note:** This implementation must be kept in sync with `wallets/core/src/namespaces/evm/builders`

- [x] Tested connection with SafePal account that has only Solana address - connection is now blocked with error message
- [x] Tested connection with SafePal account that has valid EVM address - connection works normally
- [x] Verified error message displays: "Non valid EVM address returned from wallet"

# Checklist:

### DIFF
--- a/wallets/provider-safepal/src/namespaces/evm.ts
+++ b/wallets/provider-safepal/src/namespaces/evm.ts
@@ -1,23 +1,42 @@
 import type { EvmActions } from '@rango-dev/wallets-core/namespaces/evm';
 
-import { NamespaceBuilder } from '@rango-dev/wallets-core';
+import { ActionBuilder, NamespaceBuilder } from '@rango-dev/wallets-core';
 import {
   builders as commonBuilders,
+  connectAndUpdateStateForMultiNetworks,
+  intoConnecting,
+  intoConnectionFinished,
   standardizeAndThrowError,
 } from '@rango-dev/wallets-core/namespaces/common';
 import { actions, builders } from '@rango-dev/wallets-core/namespaces/evm';
 
 import { WALLET_ID } from '../constants.js';
-import { evmSafepal } from '../utils.js';
+import { evmSafepal, isValidEvmAddress } from '../utils.js';
 
 const [changeAccountSubscriber, changeAccountCleanup] = builders
   .changeAccountSubscriber(evmSafepal)
   .build();
 
-const connect = builders
-  .connect()
+/**
+ * Important: Keep this implementation in sync with wallets/core/src/namespaces/evm/builders
+ */
+const connect = new ActionBuilder<EvmActions, 'connect'>('connect')
   .action(actions.connect(evmSafepal))
   .before(changeAccountSubscriber)
+  .before(intoConnecting)
+  /*
+   * Validate all accounts are valid EVM addresses
+   * Sometimes wallets may return non-EVM addresses - this check prevents that
+   */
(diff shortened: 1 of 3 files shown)


wallets/provider-safepal/package.json
```diff
@@ -25,9 +25,10 @@
     "@rango-dev/signer-solana": "^0.44.1-next.0",
     "@rango-dev/wallets-core": "^0.51.1-next.1",
     "@rango-dev/wallets-shared": "^0.52.1-next.1",
+    "ethers": "^6.15.0",
```
RyukTheCoder: use "ethers": "^6.13.2" to avoid installing new version of ethers. `yarn.lock` should not contain any changes.

wallets/provider-safepal/src/namespaces/evm.ts
```diff
@@ -1,23 +1,42 @@
...
+const connect = new ActionBuilder<EvmActions, 'connect'>('connect')
   .action(actions.connect(evmSafepal))
   .before(changeAccountSubscriber)
+  .before(intoConnecting)
+  /*
+   * Validate all accounts are valid EVM addresses
+   * Sometimes wallets may return non-EVM addresses - this check prevents that
+   */
+  .and((_, args) => {
+    if (args.accounts.some((account: string) => !isValidEvmAddress(account))) {
```
RyukTheCoder, line 32: You can use `isEvmAddress` from '@rango-dev/wallets-shared'

wallets/provider-safepal/package.json
arlert-armin: Used existing isEvmAddress from wallet/shared and removed this dependency.

wallets/provider-safepal/src/namespaces/evm.ts
arlert-armin: Yes I've used this instead but since the wallet's address is in CAIP I had to extract the wallet address first before using that util so I've changed utils file to use this method instead of ethers one.

arlert-armin:
I will make sure to remove `package.json` and `yarn.lock` on squash before merging this.

wallets/provider-safepal/src/utils.ts
```diff
@@ -36,3 +37,28 @@ export function getInstanceOrThrow(): Provider {
...
+    return false;
+  }
+
+  // Extract hex portion (without 0x prefix)
+  const hex = walletAddress.slice(2);
+
+  // If all-lower or all-upper, treat as valid (legacy non-checksummed)
+  if (hex === hex.toLowerCase() || hex === hex.toUpperCase()) {
+    return true;
+  }
```
RyukTheCoder: You can just use `isEvmAddress` from '@rango-dev/wallets-shared'. Also please check with @yeager-eren if we can move checking for address to evm namespace of core package.

arlert-armin: Line 59 is checking if it is a legacy address or not which they didn't had any checksum. But I've removed them all.
I've checked and his opinion was negative. We have to wait until see how other providers behave to add this.

PR #1298, opened 2025-11-18
hotfix: set vercel's enviroment correctly for deploy command

yeager-eren:
# Summary

Vercel has changed its policy and don't let a deployment with `preview` environment to be promoted as prod.
This PR set the enviroment (using target flag) when we are deploying.


nikaaru (approved):
LGTM

PR #1301, opened 2025-11-20
feat: migrate keplr to the hub

arlert-armin:
# Summary

This PR introduces Keplr migration support to the Hub, including Cosmos connectivity, chain suggestion, and Account Switching.

Key changes:

* Added metadata passing during connection to support Cosmos filtering and logic previously handled in the legacy version.
* Updated `mapCaipNamespaceToLegacyNetworkName` to accept meta and correctly resolve legacy network names for connection and address selection.
* Implemented `suggestAndConnect` for the Hub adapter; updated the legacy version to accept namespaces.
* Keplr does not emit updated accounts on account change — it requires reconnecting.
  To support this:

  * Added action type to the subscriber builder so `context.action` includes the connect action.
  * Introduced a new namespace state `connectArgs` to persist and reuse original connection arguments during reconnects.

Fixes # (issue)

# Related PRs
- https://github.com/rango-exchange/rango-client/pull/1316
- https://github.com/rango-exchange/rango-client/pull/1315
- https://github.com/rango-exchange/rango-client/pull/1313
- https://github.com/rango-exchange/rango-client/pull/1312
---

* Tested connection flows
* Verified correct metadata propagation for Cosmos chains.
* Confirmed `suggestAndConnect` behavior for new namespaces.
* Simulated account switching to ensure reconnect logic uses `connectArgs` correctly.
* Ensured that address selection and network mapping work as expected for Cosmos chains.

* [ ] Test A
* [ ] Test B

---

# Checklist:

* [x] I have made corresponding changes to the documentation
* [x] I have added tests that prove my fix is effective or that my feature works
* [ ] Implemented a user interface (UI) change, referencing our Figma design to ensure pixel-perfect precision.

### DIFF
--- a/wallets/provider-keplr/package.json
+++ b/wallets/provider-keplr/package.json
@@ -7,12 +7,12 @@
   "main": "./dist/mod.js",
   "exports": {
     ".": {
-      "default": "./dist/mod.js",
-      "types": "./dist/mod.d.ts"
+      "types": "./dist/mod.d.ts",
+      "default": "./dist/mod.js"
     },
     "./lib": {
-      "default": "./dist/lib.js",
-      "types": "./dist/lib.d.ts"
+      "types": "./dist/lib.d.ts",
+      "default": "./dist/lib.js"
     }
   },
   "typings": "dist/mod.d.ts",
--- a/widget/playground/src/containers/FunctionalLayout/FunctionalLayout.Wallets.tsx
+++ b/widget/playground/src/containers/FunctionalLayout/FunctionalLayout.Wallets.tsx
@@ -110,7 +110,9 @@ export function WalletSection() {
               if (state(WalletTypes.KEPLR).connected) {
                 void disconnect(WalletTypes.KEPLR);
               } else {
-                void connect(WalletTypes.KEPLR);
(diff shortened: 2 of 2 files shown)


wallets/provider-keplr/src/mod.ts
```diff
@@ -0,0 +1,11 @@
+import { defineVersions } from '@rango-dev/wallets-core/utils';
+
+import { buildProvider } from './provider.js';
+
+export { cosmosActions as keplrCosmosActions } from './actions/cosmos.js';
+export { getCosmosAccounts as keplrGetCosmosAccounts } from './utils.js';
```
yeager-eren: let's keep the `mod.ts` same as others. i suggests create a `lib.ts` from now on, to export these kind of functionalities from providers. 

don't forget to add a new `exports` in package.json.

arlert-armin: I've added it.

yeager-eren: please use `lib.ts` instead of `libs`. also you should add the entrypoint in `build` script (`--inputs`)

arlert-armin: Done

PR #1344, opened 2025-12-22
feat: migrated tonconnect to the hub

llawliet-l-l:
# Summary

Added Ton builders, utils and types to the wallets core.

Migrated TonConnect to use the new hub APIs for ton namespaces and updated documentation:
- updated resolving namespace for ton wallets to legacy networks
- created an initialization function to be called with TonConnect configs before using TonConnect

TonConnect was tested end-to-end within the Hub environment:

- [x] Connecting to Ton Wallets
- [x] Reconnecting on page reload if previously connected
- [x] Disconnecting when user disconnects from the wallet or directly from the app
- [x] Executing transactions from Ton namespace

# Checklist:


wallets/core/src/namespaces/ton/types.ts
```diff
@@ -0,0 +1,19 @@
...
+  extends AutoImplementedActionsByRecommended,
+    CommonActions {
+  connect: (options?: ConnectOptions) => Promise<Accounts>;
+  canEagerConnect: () => Promise<boolean>;
+}
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export type ProviderAPI = Record<string, any>;
+
+export type ConnectOptions = {
```
arlert-armin: Remove it since it's not being using.

wallets/core/src/namespaces/ton/constants.ts
```diff
@@ -0,0 +1,2 @@
+export const CAIP_NAMESPACE = 'ton';
```
arlert-armin: Change it according to CAIP

wallets/provider-tonconnect/src/actions/ton.ts
```diff
@@ -0,0 +1,54 @@
...
+  };
+}
+
+export function disconnect(
+  getInstance: () => TonProviderAPI
+): FunctionWithContext<TonActions['disconnect'], Context> {
+  return async (context) => {
+    const tonInstance = getInstance();
+    await tonInstance.disconnect();
+    commonActions.disconnect(context);
```
arlert-armin, line 42: Do it separately in a and added to its action builder.

wallets/provider-tonconnect/src/namespaces/ton.ts
```diff
@@ -0,0 +1,40 @@
...
+
+const canEagerConnect = builders
+  .canEagerConnect()
+  .action(tonActions.canEagerConnect(tonConnect))
+  .build();
+
+const disconnect = new ActionBuilder<TonActions, 'disconnect'>('disconnect')
+  .action(tonActions.disconnect(tonConnect))
+  .after(changeAccountCleanup)
+  .build();
```
arlert-armin, line 31: Use disconnect from common builders

wallets/provider-tonconnect/src/builders/ton.ts
```diff
@@ -0,0 +1,37 @@
...
+    .format(async () => [])
+    .addEventListener((instance, callback) => {
+      return instance.onStatusChange(callback);
+    })
+    .removeEventListener(() => {
+      /*
+       * We handle unsubscribing in the builder and by returning an unsubscribe function from 'addEventListener'
+       */
+    });
+
```
arlert-armin: Don't use change account subscriber builder for disconnect use another disconnect handler.

wallets/provider-tonconnect/src/actions/ton.ts
```diff
@@ -0,0 +1,54 @@
...
+    return utils.formatAccountsToCAIP([userFriendlyAddress]);
+  };
+}
+
+export function disconnect(
+  getInstance: () => TonProviderAPI
+): FunctionWithContext<TonActions['disconnect'], Context> {
+  return async (context) => {
+    const tonInstance = getInstance();
+    await tonInstance.disconnect();
```
arlert-armin: Check if it's been disconnected don't disconnect again

wallets/core/src/namespaces/ton/constants.ts
llawliet-l-l: Fixed in https://github.com/rango-exchange/rango-client/pull/1344/commits/8dddf1864c2aa6a386d78302a3682a42960cec5e

wallets/core/src/namespaces/ton/types.ts
llawliet-l-l: Fixed in https://github.com/rango-exchange/rango-client/pull/1344/commits/2ccb6c14fab0f64fee546e6f623760369c947fc9

wallets/provider-tonconnect/src/actions/ton.ts
llawliet-l-l: Fixed in https://github.com/rango-exchange/rango-client/pull/1344/commits/7ee58a79b9528050383269db69c0ac4a4eda43a4

wallets/provider-tonconnect/src/builders/ton.ts
llawliet-l-l: Fixed in https://github.com/rango-exchange/rango-client/pull/1344/commits/7ee58a79b9528050383269db69c0ac4a4eda43a4

wallets/provider-tonconnect/src/actions/ton.ts
llawliet-l-l: Fixed in https://github.com/rango-exchange/rango-client/pull/1344/commits/7ee58a79b9528050383269db69c0ac4a4eda43a4

wallets/provider-tonconnect/src/tonConnect.ts
```diff
@@ -0,0 +1,43 @@
...
+  #tonModule?: typeof import('@tonconnect/ui');
+  #env?: Environments;
+  #tonConnectInstance?: TonConnectUI;
+
+  async initialize(env?: Environments) {
+    if (!env) {
+      throw new Error('Environments are not set');
+    }
+    this.#env = env;
+    if (this.#env) {
```
arlert-armin: This guard always throws — `#env` was assigned a truthy value on the line above. `dynamicImportWithRefinedError` below is never reached, the error is swallowed by the empty `catch (_)` in `provider.ts`, and every subsequent `getInstance()` throws "TonConnect instance isn't initialized…". The SDK isn't actually loaded on this branch — please retest the connect/disconnect flow end-to-end.

wallets/provider-tonconnect/src/namespaces/ton.ts
```diff
@@ -0,0 +1,39 @@
...
+  .or(disconnectCleanUp)
+  .or(standardizeAndThrowError)
+  .build();
+
+const canEagerConnect = builders
+  .canEagerConnect()
+  .action(tonActions.canEagerConnect(tonConnect.getInstance))
+  .build();
+
+const disconnect = new ActionBuilder<TonActions, 'disconnect'>('disconnect')
```
arlert-armin: Prior thread asked to use `commonBuilders.disconnect` here — this still hand-rolls `new ActionBuilder<TonActions, 'disconnect'>(...)`, and `tonActions.disconnect` mixes wallet-side disconnect with state cleanup (it calls both `tonInstance.disconnect()` and `commonActions.disconnect(context)`).

Every other hub provider uses the common builder — see `provider-metamask/src/namespaces/evm.ts:58-61` and `provider-cosmostation/src/namespaces/cosmos.ts:34`. Suggested shape:

```ts
// actions/ton.ts — drop the commonActions.disconnect call, only do wallet-side
export function disconnect(getInstance: () => TonConnectUI) {
  return async () => {
    const inst = getInstance();
    if (inst.connected) await inst.disconnect();
  };
}

// namespaces/ton.ts
const disconnect = commonBuilders
  .disconnect<TonActions>()
  .before(tonActions.disconnect(() => tonConnect.getInstance()))
  .after(disconnectCleanUp)
  .build();
```

wallets/provider-tonconnect/src/utils.ts
```diff
@@ -0,0 +1,50 @@
...
+export const tonConnect = new TonConnect();
+
+export function getInstanceOrThrow(): Provider {
+  const instance = tonConnect.getInstance();
+
+  const instances = new Map([[Networks.TON, instance]]);
+  return instances as Provider;
+}
+
+export async function waitForConnection(
```
arlert-armin: `waitForConnection` only makes sense relative to a TonConnect instance, and every other instance-touching helper already routes through the `TonConnectAdapter` class. Moving it onto the class puts all SDK-instance logic in one place and removes the parameter at the call site:

```ts
class TonConnectAdapter {
  async waitForConnection(): Promise<string> {
    const instance = this.getInstance();
    return new Promise((resolve, reject) => { /* same body */ });
  }
}

// actions/ton.ts
const result = await tonConnect.waitForConnection();
```

`utils.ts` then keeps just `getInstanceOrThrow` and the singleton.

wallets/provider-tonconnect/src/tonConnect.ts
```diff
@@ -0,0 +1,43 @@
+import type { Environments } from './types.js';
+import type { TonConnectUI } from '@tonconnect/ui';
+
+import { dynamicImportWithRefinedError } from '@rango-dev/wallets-shared';
+
+export class TonConnect {
+  // eslint-disable-next-line @typescript-eslint/consistent-type-imports
+  #tonModule?: typeof import('@tonconnect/ui');
```
arlert-armin: The `eslint-disable` here can go away by switching to a top-level type-only namespace import — fully erased at runtime, no rule fires:

```ts
import type * as TonConnectUIModule from '@tonconnect/ui';

class TonConnectAdapter {
  #tonModule?: typeof TonConnectUIModule;
  ...
}
```

The dynamic `import('@tonconnect/ui')` still does the actual loading.

wallets/provider-tonconnect/src/tonConnect.ts
```diff
@@ -0,0 +1,43 @@
...
+  #env?: Environments;
+  #tonConnectInstance?: TonConnectUI;
+
+  async initialize(env?: Environments) {
```
arlert-armin: Two cleanups for `initialize()`:
- `#env` doesn't need to be a class field — it's only read once on the line where `TonConnectUI` is constructed.
- The check that should be here is a re-init guard on `#tonConnectInstance`, not a duplicate of the env check.

Combined with the rename in the other comment, the body becomes:

```ts
async initialize(options: TonConnectOptions): Promise<void> {
  if (this.#tonConnectInstance) return;
  if (!options) throw new Error('TonConnect options are required');

  this.#tonModule = await dynamicImportWithRefinedError(
    () => import('@tonconnect/ui')
  );
  this.#tonConnectInstance = new this.#tonModule.TonConnectUI(options);
}
```

wallets/provider-tonconnect/src/types.ts
```diff
@@ -1,3 +1,16 @@
-export interface Environments extends Record<string, string | undefined> {
-  manifestUrl: string;
-}
+import type { LegacyNetworks } from '@rango-dev/wallets-core/legacy';
+import type {
+  TonConnectUI,
+  TonConnectUiOptionsWithManifest,
+} from '@tonconnect/ui';
+
+export type Environments = TonConnectUiOptionsWithManifest;
```
arlert-armin, line 7: `Environments`/`env` reads like environment variables, but the value is `TonConnectUiOptionsWithManifest` — a UI-config object (`manifestUrl`, optional UI prefs). Suggest renaming for clarity:

```ts
export type TonConnectOptions = TonConnectUiOptionsWithManifest;
```

…and use `options` everywhere (`provider.ts`, `tonConnect.ts`, `init` signature, readme). I greppped — `Environments` isn't a shared convention across other providers, so the rename is local to this package.

wallets/provider-tonconnect/src/provider.ts
```diff
@@ -0,0 +1,35 @@
...
+
+import { ProviderBuilder } from '@rango-dev/wallets-core';
+
+import { metadata, WALLET_ID } from './constants.js';
+import { ton } from './namespaces/ton.js';
+import { tonConnect } from './utils.js';
+
+const buildProvider = () =>
+  new ProviderBuilder(WALLET_ID)
+    .init(async function (context, environments: Environments) {
```
arlert-armin: Three things to clean up here:

1. The inner `initializeTon` wrapper isn't needed — the outer `init` callback is already async and the hub doesn't await it (`hub.ts:64-67`), so any rejection from the try/catch is contained without the wrapper.
2. `setState('installed', true)` is in `finally` — i.e. set even on failure. Compare to `provider-metamask/src/provider.ts:13-16`, which only flips `installed` when the SDK is actually detected. With the current code, the wallet appears installed/clickable, then connect throws.
3. The empty `catch (_)` hides genuine init failures.

Suggested shape:

```ts
.init(async function (context, options: TonConnectOptions) {
  const [, setState] = context.state();
  try {
    await tonConnect.initialize(options);
    setState('installed', true);
    console.debug('[ton] instance initialized.', context);
  } catch (err) {
    console.warn('[ton] failed to initialize TonConnect:', err);
  }
})
```

The "⚠️ Init" note in `readme.md:22-25` no longer applies and should be removed/updated.

wallets/provider-tonconnect/src/tonConnect.ts
```diff
@@ -0,0 +1,43 @@
+import type { Environments } from './types.js';
+import type { TonConnectUI } from '@tonconnect/ui';
+
+import { dynamicImportWithRefinedError } from '@rango-dev/wallets-shared';
+
+export class TonConnect {
```
arlert-armin: The class name `TonConnect` collides with the `TonConnect` symbol exported by `@tonconnect/ui`, which is part of why this file needs the inline `typeof import(...)` workaround and just makes the file harder to read. Suggest renaming the class to `TonConnectAdapter` and the file to `tonConnectAdapter.ts` — matches the role (the thing that adapts the TonConnect SDK to our hub shape) and removes the symbol clash.

wallets/provider-tonconnect/src/provider.ts
llawliet-l-l: Fixed in https://github.com/rango-exchange/rango-client/pull/1344/commits/80e716e7ad7eb0991bc026367177a2d3f9c06c0d

wallets/provider-tonconnect/src/utils.ts
llawliet-l-l: Fixed in https://github.com/rango-exchange/rango-client/pull/1344/commits/f3c9abb40e71e1c9e5daed5d3ee1fb60bb54253a

wallets/provider-tonconnect/src/namespaces/ton.ts
llawliet-l-l: Fixed in https://github.com/rango-exchange/rango-client/pull/1344/commits/07feb782a1e849be2389e822b1c21e0e38d90fef

wallets/provider-tonconnect/src/namespaces/ton.ts
llawliet-l-l: Fixed in https://github.com/rango-exchange/rango-client/pull/1344/commits/07feb782a1e849be2389e822b1c21e0e38d90fef

wallets/provider-tonconnect/src/tonConnect.ts
llawliet-l-l: Fixed in https://github.com/rango-exchange/rango-client/pull/1344/commits/67fd053489b50e3b9d225186eade8bb0ce907932

wallets/provider-tonconnect/src/tonConnect.ts
llawliet-l-l: Fixed in https://github.com/rango-exchange/rango-client/pull/1344/commits/6f0c77dfe2fd36e412a7329130ab272b084c03ca

wallets/provider-tonconnect/src/tonConnect.ts
llawliet-l-l: Fixed in https://github.com/rango-exchange/rango-client/pull/1344/commits/0a82f80aafa9b9c0c0bdd3aff803f1497e2c7f72

wallets/provider-tonconnect/src/tonConnect.ts
llawliet-l-l: Fixed in https://github.com/rango-exchange/rango-client/pull/1344/commits/67fd053489b50e3b9d225186eade8bb0ce907932

wallets/provider-tonconnect/src/tonConnectAdapter.ts
```diff
@@ -0,0 +1,74 @@
...
+    if (!this.#tonModule) {
+      throw new Error("Couldn't initialize the TonConnect module");
+    }
+    return this.#tonModule;
+  }
+
+  async waitForConnection(): Promise<string> {
+    const tonConnectUI = this.getInstance();
+    return new Promise((resolve, reject) => {
+      console.log('WATING');
```
arlert-armin: Remove it.

wallets/provider-tonconnect/src/tonConnectAdapter.ts
```diff
@@ -0,0 +1,74 @@
...
+      const unsubscribeStatusChange = tonConnectUI.onStatusChange(
+        (state) => {
+          console.log('TONCONNECTUI');
```
arlert-armin: Also this.

wallets/provider-tonconnect/src/namespaces/ton.ts
```diff
@@ -0,0 +1,43 @@
...
+  .build();
+
+const canEagerConnect = builders
+  .canEagerConnect()
+  .action(tonActions.canEagerConnect(tonConnect.getInstance.bind(tonConnect)))
+  .build();
+
+const disconnect = commonBuilders
+  .disconnect<TonActions>()
+  .before(tonHooks.beforeDisconnect)
```
arlert-armin: Hook name is too general, we have to tell what is it actually doing.

wallets/provider-tonconnect/src/tonConnectAdapter.ts
llawliet-l-l: Fixed in https://github.com/rango-exchange/rango-client/pull/1344/commits/6d31abe65a23d8ee63f8ce07094a7e58c54502af

wallets/provider-tonconnect/src/tonConnectAdapter.ts
llawliet-l-l: Fixed in https://github.com/rango-exchange/rango-client/pull/1344/commits/6d31abe65a23d8ee63f8ce07094a7e58c54502af

arlert-armin (approved):
LGTM

PR #1345, opened 2025-12-22
fix: address getting for cosmos chains with null info

arlert-armin:
# Summary

This PR fixes an issue where users were unable to generate addresses for certain Cosmos chains (e.g. Rune) when using wallets like Cosmostation, Keplr, and Leap.

The problem was caused by the experimental-chain detection logic. We were first checking for the existence of `chain.info`, and then reading `info.experimental`. For some chains, `info` can be `null`, which caused them to be incorrectly treated as experimental and blocked from address generation.

This change updates the logic to **assume chains with `info === null` are non-experimental**, and only treat a chain as experimental when `info.experimental === true`.

Fixes # (issue)

---

* Verified address generation for chains with `info === null` (e.g. Rune)
* Confirmed existing behavior remains unchanged for chains explicitly marked as experimental
* Tested wallet integrations with Cosmostation, Keplr, and Leap to ensure addresses are returned correctly

# Checklist:

* [ ] I have commented my code, particularly in hard-to-understand areas

---

If you want, I can also:

* Make this **more formal or more concise**
* Align it with your team’s usual PR tone
* Add a **small code snippet** to the summary for extra clarity

### DIFF
--- a/wallets/react/src/hub/useHubAdapter.ts
+++ b/wallets/react/src/hub/useHubAdapter.ts
@@ -265,7 +265,7 @@ export function useHubAdapter(params: UseAdapterParams): ProviderContext {
             connectNamespacePromise = async () => {
               return namespace.connect({
                 chainIds: cosmosBlockChains
-                  .filter((chain) => chain.info && !chain.info.experimental)
+                  .filter((chain) => !chain.info || !chain.info.experimental)
                   ?.map((chain) => chain.chainId!),
                 customChainIds: cosmosBlockChains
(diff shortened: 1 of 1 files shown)


arlert-armin:
Closing this PR as obsolete. It fixed a Cosmos-specific issue (experimental-chain detection for chains like Rune via Keplr/Leap/Cosmostation), but Cosmos blockchain support has since been removed (#1405) and the experimental-chain filter this patched no longer exists in useHubAdapter.

PR #1360, opened 2026-02-09
fix: use the exact balance instead of the rounded value when the user clicks max balance

Ikari-Shinji-re:
# Summary

When user wants to swap full token balance, if the number of tokens is above a threshold and user clicks on max balance, we allow him to swap rounded balance not max balance.

If user clicks on Max button, max balance should be selected. Not rounded down balance.

Fixes # (issue)

Fixed by using the actual token balance amount instead of the formatted value shown to the user.

You can compare the production deployment with the current pull request deployment.
Choose a source token that has a large balance in your wallet, for example 100,000,000.

# Checklist:


schrodingers-fox (approved):
LGTM

PR #1361, opened 2026-02-09
fix: allow users to enter text in the swap input and sanitize it into a valid number

Ikari-Shinji-re:
# Summary

Pasting amounts with thousand separators (e.g., 12,500) into the widget’s amount input is not supported, requiring users to manually remove the commas first.

Fixes # (issue)

Previously, the swap amount input used the number type. To allow users to paste token amounts with thousand separators, we needed to switch the input to text, and I’ve made that change.

We also need a way to parse the raw text input into a valid number. For this, I added a function called `parseNumericValue`, which takes the raw input string and converts it into a valid numeric value. This function:

- strips letters, spaces, commas, and symbols

- preserves digits (0–9)

- keeps only the first . as the decimal separator

We already have existing sanitizers for user input (such as removing leading or trailing zeros). These can remain unchanged; we only use parseNumericValue to convert the raw text into a valid number that the existing sanitizers can then process.

Additionally, I updated the `sanitizeInputAmount` function (triggered on input blur) so that if the value ends with a decimal separator and has no decimal part, the separator is removed.

Try entering both text and numbers. Only digits and a single decimal separator should be accepted in the input.

Paste different combinations of characters and numbers—the value displayed in the input should always be a valid number. For example:

- "300,222" → "300222"

- "12a3.4b5" → "123.45"

- "1.2.3.4" → "1.234"

- ".5" → ".5"

# Checklist:

### DIFF
--- a/widget/embedded/src/utils/sanitizers.ts
+++ b/widget/embedded/src/utils/sanitizers.ts
@@ -39,3 +39,29 @@ export function stripTrailingZeros(input: string): string {
   const s = input.replace(/(\.\d*?[1-9])0+$/, '$1');
   return s.replace(/\.0+$/, '');
 }
+
+/**
+ * Normalize a numeric input string by removing all non-numeric characters
+ * except digits and a single decimal separator.
+ *
+ * This function:
+ * - strips letters, spaces, commas, and symbols
+ * - preserves digits (`0–9`)
+ * - keeps only the first `.` as the decimal point
+ *
+ * @example "300,222"     → "300222"
+ * @example "12a3.4b5"   → "123.45"
+ * @example "1.2.3.4"    → "1.234"
+ * @example ".5"         → ".5"
+ */
+export function parseNumericValue(value: string): string {
+  value = value
+    // 1. Remove everything except digits and dots
+    .replace(/[^\d.]/g, '')
+    // 2. Keep only the first dot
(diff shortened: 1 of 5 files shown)


widget/embedded/src/utils/sanitizers.ts
```diff
@@ -39,3 +39,33 @@ export function stripTrailingZeros(input: string): string {
...
+  value = value.replace(/[^\d.]/g, '');
+
+  // keep only first dot
+  const firstDot = value.indexOf('.');
+
+  if (firstDot !== -1) {
+    value =
+      value.slice(0, firstDot + 1) +
+      value.slice(firstDot + 1).replace(/\./g, '');
+  }
```
schrodingers-fox: I believe the entire function can be managed using regex. I experimented with the following code and got the same result:
```tsx
value.replace(/[^\d.]/g, "") // 1. Remove everything except digits and dots
  .replace(/\./g, (match, offset, string) => 
    string.indexOf('.') === offset ? '.' : ''
  ); // 2. Keep only the first dot
```


this approach effectively filters out any non-digit and non-period characters, while also ensuring that only the first dot is kept in the string.

Ikari-Shinji-re: Thanks, I’ve made the change based on your suggestion.

schrodingers-fox (approved):
Approved

PR #1362, opened 2026-02-23
chore: remove fallback api keys and make it mandatory for clients to pass

yeager-eren:
# Summary

Remove fallback keys and api key should be set.

Hardcoding api key is not a good practice even if the key had a good amount restriction to use. In this PR all the keys have been removed and force to clients to set the key by themselves or they face with a runtime error.

# Technical Details

I've created 3 github environments: `preview`, `next`, `production` and used it in `checks` (preivew), `prerelase` (next), `production relase` (prod) workflows. the point is these vars should be injected when we are building the clients.
 
For each environment, `PLAYGROUND_RANGO_API_KEY` and `RANGO_API_KEY` has been set as `vars`.

Parcel reading `.env` files from the root of monerepo, not happy with that but it's ok for now. each client should have its own .env file in its directory.

I also not happy with copying `env.ts` all over the codebase, but we need a shared package for these kind of utils which we don't have atm. we can fix it in future as a techincal debt.

Resources:
- https://docs.github.com/en/actions/reference/workflows-and-actions/deployments-and-environments#environment-secrets
- https://docs.github.com/en/actions/how-tos/deploy/configure-and-manage-deployments/manage-environments#creating-an-environment
- https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/deploy-to-environment

- [ ] Test A
- [ ] Test B

# Checklist:


nikaaru (approved):
LGTM

PR #1373, opened 2026-04-25
fix: update cross chain provider balances probelm

arlert-armin:
**Description:**
Previously, `fetchMainTokenBalance` was only called when a wallet was connected, so handling a single wallet type was sufficient.

However, with the introduction of refetch logic that also relies on this action, scenarios with multiple wallet types are now possible. This caused an issue where some wallets were not being updated properly during refetch.

This PR updates the logic to ensure that token balances are fetched and refreshed correctly for all connected wallet types.

### DIFF
--- a/widget/embedded/src/store/slices/wallets.ts
+++ b/widget/embedded/src/store/slices/wallets.ts
@@ -466,7 +466,7 @@ export const createWalletsSlice = keepLastUpdated<AppStoreState, WalletsSlice>(
       const { _balances, _aggregatedBalances } =
         computeNextStateAfterWalletBalanceRemoval(
           partialCurrentState,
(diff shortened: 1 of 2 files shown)


widget/embedded/src/store/slices/wallets.ts
```diff
@@ -630,6 +633,7 @@ export const createWalletsSlice = keepLastUpdated<AppStoreState, WalletsSlice>(
         return;
       }
       const walletsDetails = response.wallets;
+      console.log(walletsDetails, 'wallet details');
```
Ikari-Shinji-re: Please remove this

Ikari-Shinji-re (approved):

PR #1384, opened 2026-05-05
fix(ci): add workspaces list hash to prepare action cache key

RyukTheCoder:
# Summary

Added hash of the list of workspaces to the cache key in prepare action so if a local dependency gets added to the dependency list and `yarn.lock` does not change, still a cache hit will not happen and it will run the `yarn install` so symlink of local packages gets updated in `node_modules`.

# Checklist:

### DIFF
--- a/scripts/workspace-keys-hash/command.mjs
+++ b/scripts/workspace-keys-hash/command.mjs
@@ -1,0 +1,42 @@
+import { createHash } from 'node:crypto';
+import { exec as execCallback } from 'node:child_process';
+import { promisify } from 'node:util';
+
+const execCallbackAsync = promisify(execCallback);
+
+function getWorkspaceKeysHash(stringifiedWorkspaceInfo) {
+  let workspaceInfo;
+  try {
+    workspaceInfo = JSON.parse(stringifiedWorkspaceInfo);
+  } catch (error) {
+    throw new Error(
+      'Unable to parse output from yarn workspaces -s info --json',
+      { cause: stringifiedWorkspaceInfo }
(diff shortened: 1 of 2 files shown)


scripts/workspace-keys-hash/command.mjs
```diff
@@ -0,0 +1,33 @@
...
+
+  const jsonPayload = rawOutput.slice(jsonStartIndex, jsonEndIndex + 1);
+  const workspaceInfo = JSON.parse(jsonPayload);
+  const workspaceKeys = Object.keys(workspaceInfo).sort().join('|');
+
+  return createHash('sha256').update(workspaceKeys).digest('hex');
+}
+
+async function run() {
+  const { stdout } = await $`yarn workspaces info --json`;
```
yeager-eren: use `yarn workspaces -s info --json`, then you can pass the output directly to JSON.Parse.

Don't forget to catch `parse` error and rethrow with a meaningful error.

yeager-eren: nit: actually `--json` is following [JSON Line](https://jsonlines.org/) protocol. which means each new line will be a json. you can ignore this as I don't think `info` always return one object as output. an improvement can be check line, if it more than one line, throw an error. don't want you to put time on this. you can completely ignore.

RyukTheCoder: As I see by checking the result of command, it returns a multi-line pretty JSON, not single-line JSONL.

yeager-eren (approved):

PR #1390, opened 2026-05-10
chore: eliminate long dependency where possible

arlert-armin:
## Summary

Reduce the surface area of the [`long`](https://www.npmjs.com/package/long) dependency in this monorepo, drop unused cosmos plumbing from the WalletConnect provider, and bump several cosmos-related packages that had drifted significantly behind upstream.
### 1. `signer-cosmos`: drop `long` and the no-op gas-string round-trip

- `signers/signer-cosmos/src/helpers.ts` — removed `import Long from 'long'` and replaced `Long.fromString(fee?.gas).toString()` with `fee?.gas` directly.
- `signers/signer-cosmos/package.json` — removed `\"long\": \"^4.0.0\"` from dependencies.

**Why:** The previous `Long.fromString(s).toString()` round-trip on the gas string was effectively a no-op for any valid gas value, and *worse* than CosmJS's own validation for invalid input — `Long.fromString` silently accepts garbage, while `@cosmjs/stargate`'s `signAndBroadcast` already strictly validates the gas string via `Int53.fromString` (which uses `/^-?[0-9]+$/` and throws clearly on bad input or out-of-range values). So passing the string through directly is both safer (clearer errors) and lets us drop the dependency. The only behavioural loss is the silent `\"NaN\"`/`\"Infinity\"` → `\"0\"` coercion that `Long.fromString` did, which is not a desired behaviour for a gas value.

### 2. `signer-cosmos`: bump `@cosmjs/stargate` `^0.31.0` → `^0.34.0`

**Why:** This package had not been updated for a long time. Bumping it removes one transitive `long` source from the tree, brings in upstream fixes, and unblocks future cosmos updates that have been blocked by the older version pin.

### 3. Providers: pin `@rango-dev/signer-cosmos` to `^0.39.2-next.0`

`provider-cosmostation`, `provider-keplr`, `provider-leap-cosmos`, `provider-xdefi` all had `^0.39.1`; bumped to `^0.39.2-next.0`.

**Why:** The local workspace `signer-cosmos` is at `0.39.2-next.0` (a pre-release). Per semver, `^0.39.1` does *not* match a pre-release version, so yarn workspaces was falling back to fetching the published `0.39.1` from the npm registry — meaning the local long-removal would not actually take effect for downstream providers. Bumping the range to `^0.39.2-next.0` lets yarn link to the workspace package. **This range will be tightened back to `^0.39.2` once `signer-cosmos@0.39.2` is published.**

### 4. `provider-walletconnect-2`: drop unused cosmos signer

- Removed deps: `@cosmjs/launchpad`, `@keplr-wallet/cosmos`, `@keplr-wallet/simple-fetch`, `@rango-dev/signer-cosmos`, `cosmos-wallet`.
- Deleted files: `src/signers/cosmos.ts`, `src/signers/helper.ts`, `src/signers/mock.ts`.
- `src/signer.ts` — removed COSMOS signer registration and dynamic import.

**Why:** WalletConnect doesn't currently support cosmos in this provider, so the cosmos signer code and its dependencies were dead weight. Removing them eliminates several transitive `long` (and other) dependencies and cuts ~480 lines of unused code.

> **Possible follow-up:** the WalletConnect provider also pulls in solana dependencies that may be similarly unused; worth auditing in a separate PR.

### 5. `provider-keplr`: bump keplr/chain-registry packages

- `@chain-registry/keplr` `^2.0.110` → `^2.0.198`
- `@keplr-wallet/types` `^0.11.21` → `^0.13.26` (was a major-version drift)
- Added `starknet ^8` to devDependencies — required as a peer of `@keplr-wallet/types@^0.13.26`.

**Why:** Same drift situation as cosmos packages — these had not been updated in a long time. Bringing them current trims one transitive `long` source and aligns with what `signer-cosmos` is now using.

## What is *not* fixed in this PR

`yarn why long` still reports the package as present transitively via:

- `@keplr-wallet/types` and `@keplr-wallet/proto-types` — no stable upstream release drops `long`.
- `cosmjs-types` — the stable release still depends on `long`.
- `protobufjs` — still pulls `long` for 64-bit integer support.
- `@trezor/transport` (via `provider-trezor`) — out of scope for this branch.

These would each require either an upstream release or a more invasive migration (e.g. swapping `cosmjs-types` for the BigInt fork), so they are intentionally deferred.

`yarn audit` currently reports **159 Low / 408 Moderate / 386 High / 53 Critical** vulnerabilities. These are pre-existing across the workspace and not addressed here; calling them out for visibility.

## How did you test this change?

- [ ] `yarn install` resolves cleanly with the new ranges
- [ ] `yarn build` for `signer-cosmos`, `provider-keplr`, `provider-cosmostation`, `provider-leap-cosmos`, `provider-walletconnect-2`, `provider-xdefi`
- [ ] Smoke-test a Cosmos transaction end-to-end via Keplr (gas string passes through `signAndBroadcast` correctly)
- [ ] Smoke-test WalletConnect EVM + Solana flows still work after the cosmos signer removal

## Checklist


nikaaru (approved):
Thanks, LGTM

PR #1392, opened 2026-05-11
fix: display the response error message instead of the axios error

Ikari-Shinji-re:
# Summary

When a user tries to make a swap and the server cannot create the transaction, the API returns an HTTP error status along with an error message in the response body. Previously, we incorrectly displayed the HTTP library (Axios) error instead of the server-provided error message.

Fixes # (issue)

* Updated the `prettifyErrorMessage` helper to check whether the error is an Axios error.
* If the error is an `Axios` error, we now return the `error` field from the response object instead of the default Axios error message.

Override the `/tx/create` response with the following:

status code: `400`

```json
{
  "ok": false,
  "error": "failed for test",
  "errorCode": 1234,
  "traceId": 123456,
  "transaction": null
}
```

Steps:

1. Select a route for a swap.
2. Start the swap flow.

Expected result:
The displayed error message should be `"failed for test"` from the response object, instead of the `Axios` error mess

You can also verify this by trying this route on XO Swap:
`/?fromBlockchain=ARBITRUM&fromToken=USDT--0xfd086bc7cd5c481dcc9c85ebe478a1c0b69fcbb9&toBlockchain=TRON&toToken=USDT--TR7NHqjeKQxGTCi8q8ZY4pL8otSzgjLj6t&fromAmount=10`

- [ ] Test A
- [ ] Test B

# Checklist:

### DIFF
--- a/queue-manager/rango-preset/src/shared-errors.ts
+++ b/queue-manager/rango-preset/src/shared-errors.ts
@@ -2,12 +2,14 @@ import type {
   APIErrorCode,
   SignerErrorCode as SignerErrorCodeType,
 } from 'rango-types';
+
 import {
-  SignerErrorCode,
-  SignerError,
-  isSignerErrorCode,
   isAPIErrorCode,
+  isSignerErrorCode,
+  SignerError,
+  SignerErrorCode,
 } from 'rango-types';
+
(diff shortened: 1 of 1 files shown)


nikaaru (approved):
LGTM

PR #1396, opened 2026-05-16
fix: use `tokenUrl` from blockchain metadata for token explorer links with `addressUrl` fallback

Ikari-Shinji-re:
# Summary

We display explorer links in the token selector and custom token pages.

Currently, the widget uses `addressUrl` from blockchain metadata, which causes incorrect explorer pages to open for some chains, including:

- Sui
- Stellar
- TON
- Solana

The widget should use `tokenUrl` instead of `addressUrl` when opening token explorer pages, with `addressUrl` as a fallback if `tokenUrl` is unavailable.

Fixes # (issue)

Go to the token selector page or the add custom token page in settings, and verify that explorer links for tokens on the listed blockchains open the correct pages.

- [ ] Test A
- [ ] Test B

# Checklist:


widget/embedded/src/components/CustomTokenModal/CustomTokenModal.tsx
```diff
@@ -29,7 +29,12 @@ const MAX_SYMBOL_LENGTH_THRESHOLD = 9;
 export function CustomTokenModal(props: PropTypes) {
   const { open, onClose, token, onExit, onSubmitClick, blockchain } = props;
 
-  const explorerLink = generateExplorerLink(token.address, blockchain);
+  const explorerLink = token.address
```
schrodingers-fox, line 32: Since `token.address` is of type `string | null`, I think using a nullish operator could be better. 
Although the `explorerLink`'s type is set to `any` because we didn't specify the return type of `createTokenExplorerAddress‍`.

widget/embedded/src/components/TokenList/TokenList.tsx
```diff
@@ -318,19 +318,18 @@ export function TokenList(props: PropTypes) {
...
                         customCssForTagTitle,
                         name: token.name,
-                        url: blockchain.info.addressUrl
-                          .split('{wallet}')
-                          .join(address),
+                        url:
+                          createTokenExplorerAddress({
+                            contractAddress: address,
+                            blockchainMeta: blockchain,
+                          }) ?? '',
```
schrodingers-fox: When we insert an empty string as a URL (''), maybe `renderDesc` then renders `<StyledLink href="">` - clicking the external-link icon navigates to the current page. Previously, no link was rendered at all. to fix it, you can guard `renderDesc` so the `<StyledLink>` is only rendered when url is truthy, or restore the conditional at the call site:
```
const url = createTokenExplorerAddress(...); 
.
.
.
if (!url)  {
   // render desc without link
}
```

widget/embedded/src/utils/meta.ts
```diff
@@ -70,3 +70,21 @@ export function isValidTokenAddress(
   const valid = regex.filter((r) => new RegExp(r).test(address)).length > 0;
   return valid;
 }
+
+export function createTokenExplorerAddress(params: {
```
schrodingers-fox: Due to the body, I think this function returns a URL string. As a suggestion, you can rename it to `createTokenExplorerUrl` or `getTokenExplorerLink`, they would be more readable.

widget/embedded/src/utils/meta.ts
```diff
@@ -70,3 +70,21 @@ export function isValidTokenAddress(
...
+  contractAddress: string;
+  blockchainMeta: BlockchainMeta;
+}) {
```
schrodingers-fox: Although TypeScript infers the type of this function as `any`, we know it would return `string | undefined`; declaring it makes optional handling at call sites explicit and prevents future regressions like what I mentioned above.

widget/embedded/src/components/TokenList/TokenList.tsx
Ikari-Shinji-re: fixed

widget/embedded/src/utils/meta.ts
Ikari-Shinji-re: fixed

widget/embedded/src/components/CustomTokenModal/CustomTokenModal.tsx
Ikari-Shinji-re: TypeScript infers the return type of `createTokenExplorerAddress‍` as `string | undefined`, and I think the truthy check is fine in practice.

widget/embedded/src/utils/meta.ts
Ikari-Shinji-re: TypeScript infers the function’s return type as `string | undefined`. I explicitly declared the return type to make future refactoring of the implementation easier. I also think we should enforce explicit function return types through our ESLint rules.

schrodingers-fox: yeah I agree to enforce it in eslint

schrodingers-fox (approved):
LGTM

PR #1406, opened 2026-05-31
ci(publish): upgrade dependents before Nx build and rollback on failure

arlert-armin:
# Summary

After upgrading Nx to the latest version, the publish pipeline started failing at the build step. The root cause was a ordering bug: `upgradeDependents` was called inside `publishTask` — i.e. **after** the Nx build had already finished. Nx was resolving dependency versions from `package.json` files that still referenced the old versions, causing it to fail to find the freshly-bumped packages.

This PR fixes the ordering and adds a rollback mechanism for dependents when a publish fails.

Fixes # (issue)

# Checklist:


nikaaru (approved):
LGTM

PR #1407, opened 2026-06-01
fix: fixed ton signer data

llawliet-l-l:
# Summary

Tonkeeper wallet checks the schema of the transaction we are passing to it before running it. The transaction should have raw address format and shouldn't contain null values

Fixes # (issue)
Fixed siging ton transacitons

- [x] Tested with both Tonkeeper and MyTonWallet wallets in localhost

# Checklist:

### DIFF
--- a/wallets/provider-tonconnect/src/signers/ton.ts
+++ b/wallets/provider-tonconnect/src/signers/ton.ts
@@ -1,7 +1,7 @@
 import type { TonConnectUI } from '@tonconnect/ui';
 import type { GenericSigner, TonTransaction } from 'rango-types';
 
-import { Cell } from '@ton/core';
+import { Address, Cell } from '@ton/core';
 import { CHAIN } from '@tonconnect/ui';
 import { SignerError } from 'rango-types';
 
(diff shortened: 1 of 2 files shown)


wallets/provider-tonconnect/src/signers/ton.ts
```diff
@@ -17,9 +17,15 @@ export class CustomTonSigner implements GenericSigner<TonTransaction> {
   }
 
   async signAndSendTx(tx: TonTransaction): Promise<{ hash: string }> {
```
arlert-armin, line 19: Now since they are exact the same can we use the one from singers instead of defining our own?

llawliet-l-l: We only use the signer in the TonConnect provider, and as far as I can tell, we don't use it anywhere else — I only changed it there, so if it was being used somewhere I'm not aware of, it wouldn't break anything.
I think if we want to use this signer in our provider, we should handle that in a separate task. This one should only fix the existing ones.

arlert-armin (approved):
LGTM

PR #1411, opened 2026-06-08
Feat/rc 240 migrate provider dafault to hub3

llawliet-l-l:
## Migrate provider-default (window.ethereum fallback) to hub3

  ### Summary

  Migrates `provider-default` — the catch-all that handles `window.ethereum`
  injection when no known wallet is detected — to **hub3**. This is the fallback
  path for unrecognized injected EVM providers, so the goal is to bring it onto
  the new architecture while keeping the fallback behavior stable.

  ### Changes

  **`provider-default` restructured to the hub3 layout:**

  | File | Purpose |
  |------|---------|
  | `mod.ts` | Entry point — exposes `versions()` with a single `1.0.0` hub provider |
  | `provider.ts` | `buildProvider()` with `ProviderBuilder`, EVM namespace, deferred install detection |
  | `constants.ts` | `WALLET_ID`, `metadata` (single EVM namespace + signers), `INJECTION_DELAY` |
  | `namespaces/evm.ts` | connect, disconnect, canEagerConnect, canSwitchNetwork, getChainId |
  | `utils.ts` | `defaultInjected()` (Map-based instance) / `evmDefault()` |
  | `types.ts` | `Provider` type for the instance map |
  | `signer.ts` | Reworked to resolve the EVM instance via `getNetworkInstance` |

  - **No legacy version.** The provider ships only the `1.0.0` hub version; the
    old `index.ts` / `helpers.ts` were removed.
  - **chainId handling.** The legacy `connect` manually called `getEvmAccounts`
    to obtain `chainId` (because the locked-wallet `check` phase returned a null
    network). The hub `actions.connect` already requests `eth_requestAccounts` and
    `eth_chainId` together, so that workaround is gone.
  - **Async injection.** Install detection is wrapped in a `setTimeout`
    (`INJECTION_DELAY`) since extensions may inject `window.ethereum` after init —
    important for the fallback, which accepts any injected EVM provider.

  **`package.json`**
  - `source` / `main` / `typings` / build inputs now point to `mod.ts`.
  - Added `@hub3js/core`, `@hub3js/std`, `@rango-dev/wallets-core`.

  **`provider-all`**
  - Imports `{ versions as defaultInjected }` and registers it directly, replacing
    the `lazyProvider(legacyProviderImportsToVersionsInterface(...))` wrapper.

  ### Testing

  - [x] `provider-default` builds successfully
  - [x] `tsc --noEmit -p tsconfig.build.json` passes with no errors
  - [x] `provider-all` builds successfully
  - [x] functionality was tested with Coin98 and Trustwallet

# Checklist:


wallets/provider-all/src/index.ts
```diff
@@ -82,9 +82,8 @@ export const allProviders = (
     }
   }
 
-  return [
+  console.log([
```
arlert-armin: Don't forget to remove this later.

wallets/provider-default/src/types.ts
```diff
@@ -0,0 +1,2 @@
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export type Provider = Record<string, any>;
```
arlert-armin: Make type like tomo: https://github.com/rango-exchange/rango-client/blob/f10ee1923fa2ccce3fca5a8044b3be654c5e2396/wallets/provider-tomo/src/types.ts

arlert-armin (changes requested):

wallets/provider-default/src/types.ts
llawliet-l-l: Fixed in https://github.com/rango-exchange/rango-client/pull/1411/commits/7b650ec1b0020610fac015e2e2cc2220e810a456

arlert-armin (approved):

PR #1412, opened 2026-06-08
feat(wallets): migrate provider-safe to hub3

llawliet-l-l:
Migrates the Safe{Wallet} provider from the legacy wallet architecture to the hub3 namespace/provider model, wires it into `provider-all` as a native hub provider, and fixes the regressions that surfaced from the migration (Safe not appearing in the wallet-selection modal).
  
  ### What changed

  #### `@rango-dev/provider-safe` — rewritten on hub3
  - Replaced the legacy single-file provider (`src/index.ts`, `src/helpers.ts`) with the hub structure:
    - `mod.ts` — package entrypoint / `versions` export
    - `provider.ts` — hub `ProviderBuilder` wiring
    - `namespaces/evm.ts` — EVM `NamespaceBuilder` with `connect` / `disconnect` / `canEagerConnect` / `canSwitchNetwork` / `getChainId` actions
    - `signers/evm.ts` — Safe transaction-service signer (unchanged behavior, now typed against `ProviderAPI` instead of `any`)
    - `utils.ts` — Safe SDK/provider bootstrap and account reading
    - `constants.ts` — wallet id / metadata
  - Safe-specific connect flow: Safe runs as a Safe App inside the Safe{Wallet} iframe, so the account is already available without a prompt. `getSafeAccounts` reads `eth_accounts` + `eth_chainId` directly instead of `eth_requestAccounts` (which Safe's provider does not implement).
  - Async provider detection: `initSafe()` builds and caches the `SafeAppProvider` by racing `sdk.safe.getInfo()` against a short timeout, so detection fails fast when the dApp is not embedded in Safe.
  - Dependencies: added `@hub3js/core` and `@hub3js/std`; bumped `@safe-global/safe-apps-provider` to `^0.18.6`. Package entrypoints moved from `index` to `mod`.

  #### `@rango-dev/provider-all`
  - Safe is now registered as a native hub provider (`versions` export) instead of being wrapped with `legacyProviderImportsToVersionsInterface`.

#### `wallets-react` — hub→legacy account mapping for contract wallets
  - In the hub→legacy event mapping (`hub/utils.ts`, `namespace_connected` / `namespace_account_switched` → `ACCOUNTS`), contract wallets now format accounts differently from regular wallets. For a contract wallet we map each CAIP account to **only its enabled blockchain name** (via
  `getBlockChainNameFromId` on the account's `chainId.reference`), instead of emitting a plain EVM address.
  - Why: regular EVM wallets fan the single address out across every supported EVM chain (see `prepareAccountsForWalletStore` → `evmChainsSupportedByWallet.forEach(... addAccount ...)` in `widget/embedded/src/utils/wallets.ts`). A contract wallet like Safe only exists on the chain it's deployed on, so it
  must be added to that one enabled chain — not spread across all EVM chains.

  #### `widget-embedded` — fix Safe missing from `ConfirmWalletsModal`
  - After the migration, hub writes the raw **chain id** into `state.network`, whereas legacy converted it to a rango blockchain **name** centrally inside `Wallet.connect()` before storing it. The contract-wallet filter in `mapWalletTypesToWalletInfo` compares `network` against a blockchain name (`chain`), 
  so the comparison silently failed and Safe was filtered out.
  - The filter now reads the connected namespaces' networks and normalizes each chain id to a blockchain name via `getBlockChainNameFromId` before comparing. `getBlockChainNameFromId` returns the value unchanged when it's already a name, so this stays correct for legacy wallets too.

  ### Notes / follow-ups
  - `readme.md`: documented that `changeAccountSubscriber` is wired up but not exercised in practice — switching the active account in Safe{Wallet} reloads the Safe App iframe, so the dApp reinitializes from scratch (reading the new account via `eth_accounts`) rather than reacting to an in-place
  account-change event.

  ### Testing
  - For testing it, we should add the app link from the **app-v2**  [PR](https://github.com/rango-exchange/app-v2/pull/363) and register that link under the custom apps in your Safe account: **Safe account → Apps → My custom apps**. Then open the custom app and run the test cases above (connecting, disconnecting, sending a transaction, …) from inside the Safe{Wallet} iframe.

# Checklist:


wallets/provider-safe/src/namespaces/evm.ts
```diff
@@ -0,0 +1,58 @@
...
+const connect = builders
+  .connect()
+  .action(actions.connect(evmSafe, { getAccounts: getSafeAccounts }))
+  .before(changeAccountSubscriber)
+  .before(intoConnecting)
+  .and(connectAndUpdateStateForMultiNetworks)
+  .after(intoConnectionFinished)
+  .or(changeAccountCleanup)
+  .or(standardizeAndThrowError)
+  .build();
```
RyukTheCoder, line 20: some operators are redundant as they are attached in @hub3js/evm/src/builders.ts

wallets/provider-safe/src/utils.ts
```diff
@@ -0,0 +1,73 @@
...
+}
+export const sdk = new SDK(options);
+
+/*
+ * Safe runs as a Safe App inside the Safe{Wallet} iframe, so the provider can
+ * only be created asynchronously once the SDK resolves the Safe info. We cache
+ * the instance here so the hub actions (connect, getChainId, signers, …) can
+ * retrieve it synchronously via `evmSafe()` after detection has completed.
+ */
+let safeProvider: SafeAppProvider | null = null;
```
RyukTheCoder: NIT: maybe its better to not keep provider and sdk in utils file. In some providers, such things are kept in a dedicated file with their getters.

widget/embedded/src/utils/wallets.ts
```diff
@@ -74,9 +76,39 @@ export function mapWalletTypesToWalletInfo(
...
-      const { installed, network } = getState(wallet);
+      const { installed, namespaces } = getState(wallet);
+      const networks = [...(namespaces?.values() || [])]
+        .filter(
+          (namespace): namespace is NamespaceData & { network: string } =>
+            namespace.connected && !!namespace.network
+        )
+        .map((namespace) => namespace.network);
+
+      /*
```
RyukTheCoder: Too Long To Read. You can improve the code by adding named helpers. I think it's better to read a well defined flow instead of reading a long comment for one strange line of code.

RyukTheCoder (changes requested):
Tnx, just added some comments. Also do not forget to update wallets/readme.md for safe wallet and explain the limitations in the provider readme file.

widget/embedded/src/utils/wallets.ts
llawliet-l-l: Fixed in https://github.com/rango-exchange/rango-client/pull/1412/commits/9c015b6a490661e763f11780c134304f51a98337

wallets/provider-safe/src/utils.ts
llawliet-l-l: Fixed in https://github.com/rango-exchange/rango-client/pull/1412/commits/8bb77dabda691ff672b8f80ecf948ada9eb7380a

wallets/provider-safe/src/namespaces/evm.ts
llawliet-l-l: Fixed in https://github.com/rango-exchange/rango-client/pull/1412/commits/4b53d9f49e464ce6e1bf2cc29ed63991812ac38e

wallets/provider-safe/src/provider.ts
```diff
@@ -2,7 +2,7 @@ import { ProviderBuilder } from '@hub3js/core';
 
 import { metadata, WALLET_ID } from './constants.js';
 import { evm } from './namespaces/evm.js';
-import { initSafe } from './utils.js';
+import { initSafe } from './safe.ts';
```
RyukTheCoder: use "js" postfix for consistency

wallets/provider-safe/src/utils.ts
```diff
@@ -1,58 +1,15 @@
...
-  }
-
-  safeProvider = new SafeAppProvider(accountInfo, sdk);
-  return true;
-}
+import { getSafeProvider } from './safe.ts';
 
 export function evmSafe(): ProviderAPI {
-  if (!safeProvider) {
+  if (!getSafeProvider()) {
```
RyukTheCoder: you can rename to `getSafeProviderOrThrow`

RyukTheCoder (changes requested):

widget/embedded/src/utils/wallets.ts
llawliet-l-l: Removed in https://github.com/rango-exchange/rango-client/pull/1412/commits/e6a20fc61dc88a7f60884e3a461722cf7c8699f6 and moved to a new PR: https://github.com/rango-exchange/rango-client/pull/1423

wallets/provider-safe/src/utils.ts
llawliet-l-l: Fixed in https://github.com/rango-exchange/rango-client/pull/1412/commits/f8d296da72e4858fa148b50cc72b3d187b1d8eea

wallets/provider-safe/src/provider.ts
llawliet-l-l: Fixed in https://github.com/rango-exchange/rango-client/pull/1412/commits/f8d296da72e4858fa148b50cc72b3d187b1d8eea

RyukTheCoder (approved):
LGTM

llawliet-l-l:
About https://github.com/rango-exchange/rango-client/pull/1412/commits/a828267b1dcaa9678c4251ff0e904d71865a495a


  When running the widget as a Safe App inside Safe{Wallet}, a few seconds after
  executing a transaction the host console logs:                                                                                                                                                                                                        

  > `Code 901: Error processing Safe Apps SDK request (An error occurred)`

  It's cosmetic (the swap still tracks and completes), but it's noise on every
  Safe transaction and looks like a real failure.

  ## Root cause

  Safe is a contract (multisig) wallet, so the hash we get back when a transaction
  is submitted is **not** the on-chain transaction hash — it's the **safeTxHash**,
  an off-chain EIP-712 identifier for the *proposed* Safe transaction that owners
  sign. The Safe Apps provider returns this safeTxHash from `eth_sendTransaction`
  (and stores a placeholder "fake" tx for it); the real on-chain hash only exists
  once the Safe transaction is executed.

  - Safe Apps provider source — the `eth_sendTransaction` case returns
    `resp.safeTxHash`:
    https://github.com/safe-global/safe-apps-sdk/blob/main/packages/safe-apps-provider/src/provider.ts
  - Plain-English corroboration (wagmi) — "`eth_sendTransaction` through the
    connector returns a `safeTxHash`, which isn't equivalent to the onchain hash":
    https://github.com/wevm/wagmi/issues/2461

  Our status-polling loop (`checkStatus`) calls `CustomEvmSigner.wait()` roughly
  every 5s until the backend reports the swap as finished. The `wait`
  implementation resolves the on-chain hash via
  `sdk.txs.getBySafeTxHash(safeTxHash)`:

  - **First poll** — `wait` receives the safeTxHash → `getBySafeTxHash` succeeds →
    returns the on-chain hash. The queue updates `executedTransactionId` to that
    on-chain hash.
  - **Subsequent polls** — `wait` now receives the *on-chain hash* and calls
    `getBySafeTxHash(onChainHash)` again. The Safe Client Gateway resolves only by
    safeTxHash, so querying it with the on-chain hash returns `404`, and the Safe
    host logs `Code 901`.

  The `901` is emitted inside Safe{Wallet}'s own code before it returns the error
  to us, so it can't be suppressed from the app side — it can only be avoided by
  not making the redundant lookup.

  ## Fix

  Use the existing `hashRequiringUpdate` flag (the same flag `rango-preset` already
  reads to decide when to attach an explorer URL) to short-circuit the redundant
  call:

  - `signAndSendTx` marks the provisional hash with `hashRequiringUpdate: true`.
  - The first `wait` resolves the on-chain hash and flips it to
    `hashRequiringUpdate: false`.
  - Any later `wait` sees `hashRequiringUpdate === false` and returns the already-
    resolved hash immediately, without hitting the Safe gateway.

  The legitimate first-poll resolution (safeTxHash → on-chain hash) is unchanged;
  only the redundant follow-up lookups that produced the `901` are skipped.

PR #1413, opened 2026-06-08
feat: migrate Ctrl (formerly XDEFI) wallet to hub3js

arlert-armin:
Migrates the **Ctrl** wallet (formerly XDEFI) from legacy `wallets-core` to the new hub3js architecture, and renames the package `provider-xdefi` → `provider-ctrl`.

Three namespaces: **EVM** (rabby/tomo pattern), **UTXO** (grouping BTC/LTC/DOGE/BCH in one namespace — a first in the hub), **Solana** (phantom pattern). Cosmos (THORChain/MayaChain) support is dropped.

### 1. Shared core change — affects every UTXO wallet
- `wallets/core/src/namespaces/utxo/constants.ts` (+ `mod.ts`): new `bip122` CAIP ids for LTC / DOGE / BCH. **Note:** BCH shares Bitcoin's genesis block, so its CAIP-2 reference uses a later block-hash prefix to stay unique — worth a sanity check.
- `wallets/react/src/hub/helpers.ts`: the matching reverse-mapping (`CAIP reference → LTC/DOGE/BCH`). This is the consumer that resolves each grouped account to its blockchain.

### 2. Multi-chain UTXO + account-change handling (the trickiest logic)
- `provider-ctrl/src/utils.ts`, `actions/utxo.ts`, `builders/utxo.ts`: `connect` returns the addresses of all four UTXO chains at once (each CAIP-encoded with its own chain id).
- **Account switching is driven off the EVM `accountsChanged` event, not the UTXO providers.** Ctrl's UTXO providers emit an empty `{}` payload and re-fetching them while disconnected opens a wallet popup; the EVM provider gives the only usable signal (non-empty = switch, empty = disconnect). Rationale is documented at the top of `builders/utxo.ts`. This relies on EVM being connected alongside UTXO — always true since Ctrl grants all chains together.

### 3. Multi-chain UTXO signer
- `provider-ctrl/src/signers/utxo.ts`: BTC signs+broadcasts via the wallet's `sign_psbt` (`broadcast: true`); LTC/DOGE/BCH use the wallet's `transfer` method. All wallet-internal (no external RPC).

### 4. Rename / compatibility — please confirm
- `wallets/shared/src/rango.ts`: `WalletTypes.XDEFI = 'xdefi'` → `WalletTypes.CTRL = 'ctrl'`. **The provider id string changed**, which affects persisted auto-connect state and any backend wallet-type matching — confirm this is coordinated.
- `global-wallets-env.d.ts`: detection moved from `window.xfi` → `window.ctrl`.
- Consumers updated: `provider-all`, `queue-manager-demo`.

The rest (EVM/Solana namespaces, provider/mod boilerplate, file moves) follows the existing rabby/tomo/phantom patterns.


queue-manager/queue-manager-demo/src/flows/rango/helpers.ts
```diff
@@ -658,13 +653,7 @@ export const walletsAndSupportedChainsMetaSelector = (
     [WalletTypes.META_MASK]: evmBlockchains,
     [WalletTypes.COINBASE]: [...evmBlockchains, ...solanaBlockchain],
     [WalletTypes.PHANTOM]: solanaBlockchain,
-    [WalletTypes.XDEFI]: blockchainsArray.filter((blockchainMeta) =>
-      [
-        ...XDEFI_WALLET_SUPPORTED_EVM_CHAINS,
-        ...XDEFI_WALLET_SUPPORTED_NATIVE_CHAINS,
-        Networks.SOLANA,
-      ].includes(blockchainMeta.name as Networks)
-    ),
```
llawliet-l-l, line 667: Since we don't use this function anymore and also the other wallets data here are deprecated, I think it's better we don't change this function and have as few changes as possible

queue-manager/queue-manager-demo/package.json
```diff
@@ -20,7 +20,7 @@
   },
   "dependencies": {
     "@rango-dev/provider-all": "^0.62.2-next.1",
-    "@rango-dev/provider-xdefi": "^0.62.1-next.1",
+    "@rango-dev/provider-ctrl": "^0.62.1-next.1",
```
llawliet-l-l: Do we need it? isn't it better to completely remove it?

wallets/provider-ctrl/src/constants.ts
```diff
@@ -0,0 +1,80 @@
...
+
+// exported because it is also consumed by /queue-manager/queue-manager-demo
+export const SUPPORTED_ETH_CHAINS = [
+  LegacyNetworks.POLYGON,
+  LegacyNetworks.ETHEREUM,
+  LegacyNetworks.BSC,
+  LegacyNetworks.AVAX_CCHAIN,
+  LegacyNetworks.FANTOM,
+  LegacyNetworks.ARBITRUM,
+];
```
llawliet-l-l: I can't find its usage

wallets/provider-ctrl/src/constants.ts
```diff
@@ -0,0 +1,80 @@
...
+      value: {
+        selection: 'multiple',
+        data: [
+          {
+            label: 'EVM',
+            value: 'EVM',
+            id: 'ETH',
+            getSupportedChains: (allBlockchains: BlockchainMeta[]) =>
+              evmBlockchains(allBlockchains),
+          },
```
llawliet-l-l, line 59: shouldn't we filter by the `SUPPORTED_ETH_CHAINS` here?

wallets/provider-ctrl/src/utils.ts
```diff
@@ -0,0 +1,151 @@
...
+/**
+ * The UTXO chains Ctrl exposes, grouped under the single UTXO namespace, each
+ * paired with its CAIP-2 (bip122) reference so accounts can be self-describing.
+ */
+export const UTXO_CHAINS: { network: UtxoNetwork; caip: string }[] = [
+  { network: LegacyNetworks.BTC, caip: CAIP_BITCOIN_CHAIN_ID },
+  { network: LegacyNetworks.LTC, caip: CAIP_LITECOIN_CHAIN_ID },
+  { network: LegacyNetworks.DOGE, caip: CAIP_DOGECOIN_CHAIN_ID },
+  { network: LegacyNetworks.BCH, caip: CAIP_BITCOINCASH_CHAIN_ID },
+];
```
llawliet-l-l: Shouldn't it be in the `constants.ts` file? I think it's even better if we put it in `wallets/core/src/namespaces/utxo`. And use `as const` instead of defining `UtxoNetwork` type

wallets/provider-ctrl/src/utils.ts
```diff
@@ -0,0 +1,151 @@
...
+ * reliably through the EVM provider's `accountsChanged` (the UTXO providers emit an
+ * empty `{}` payload, and re-fetching them while disconnected opens a wallet popup).
+ * Returns an empty object when EVM isn't injected so the subscriber attaches safely.
+ */
+export function evmEventSource(): EvmProviderApi {
+  return (ctrl()?.get(LegacyNetworks.ETHEREUM) ?? {}) as EvmProviderApi;
+}
+
+/** Promisified `request_accounts` for a callback-style ctrl UTXO instance. */
+export async function requestUtxoAccounts(
```
llawliet-l-l: No need to export this

wallets/provider-ctrl/src/signers/utxo.ts
```diff
@@ -0,0 +1,166 @@
...
+
+import { LegacyNetworks } from '@rango-dev/wallets-core/legacy';
+import { SignerError, SignerErrorCode } from 'rango-types';
+
+const SUPPORTED_UTXO_CHAINS: string[] = [
+  LegacyNetworks.BTC,
+  LegacyNetworks.LTC,
+  LegacyNetworks.DOGE,
+  LegacyNetworks.BCH,
+];
```
llawliet-l-l: use the one in the `constants.ts` file

wallets/provider-ctrl/src/signers/utxo.ts
```diff
@@ -0,0 +1,166 @@
...
+/** Callback-style transfer for non-PSBT UTXO chains (LTC/DOGE/BCH). */
+async function ctrlTransfer(
+  blockchain: string,
+  ticker: string,
+  from: string,
+  amount: string,
+  decimals: number,
+  recipientAddress: string | null,
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  provider: any,
```
llawliet-l-l: Can't it be the `UtxoProviderApi` ?

llawliet-l-l (changes requested):
tnx, left some comments

queue-manager/queue-manager-demo/src/flows/rango/helpers.ts
arlert-armin: The demo no longer imports anything from `provider-ctrl`, and the selector entry was simplified. Also removed the now-unused `provider-ctrl` dependency from its `package.json` (see the other comment). (c37e0b690)

queue-manager/queue-manager-demo/package.json
arlert-armin: Removed — the demo no longer imports from `provider-ctrl`. (c37e0b690)

wallets/provider-ctrl/src/constants.ts
arlert-armin: Right, it was only used by the queue-manager-demo. Removed `SUPPORTED_ETH_CHAINS` entirely; EVM now exposes all chains via `evmBlockchains`. (c37e0b690)

wallets/provider-ctrl/src/constants.ts
arlert-armin: Intentionally not filtered. Ctrl supports the full EVM set (~21 named chains incl. Optimism/Base/Blast/Linea/zkSync) plus arbitrary chains via `wallet_addEthereumChain`/`wallet_switchEthereumChain`, so we expose all of them via `evmBlockchains` like rabby/tomo/bitget. The old 6-chain list was just a legacy restriction — removed. (c37e0b690)

wallets/provider-ctrl/src/utils.ts
arlert-armin: Switched to `as const` and dropped the `UtxoNetwork` type. On moving it into `wallets/core/src/namespaces/utxo`: since Ctrl is currently the only wallet that groups multiple UTXO chains, I'd prefer to keep it in the provider for now and promote it to core once there's a second consumer — feels out of scope for this PR. Happy to revisit if you disagree. (c37e0b690)

wallets/provider-ctrl/src/utils.ts
arlert-armin: Made it module-private (no longer exported). (c37e0b690)

wallets/provider-ctrl/src/signers/utxo.ts
arlert-armin: Done — `SUPPORTED_UTXO_CHAINS` is now defined once in `constants.ts` (derived from `UTXO_CHAINS`) and imported by the signer. (c37e0b690)

wallets/provider-ctrl/src/signers/utxo.ts
arlert-armin: Yes — typed it as `UtxoProviderApi` and dropped the `any`/eslint-disable. (c37e0b690)

llawliet-l-l (approved):
lgtm

llawliet-l-l (approved):
lgtm

PR #1405, opened 2026-05-25
feat: remove Cosmos blockchain support

arlert-armin:
## Summary

Rango is dropping Cosmos blockchain support. This PR removes the Cosmos signer package, all Cosmos wallet provider packages, and every Cosmos-specific branch across the codebase, with no dead code or broken builds left behind.

### What was removed

**Deleted packages**
- `@rango-dev/signer-cosmos`
- `@rango-dev/provider-keplr` (Cosmos-native)
- `@rango-dev/provider-cosmostation` (Cosmos + EVM-on-Cosmos)
- `@rango-dev/provider-leap-cosmos` (Cosmos-native)
- `@rango-dev/wallets-core/namespaces/cosmos` (sub-path export)

**Surgical strips (package kept, cosmos branch removed)**
- `provider-xdefi` — deleted `cosmos-signer.ts`, dropped `SUPPORTED_COSMOS_CHAINS`
- `provider-walletconnect-2` — removed `NAMESPACES.COSMOS`, `CosmosRPCMethods`, `DEFAULT_COSMOS_METHODS`, `CosmosMeta`, orphan mock file
- `provider-all` — dropped keplr/cosmostation/leap-cosmos imports + deps
- `wallets/shared` — deleted `getCosmosAccounts.ts`, removed `KEPLR`/`COSMOSTATION`/`LEAP_COSMOS` from `WalletTypes`, `KEPLR_COMPATIBLE_WALLETS`, `DEFAULT_COSMOS_RPC_URL`, `Cosmos` namespace entry
- `wallets/react/hub` — deleted `chainRegisteryConvertors`, removed `isCosmosNamespace` + all cosmos hub branches
- `wallets/core` — narrowed `Namespace` union, removed `cosmos: CosmosActions` from `CommonNamespaces`, dropped `COSMOS`/`OSMOSIS` from legacy `Networks` enum, removed `@chain-registry/types` dep
- `widget/embedded` — removed `isExperimentalChain`, `getKeplrCompatibleConnectedWallets`, full `ExperimentalChain*` UI flow, cosmos branches in `TokenList`, `CustomTokenModal`, etc.
- `widget/ui` — deleted `CosmosCategoryIcon`, removed `BlockchainCategories.COSMOS`
- `widget/playground` — replaced Keplr demo with Rabby
- `queue-manager/rango-preset` — cosmos transaction handling now throws "no longer supported" explicitly
[...]

PR #1415, opened 2026-06-09
fix: align category list loading skeleton count with the 4 chips

arlert-armin:
## What

After Cosmos support was removed in #1405, the blockchain category list renders **4** chips (`ALL`, `EVM`, `UTXO`, `OTHER`), but the loading skeleton was still hardcoded to render **5** skeletons. When `isLoading` flipped to `false`, the extra placeholder collapsed and caused a visible UI glitch / layout shift on load.

This reduces the skeleton count from 5 to 4 so the loading state matches the number of chips actually rendered.

## Why

Regression introduced by the Cosmos removal in #1405. Reported by users.

## Changes

- `widget/ui/src/components/SelectableCategoryList/SelectableCategoryList.tsx`: loading skeleton loop `Array(5)` → `Array(4)`.


nikaaru (approved):
Approved.

PR #1420, opened 2026-06-15
fix: add wallet error logging

llawliet-l-l:
## Summary

  Wallet connect errors (e.g. Phantom's "User rejected the request") were only shown in the
  widget UI and never passed through the logging pipeline — so they appeared nowhere in
  console or Sentry, making them impossible to diagnose. This PR routes those errors through
  `@rango-dev/logging-core` and configures the loggers so expected, low-severity errors are
  visible to developers without polluting Sentry.

  ## Changes

  - Log the connect error in `NamespaceDetachedItem` (`handleConnectNamespace`) via
    `@rango-dev/logging-core` at **debug** level.
  - Debug level is deliberate: these are predominantly expected user-rejection errors.

  Why debug + split levels

  User rejections and similar connect failures are routine and user-initiated, not application
  faults. Sending them to Sentry would create noise without signal. Logging them at debug
  keeps them available for local debugging while the Warn-gated Sentry pipeline stays
  focused on genuine issues.

  # Testing

  - Trigger a Phantom BTC connect and reject the request in the wallet popup.
  - Confirm the error message appears in the browser console.

# Checklist:

### DIFF
--- a/logging/core/src/index.ts
+++ b/logging/core/src/index.ts
@@ -1,1 +1,1 @@
-export { error, warn, info, debug, trace, ev } from './logging';
+export { error, warn, info, debug, trace, ev } from './logging.js';
--- a/logging/core/src/logging.ts
+++ b/logging/core/src/logging.ts
@@ -2,7 +2,7 @@ import type { Data, EventPayload, Message } from '@rango-dev/logging-types';
 
 import { EventType, Level } from '@rango-dev/logging-types';
 
-import { isValidInstance } from './helpers';
+import { isValidInstance } from './helpers.js';
 
 export function ev(level: Level, message: Message, data?: Data) {
   if (!isValidInstance(message)) {
--- a/wallets/react/package.json
+++ b/wallets/react/package.json
@@ -40,6 +40,7 @@
     "@hub3js/namespaces": "^0.1.1",
     "@hub3js/solana": "^0.2.1",
     "@hub3js/std": "^0.1.1",
+    "@rango-dev/logging-core": "^0.12.1",
     "@rango-dev/wallets-core": "^0.59.1-next.1",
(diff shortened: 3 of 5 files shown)


llawliet-l-l:
## New Changes

  Adds error logging to the wallet provider API in `wallets-react`, and
  makes `logging-core` consumable by node16 packages (required for the
  above).

  ### wallets-react — wallet error logging
  - Wrap every method on the provider `api` so any thrown (sync) or rejected
    (async) error is logged via `@rango-dev/logging-core`'s `debug()` and
    then rethrown unchanged.
  - New `src/helpers.ts` with:
    - `withErrorLogging` — wraps a single method; inspects the *returned
      value* (`instanceof Promise`) to handle sync and async methods without
      turning sync methods async.
    - `withErrorLoggingApi` — applies the wrapper to all api methods.
    - module-level `logError` — normalizes non-Error throws and logs with
      `{ context: { method, args }, tags: { type: 'wallet-error' } }`.
  - `useProviders` now returns `withErrorLoggingApi(api)`.
  - Added `@rango-dev/logging-core` as a dependency.

  ### logging-core — node16 compatibility
  - Added explicit `.js` extensions to relative imports so the emitted
    `.d.ts` is valid under `moduleResolution: node16`. `wallets-react` is
    the first node16 consumer of this package and was failing to build with
    TS2834.

  ## Why
  `wallets-react` builds under `node16` (via `tsconfig.libnext.json`), which
  enforces explicit extensions on the relative specifiers inside any `.d.ts`
  it loads. `logging-core` emitted `from './logging'`, breaking the build.
  Fixing at the source keeps the package correct for all consumers instead
  of masking it per-package.

RyukTheCoder (approved):
LGTM

PR #1421, opened 2026-06-15
Feat/add utxo namespace to wallet connect

RyukTheCoder:
# Summary

Add a UTXO namespace to the WalletConnect hub provider with Bitcoin connect/disconnect/eager-connect support and a transfer signer.

# Changes
- New namespaces/utxo with connect, disconnect, and eager-connect actions
- Bitcoin transfer signer (signers/utxo.ts) and BTC entry in provider metadata

# Test Plan

 - UTXO: connect WalletConnect with BTC namespace selected; verify Bitcoin address is returned
 - UTXO: sign a Bitcoin transfer (if applicable in your test flow)

# Checklist:

### DIFF
--- a/wallets/provider-walletconnect-2/src/namespaces/utxo/hooks.ts
+++ b/wallets/provider-walletconnect-2/src/namespaces/utxo/hooks.ts
@@ -1,0 +1,130 @@
+import type { UtxoActions } from '@rango-dev/wallets-core/namespaces/utxo';
+
+import {
+  CAIP_BITCOIN_CHAIN_ID,
+  utils,
+} from '@rango-dev/wallets-core/namespaces/utxo';
+import { AccountId } from 'caip';
+
+import {
+  type Bip122AddressEntry,
+  filterBip122Accounts,
+  getAnnouncedAddresses,
+  pickPaymentAddress,
+} from '../../session/bip122.js';
+import { expireWalletConnectTopic } from '../../session/teardown.js';
+import { BitcoinEvents, NAMESPACES } from '../../wcConstants.js';
+import { createSessionSubscriber } from '../subscriber.js';
+
+/**
+ * WC signals a bip122 account switch via `session_update` (re-shared namespaces
+ * with the newly-selected account first), so the active account is published
+ * from this event.
+ */
+export function sessionUpdateSubscriber() {
+  return createSessionSubscriber<'session_update', UtxoActions>(
+    'utxo',
+    'session_update',
+    ({ args, session, adapter, context }) => {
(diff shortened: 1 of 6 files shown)


wallets/provider-walletconnect-2/src/namespaces/utxo/hooks.ts
```diff
@@ -0,0 +1,165 @@
...
+      const adapter = getAdapter();
+      void adapter.getClient().then((client) => {
+        if (handler) {
+          client.off('session_update', handler);
+        }
+      });
+      return err;
+    },
+  ];
+}
```
arlert-armin: Please use `switchAccountSubscriberBuilder`

arlert-armin (changes requested):

wallets/provider-walletconnect-2/src/namespaces/utxo/hooks.ts
RyukTheCoder: Applied

arlert-armin (approved):
LGTM

RyukTheCoder:
### `c40a8b0` — fix(walletconnect-2): share one AppKit modal across namespaces

**Bug:** a stray back button appeared in the WalletConnect modal header when opening one namespace's modal after the other (e.g. connect EVM, then open BTC).

**Root cause:** AppKit's controllers (`RouterController`, `ChainController`, …) are module-level singletons. The per-namespace modal was rebuilt on every switch, so each rebuild re-ran AppKit init against those same globals and left the previous open's router history behind → an extra router step → the back arrow. There was never really more than one live instance to begin with.

**Fix:** build the modal **once** with both networks registered up front (`[mainnet, bitcoin]`) and pick the active one per connect with `switchNetwork`. `open()` then sees a matching active chain and takes its clean, non-switching path — no back button.

Changes:
- **`modal.ts`** — single instance; `prepareModalForNamespace` runs `ready()` + `switchNetwork()` before `open()`. The active chain is now shared mutable state, safe only because connects are serialized by the adapter's `#sessionQueue` (documented inline).
- **`modal-cache.ts`** — build once, no per-namespace rebuild.
- **`adapter.ts`** — drop `namespace` from `getModal` (one instance now).
- **`lifecycle.ts`** — call `prepareModalForNamespace` before `open`; reuse `WC_NAMESPACE_TO_CAIP` for the open `namespace` instead of sniffing the connect-payload keys.

Note: the earlier `ConnectorController` filter attempt and the extra `@reown/appkit-controllers` dependency were reverted — the 530→9/40 wallet count turned out to be a harmless placeholder, not a bug. This commit is back-button-only, no new dependency.

Green: `ts-check`, `lint`, `build` all pass.

PR #1422, opened 2026-06-16
ci: speed up Check PR workflow

arlert-armin:
# Summary

Speeds up the **Check PR** workflow by making nx reuse already-built outputs across jobs and runs, and stops the intermittent memory-exhaustion failures in `build-examples`. No nx Cloud — everything uses nx's local cache persisted via `actions/cache`.

**Motivation:** `deploy` rebuilt the apps' whole dependency tree (~40 packages, dragged in by `provider-all`) that `build-libs` had already built, and `build-examples` built all 7 examples at once, occasionally exhausting the runner's memory (20-minute runs / OOM failures).

**Changes**

1. **nx local computation cache (no nx Cloud)**
   - `nx.json`: mark the `build` target cacheable — `cache: true` + `outputs: ["{projectRoot}/dist"]`.
   - Cache **both** `.nx/cache` (task outputs) and `.nx/workspace-data` (the cache database that indexes them).
   - Set `NX_REJECT_UNKNOWN_LOCAL_CACHE=0` so nx accepts cache restored from a different runner.
   - Key on `github.sha` with an `nx-` `restore-keys` fallback: `deploy` reuses `build-libs`' cache within the same run (cross-job); later runs reuse the most recent cache (cross-run).
   - Result: `build-libs` replays unchanged libraries; `deploy`'s `nx run-many` compiles only the 3 deployable apps.

2. **build-examples**
   - Bound the concurrent build fan-out (`--max-processes 3`) and cap each build's Node heap (`--max-old-space-size=4096`) — fixes the memory-exhaustion 20-minute / OOM runs.
   - Cache the bundlers' incremental build caches that live outside `node_modules` (`.next/cache`, `.parcel-cache`).

3. **Faster checkout** — drop `fetch-depth: 0` from jobs that don't need full git history (`check-commits` keeps it for `git log origin/<base>..HEAD`).

**Caveats**
- `NX_REJECT_UNKNOWN_LOCAL_CACHE=0` is discouraged by nx for shared *network drives*; on GitHub Actions the cache is branch-scoped and written by trusted runs, so the risk is low — this is the deliberate local-cache choice in lieu of nx Cloud.
- Workflow is `pull_request`-only, so cross-*run* reuse warms on a PR's 2nd+ push; the within-run cross-job reuse (`deploy`) works on every run.
- The `dist` artifact is unchanged and still required by `run-tests` / `build-examples` / `analyze-bundle` (they consume `dist` without running nx).

Fixes # (no linked issue — CI performance improvement)

Ran the full **Check PR** workflow on this branch multiple times (cold cache, then warm) and measured per-job durations from the Actions logs:

- [x] **build-libs** (warm): build step `200s → 1.2s` (nx replays unchanged libs); job `~240s → 58s`
- [x] **deploy** (warm): client build `232s → 35s` (nx cache hit, only the 3 apps compiled); job `~214–243s → 92s`
- [x] **build-examples**: completes within memory limits — no OOM / 20-minute runs
- [x] All jobs green on the branch (`check-commits`, `run-tests`, `analyze-bundle`, previews)

# Checklist:

- [ ] I have made corresponding changes to the documentation (N/A — CI config only)
- [ ] I have added tests that prove my fix is effective or that my feature works (N/A — verified via the workflow runs above)
- [ ] Implemented a user interface (UI) change, referencing our Figma design to ensure pixel-perfect precision (N/A)

### DIFF
--- a/.github/workflows/checks.yaml
+++ b/.github/workflows/checks.yaml
@@ -3,12 +3,18 @@ on:
   pull_request:
     types: [opened, reopened, synchronize, labeled]
 
+# Force Nx to accept and reuse computation caches restored from different CI runners
+env:
+  NX_REJECT_UNKNOWN_LOCAL_CACHE: '0'
+
 jobs:
   check-commits:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
         with:
+          # Full history is required so check-conventional-commits can run
+          # `git log origin/<base>..HEAD`.
           fetch-depth: 0
       - name: Prepare
         uses: ./.github/actions/prepare
@@ -21,11 +27,21 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
-        with:
-          fetch-depth: 0
 
       - name: Prepare
         uses: ./.github/actions/prepare
+
+      # Cache Nx outputs (.nx/cache) and metadata (.nx/workspace-data) to optimize incremental builds
+      - name: Cache nx computation cache
+        uses: actions/cache@v4
+        with:
+          path: |
+            .nx/cache
+            .nx/workspace-data
+          key: ${{ runner.os }}-nx-${{ github.sha }}
+          restore-keys: |
+            ${{ runner.os }}-nx-
+
       - name: Extract new translation keys and check if any error exists
         run: node ./scripts/lingui/command.js extract --clean
 
@@ -44,8 +60,6 @@ jobs:
     needs: build-libs
     steps:
       - uses: actions/checkout@v4
-        with:
-          fetch-depth: 0
       - name: Download Build Information For Current Branch
         uses: actions/download-artifact@v4
         with:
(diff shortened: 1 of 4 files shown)


.github/workflows/checks.yaml
```diff
@@ -3,12 +3,20 @@ on:
   pull_request:
     types: [opened, reopened, synchronize, labeled]
 
+# nx's local cache rejects artifacts created on a different machine by default.
```
nikaaru: NIT: Write comment in one line:

```suggestion

# Force Nx to accept and reuse computation caches restored from different CI runners

```

.github/workflows/checks.yaml
```diff
@@ -21,11 +29,26 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
-        with:
-          fetch-depth: 0
 
       - name: Prepare
         uses: ./.github/actions/prepare
+
+      # Persist nx's computation cache so `nx run-many` only recompiles the
```
nikaaru: NIT

```suggestion
# Cache Nx outputs (.nx/cache) and metadata (.nx/workspace-data) to optimize incremental builds
```

.github/workflows/checks.yaml
```diff
@@ -77,17 +96,31 @@ jobs:
         with:
           name: ${{needs.build-libs.outputs.hashed_branch}}
 
+      # Restore the nx cache built by build-libs so build-clients' `nx run-many`
```
nikaaru: NIT

```suggestion
# Read-only restore of the Nx build state to skip rebuilding shared dependencies
```

nikaaru (approved):
LGTM

.github/workflows/checks.yaml
arlert-armin: Done

.github/workflows/checks.yaml
arlert-armin: Applied

.github/workflows/checks.yaml
arlert-armin: Applied

PR #1423, opened 2026-06-17
fix: convert chain id to chain name to compare to the user selected c…

llawliet-l-l:
# Summary

Hub EVM wallets report `network` as a chain id (e.g. `0x1`), whereas `chain` is a Rango blockchain name (e.g. `ETH`). For contract wallets (e.g. Safe), we compare the two to check that the wallet is connected to the required chain, so we normalize the chain id to a name first. `getBlockChainNameFromId` returns the value unchanged when it's already a name (legacy wallets), so this stays correct for both.

We have to convert here because legacy and hub differ on where the conversion happens. In legacy, the provider also returned a numeric chain id, but the `Wallet` class converted it via `getBlockChainNameFromId` inside `connect()` before writing it to `state.network`, so consumers always read a network name. Hub dropped that central step: the connect action's raw chain id flows straight into `state.network` (via `connectAndUpdateStateForMultiNetworks`), so the conversion has to be reintroduced wherever `network` is read.

You can refer to this PR for test details: https://github.com/rango-exchange/rango-client/pull/1412

# Checklist:

### DIFF
--- a/widget/embedded/src/utils/wallets.ts
+++ b/widget/embedded/src/utils/wallets.ts
@@ -24,6 +24,7 @@ import { legacyReadAccountAddress as readAccountAddress } from '@rango-dev/walle
 import {
   detectInstallLink,
   HYPERLIQUID_SIGN_NETWORK,
+  getBlockChainNameFromId,
   isEvmAddress,
   Networks,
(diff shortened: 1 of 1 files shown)


RyukTheCoder (approved):
LGTM

PR #1433, opened 2026-07-01
feat: add Bitcoin (UTXO) support to Trezor provider

arlert-armin:
## Add Bitcoin (UTXO) support to the Trezor provider — Draft / RFC

Adds a `UTXO` namespace to the hub Trezor provider so users can connect a Bitcoin
account (selectable address type: Native SegWit / Nested SegWit / Legacy / Taproot) and
sign BTC swaps. Verified end-to-end on-device (connect → sign → broadcast).

### Open question I'd like a decision on

Trezor Connect has **no PSBT API** — its [`signTransaction`](https://github.com/trezor/connect/blob/develop/docs/methods/signTransaction.md)
only accepts explicit `inputs`/`outputs`. Every other wallet in the repo receives Rango's
BTC **PSBT** and hands it to the wallet's own `signPsbt`, but Trezor can't do that. So here
I parse Rango's PSBT with `bitcoinjs-lib` and translate it into Trezor's `inputs`/`outputs`
for `signTransaction`.

It works, but I'm not sure it's the best long-term approach. An alternative would be to have
the backend send Trezor a plain transfer (amount + recipient) and use Trezor's
`composeTransaction` instead. **Leaving this as a draft for a decision on which direction we
want.**

### Notes

- Rango's BTC PSBT does **not** include `bip32Derivation`, so the signer remembers the
  connect-time path in `state.ts` — the same pattern the EVM/Solana hardware signers already
  use.
- `refTxs` are omitted; Trezor Connect fetches previous transactions from its Blockbook
  backend.
- The PSBT → `signTransaction` conversion is a pure function with unit tests.

### References that Trezor Connect doesn't take a PSBT

- Official docs — [`signTransaction` requires `coin`/`inputs`/`outputs`, no PSBT param](https://github.com/trezor/connect/blob/develop/docs/methods/signTransaction.md)
- Feature request to sign externally-built / BIP-174 transactions — [trezor/python-trezor#183](https://github.com/trezor/python-trezor/issues/183)
- `@trezor/utxo-lib` ships no PSBT module (removed); `@trezor/connect` exposes no PSBT method.

### DIFF
--- a/wallets/provider-trezor/src/signers/utxo.ts
+++ b/wallets/provider-trezor/src/signers/utxo.ts
@@ -1,0 +1,90 @@
+import type { GenericSigner, Transfer } from 'rango-types';
+
+import { Networks } from '@rango-dev/wallets-shared';
+import { SignerError } from 'rango-types';
+
+import { getTrezorModule } from '../legacy/helpers.js';
+import { getBitcoinDerivationPath } from '../state.js';
+import { BITCOIN_COIN_NAME } from '../utxo/config.js';
+import { buildTrezorBitcoinTransaction } from '../utxo/psbt.js';
+
+function getTrezorErrorMessage(payload: { error: string; code?: string }) {
+  return new Error(
+    payload.code ? `${payload.error} (${payload.code})` : payload.error
+  );
+}
+
+export class BTCSigner implements GenericSigner<Transfer> {
+  async signMessage(): Promise<string> {
+    throw SignerError.UnimplementedError('signMessage');
+  }
+
+  async signAndSendTx(tx: Transfer): Promise<{ hash: string }> {
+    const { blockchain } = tx.asset;
+    if (blockchain !== Networks.BTC) {
+      throw new Error(
+        `Signing ${blockchain} transactions is not supported by Trezor.`
+      );
+    }
+
+    const { psbt } = tx;
+    if (!psbt) {
+      throw new Error(
+        'No PSBT found to sign. Ensure a valid PSBT is provided.'
+      );
+    }
+
+    const path = getBitcoinDerivationPath();
+    if (!path) {
+      throw new Error(
+        'No connected Bitcoin account found. Please connect the wallet first.'
+      );
+    }
+
+    const TrezorConnect = await getTrezorModule();
+    const { inputs, outputs, version, locktime } =
+      buildTrezorBitcoinTransaction(psbt.unsignedPsbtBase64, path);
+
+    // `push: true` signs and broadcasts via Trezor's Blockbook backend.
+    const result = await TrezorConnect.signTransaction({
+      coin: BITCOIN_COIN_NAME,
+      inputs,
+      outputs,
+      version,
+      locktime,
+      push: true,
+    });
+
+    if (!result.success) {
+      throw getTrezorErrorMessage(result.payload);
+    }
+
+    return { hash: await this.#resolveHash(result.payload) };
+  }
+
+  /**
+   * `push: true` usually returns the txid directly. If it isn't present (older
+   * firmware/connect), push the serialized transaction explicitly as a fallback.
+   */
+  async #resolveHash(payload: {
+    txid?: string;
+    serializedTx?: string;
+  }): Promise<string> {
+    if (payload.txid) {
+      return payload.txid;
+    }
+    if (!payload.serializedTx) {
+      throw new Error('Trezor did not return a transaction id.');
+    }
+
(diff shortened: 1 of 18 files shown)


wallets/shared/src/rango.ts
```diff
@@ -115,6 +115,28 @@ export const namespaces: Record<
   UTXO: {
     title: 'UTXO',
     mainBlockchain: 'BTC',
+    derivationPaths: [
```
RyukTheCoder, line 118: We need to make sure other dexes use similar derivation paths. Also label should match the output of `generateDerivationPath`. for example, label for the first item should be "Native SegWit 84'/0'/${index}'/0/0"

RyukTheCoder (changes requested):
Tnx, just added one comment. Also do not forget to update readme files.

wallets/provider-trezor/src/signers/utxo.ts
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    const TrezorConnect = await getTrezorModule();
+    const { inputs, outputs } = buildTrezorBitcoinTransaction(
+      psbt.unsignedPsbtBase64,
+      path
+    );
+
+    // `push: true` signs and broadcasts via Trezor's Blockbook backend.
+    const result = await TrezorConnect.signTransaction({
```
RyukTheCoder, line 49: Please make sure nothing gets dropped from the original PSBT when you want to sign it. For example, here `version` and `locktime` for PSBT and sequence for each input are dropped.

wallets/shared/src/rango.ts
arlert-armin: Fixed — labels now include the full generated path (e.g. `Native SegWit (m/84'/0'/index'/0/0)`) so they match `generateDerivationPath`. The four paths follow the standard BIP registry purposes used across wallets: BIP-84 (`84'`) native segwit, BIP-49 (`49'`) nested segwit, BIP-44 (`44'`) legacy, BIP-86 (`86'`) taproot — all under Bitcoin coin type `0'`.

wallets/provider-trezor/src/signers/utxo.ts
arlert-armin: Good catch — fixed. `version`, `locktime` and per-input `sequence` are now read from the PSBT and passed to `signTransaction`, so the signed tx matches Rango's exactly (added a unit test).

I also did a full BIP-174 field audit so nothing that changes the resulting transaction is dropped:
- `sighashType` → now rejected if it isn't `SIGHASH_ALL` (Trezor's signTransaction can't honor a custom per-input sighash, so we throw rather than sign the wrong thing).
- `partialSig` / `finalScriptSig` / `finalScriptWitness` → don't appear in an unsigned PSBT.
- `bip32Derivation` → Rango's PSBT doesn't include it, so we apply the connect-time path.
- `redeemScript` / `witnessScript` (multisig) and tap script-path fields → out of scope for single-key swaps; single-key nested-segwit/taproot are derived by Trezor via `script_type`.
- output `bip32Derivation` → only affects change display on-device, not the tx bytes.

wallets/shared/src/rango.ts
arlert-armin: Checked how other DEX aggregators / hardware-wallet apps derive BTC paths:

- **THORSwap / SwapKit** (cross-chain DEX aggregator, open-source SDK): BTC on Ledger uses `m/84'/0'/{account}'/0/0` (native segwit, BIP-84) — e.g. the array `[84, 0, 1, 0, 0]`. The index varies the account, first receive address — **same structure as ours** (`84'/0'/index'/0/0`).
- **Ledger Live** (and its built-in swap): default native segwit BIP-84, `m/84'/0'/{account}'/0/{index}`, coin type `0'`.
- **Trezor Suite**: BIP-44 / 49 / 84 / 86, coin type `0'`.

All of them use the SLIP-44 coin type `0'` + BIP purposes 44'/49'/84'/86'. Our four templates follow the same standard and match SwapKit's hardware-wallet BTC path exactly, so we're consistent with other dexes. (Reverted the label-text tweak.)

Refs: [THORSwap SwapKit Ledger docs](https://docs.thorswap.finance/swapkit-docs/swapkit-sdk/wallets/swapkit-wallet-ledger), [Ledger derivation paths](https://www.ledger.com/blog/understanding-crypto-addresses-and-derivation-paths), [Trezor standard derivation paths](https://wiki.trezor.io/Standard_derivation_paths).

RyukTheCoder (approved):

PR #1439, opened 2026-07-12
feat(provider-okx): add Tron namespace support

llawliet-l-l:
## Summary

Adds **Tron** namespace support to the OKX wallet provider (hub / v1 provider). The
legacy v0 provider is untouched.

## Changes

- Wire OKX's injected `window.okxwallet.tronLink` into the hub across
  `actions/`, `builders/`, `namespaces/` Tron files (`types.ts` / `utils.ts` /
  `provider.ts` glue).
- **Connect** via `tron_requestAccounts`, reading the address from
  `tronWeb.defaultAddress.base58`. User rejection (EIP-1193 `4001`) surfaces a clear
  `User rejected the request.` error; other non-OK codes surface the wallet's message.
- **Account switch** handled via OKX's `walletChanged` provider-stream event — the Tron
  entry is matched by `chainId === '195'`; empty `params` triggers a disconnect.
- **Signing** registered through the shared `@rango-dev/signer-tron`
  (`tronWeb.trx.sign` / `sendRawTransaction`); **eager-connect** via the `ready` flag.
- Metadata namespace entry + `@rango-dev/signer-tron` dependency; matrix set to ✅.

## Known limitation

On page load / in-app disconnect+reconnect, OKX returns the **first-connected** Tron
account, not the last switched-to one — the extension resets the address and emits no
load-time event, so there's no client-side fix (the official
`@tronweb3/tronwallet-adapter-okxwallet` has the same limitation). Signing with that
stale account triggers OKX's `Transaction Unavailable` error. Documented in the provider
readme.

## Testing

- `yarn build` all pass.
- Manually tested on the shipped extension: connecting, disconnecting, switching
  accounts, and other actions all work as expected.
- Real Tron transaction signed end-to-end.

# Checklist:

### DIFF
--- a/wallets/provider-okx/src/actions/tron.ts
+++ b/wallets/provider-okx/src/actions/tron.ts
@@ -1,0 +1,37 @@
+import { utils } from '@rango-dev/wallets-core/namespaces/tron';
+
+import {
+  TRON_OK_REQUEST_CODE,
+  TRON_USER_REJECTION_CODE,
+} from '../constants.js';
+import { tronOKX } from '../utils.js';
+
+const connect = async () => {
+  const instance = tronOKX();
+  const accountsResult = await instance.request({
+    method: 'tron_requestAccounts',
+  });
+
+  if (accountsResult?.code && accountsResult.code !== TRON_OK_REQUEST_CODE) {
+    if (accountsResult.code === TRON_USER_REJECTION_CODE) {
+      throw new Error('User rejected the request.');
+    }
+    throw new Error(
+      accountsResult.message ?? 'Failed to connect to OKX Wallet Tron.'
+    );
+  }
+  return utils.formatAccountsToCAIP([instance.tronWeb.defaultAddress.base58]);
+};
+
+/*
+ * Wrapped in a try-catch in the case of wallet injection delay
+ */
+const canEagerConnect = async () => {
+  try {
+    const tronInstance = tronOKX();
+    return !!tronInstance.ready;
+  } catch {
+    return false;
+  }
+};
+export const tronActions = { connect, canEagerConnect };
--- a/wallets/provider-okx/src/builders/tron.ts
+++ b/wallets/provider-okx/src/builders/tron.ts
@@ -1,0 +1,63 @@
+import { ChangeAccountSubscriberBuilder } from '@hub3js/std/hooks';
+import {
+  type ProviderAPI,
+  type TronActions,
+  utils,
+} from '@rango-dev/wallets-core/namespaces/tron';
+
+import { isOkxTronMessageEvent } from '../utils.js';
+
+export const changeAccountSubscriber = (getInstance: () => ProviderAPI) =>
+  new ChangeAccountSubscriberBuilder<
+    MessageEvent<unknown>,
+    ProviderAPI,
+    TronActions
+  >()
+    .getInstance(getInstance)
+    .onSwitchAccount((event, context) => {
+      const data = event.payload?.data;
+
+      if (!isOkxTronMessageEvent(data)) {
+        event.preventDefault();
+        return;
+      }
+
+      const { action } = data.message;
+
+      if (action === 'disconnect') {
+        event.preventDefault();
+        context.action('disconnect');
+        return;
+      }
+
+      /*
+       * Change account only on `accountsChanged` with a string address (`format`
+       * reads it). Everything else — including `accountsChanged` with address
+       * `false` (its disconnect is handled above) — is ignored.
+       */
+      if (
+        action !== 'accountsChanged' ||
+        typeof data.message.data?.address !== 'string'
+      ) {
+        event.preventDefault();
+      }
+    })
+    .format(async (_, event) => {
+      const data = event.data;
+      const address =
+        isOkxTronMessageEvent(data) && data.message.action === 'accountsChanged'
+          ? data.message.data?.address
+          : undefined;
+      if (typeof address !== 'string') {
+        throw new Error('No Tron address received from OKX Wallet.');
+      }
+      return utils.formatAccountsToCAIP([address]);
+    })
+    .addEventListener((_, callback) => {
+      window.addEventListener('message', callback);
+    })
+    .removeEventListener((_, callback) => {
+      window.removeEventListener('message', callback);
+    });
+
+export const tronBuilders = { changeAccountSubscriber };
--- a/wallets/provider-okx/src/constants.ts
+++ b/wallets/provider-okx/src/constants.ts
@@ -7,6 +7,7 @@ import {
   solanaBlockchain,
   tonBlockchain,
   type TransferBlockchainMeta,
+  tronBlockchain,
 } from 'rango-types';
 
 import getSigners from './signer.js';
@@ -15,6 +16,9 @@ import { getInstanceOrThrow } from './utils.js';
 export const WALLET_ID = 'okx';
 export const TON_CONNECT_PROTOCOL_VERSION = 2;
 export const TON_CONNECT_USER_REJECTED_CODE = 300;
+export const TRON_OK_REQUEST_CODE = 200;
+// EIP-1193 user-rejected-request code returned by `tron_requestAccounts`.
+export const TRON_USER_REJECTION_CODE = 4001;
 
 export const OKX_WALLET_SUPPORTED_CHAINS = [
   LegacyNetworks.ETHEREUM,
@@ -83,6 +87,13 @@ export const metadata: ProviderMetadata = {
             id: 'TON',
             getSupportedChains: (allBlockchains: BlockchainMeta[]) =>
               tonBlockchain(allBlockchains),
+          },
+          {
+            label: 'Tron',
+            value: 'Tron',
+            id: 'TRON',
+            getSupportedChains: (allBlockchains: BlockchainMeta[]) =>
+              tronBlockchain(allBlockchains),
           },
         ],
       },
--- a/wallets/provider-okx/src/types.ts
+++ b/wallets/provider-okx/src/types.ts
@@ -5,8 +5,10 @@ import type {
   SOLANA_NAMESPACE,
   UTXO_NAMESPACE,
   TON_NAMESPACE,
+  TRON_NAMESPACE,
 } from '@hub3js/namespaces';
 import type { ProviderAPI as SolanaProviderApi } from '@hub3js/solana';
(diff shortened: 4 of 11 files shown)


wallets/provider-okx/src/constants.ts
```diff
@@ -6,12 +6,16 @@ import {
...
+  tronBlockchain,
 } from 'rango-types';
 
 import getSigners from './signer.js';
 import { getInstanceOrThrow } from './utils.js';
 
 export const WALLET_ID = 'okx';
+export const TronOKRequestCode = 200;
+// EIP-1193 user-rejected-request code returned by `tron_requestAccounts`.
+export const TronUserRejectionCode = 4001;
```
RyukTheCoder:
```suggestion
export const EIP_1193_USER_REJECTION_ERROR_CODE= 4001;
```
we define constants as UPPER_SNAKE_CASE

wallets/provider-okx/src/builders/tron.ts
```diff
@@ -0,0 +1,74 @@
...
+import { isOkxTronStreamEvent } from '../types.js';
+
+const OKX_TRON_STREAM_TARGET = 'okexwallet-inpage-tron';
+const OKX_TRON_STREAM_PROVIDER = 'okexwallet-provider-tron';
+// Fires on an account switch; its params holds the single newly-active wallet.
+const OKX_TRON_WALLET_CHANGED_METHOD = 'walletChanged';
+// OKX's chainId (a string) for the Tron entry in a wallet's per-network address list.
+const TRON_CHAIN_ID = '195';
+
+const isOkxTronWalletChangedEvent = (
```
RyukTheCoder: Please add a reference to the related docs. Shouldn't we subscribe on "accountsChanged"? See this: https://web3.okx.com/onchainos/dev-docs/wallet/dapp-connect/chains/tron/provider#events

RyukTheCoder (changes requested):

wallets/provider-okx/src/builders/tron.ts
llawliet-l-l: There are no docs for this. The original docs say it uses TronLink message formats, but from the actual behavior I've observed, the messages are MetaMask-style.
https://github.com/rango-exchange/rango-client/pull/1439/changes/BASE..63ded043042e1db340f8559cf7fc4c319445d468#r3566742919

wallets/provider-okx/src/types.ts
```diff
@@ -12,8 +13,85 @@ export type ProviderObject = {
...
+ * provider-stream messages whose `event.data` is:
+ *   { target: 'okexwallet-inpage-tron',
+ *     data: { name: 'okexwallet-provider-tron',
+ *             data: { baseChain: 'tron', method, params } } }
+ * Observed methods (live): 'wallet_chainChanged', 'walletChanged' (active account
+ * switched — the one we act on), 'connectedWalletsChanged' (lists every connected
+ * wallet in no reliable order). For 'walletChanged', `params` holds the single
+ * newly-active wallet, whose `address` is a per-network list from which the Tron
+ * entry (see OkxTronAddressEntry) supplies the base58 address.
+ */
```
llawliet-l-l, line 35: Also added this comment to emphasize the differences between the documentation and the actual behavior.

wallets/provider-okx/src/constants.ts
llawliet-l-l: Fixed in https://github.com/rango-exchange/rango-client/pull/1439/commits/4eca2dc294c524599f694a304d3315020f89acf7

wallets/provider-okx/src/builders/tron.ts
RyukTheCoder: On the deployed preview environment, I subscribed to the documented event by running this code in the console:
```
window.addEventListener('message', function (e) {
  if (e.data.message && e.data.message.action === "accountsChanged") {
    // handler logic
    console.log('got accountsChanged event', e)
  }
})
```
And when I changed the account in the extension, this event got logged:
```
{
  "isTrusted": true,
  "data": {
    "message": {
      "action": "accountsChanged",
      "data": { "address": "{address}" }
    }
  }
}
```
Your approach works well but, as you say it is not referenced in the original docs, maybe we can not rely on that. What do you think?

llawliet-l-l: My bad, fixed in https://github.com/rango-exchange/rango-client/pull/1439/commits/2a1bbaedb0a63a2043f7f98db42c7253d9a07b03

wallets/provider-okx/src/builders/tron.ts
```diff
@@ -40,29 +17,40 @@ export const changeAccountSubscriber = (getInstance: () => ProviderAPI) =>
...
+      /*
+       * Change account only on `accountsChanged` with a string address (`format`
+       * reads it). Everything else — including `accountsChanged` with address
+       * `false` (its disconnect is handled above) — is ignored.
+       */
       if (
-        !isOkxTronWalletChangedEvent(data) ||
-        data.data.data.params.length === 0
+        action !== 'accountsChanged' ||
+        typeof data.message.data?.address !== 'string'
```
RyukTheCoder, line 40: Based on your comment, shouldn't you disconnect when `action ===  "accountsChanged" and `typeof data.message.data?.address !== 'string'`?

wallets/provider-okx/src/builders/tron.ts
```diff
@@ -40,29 +17,40 @@ export const changeAccountSubscriber = (getInstance: () => ProviderAPI) =>
...
-        return utils.formatAccountsToCAIP([]);
+        event.preventDefault();
       }
-      const address = findTronAddress(data.data.data.params[0]);
-      return utils.formatAccountsToCAIP(address ? [address] : []);
+    })
+    .format(async (_, event) => {
+      const data = event.data;
+      const address =
+        isOkxTronMessageEvent(data) && data.message.action === 'accountsChanged'
```
RyukTheCoder, line 48: validation is being done in `onSwitchAccount`. Is it needed to revalidate here?

wallets/provider-okx/src/builders/tron.ts
```diff
@@ -40,29 +17,40 @@ export const changeAccountSubscriber = (getInstance: () => ProviderAPI) =>
...
+          ? data.message.data?.address
+          : undefined;
+      return utils.formatAccountsToCAIP(
+        typeof address === 'string' ? [address] : []
```
RyukTheCoder: When you set account as `[]`, I think the namespace stay connected but the accounts gets empty, is this your desired behavior?

wallets/provider-okx/src/types.ts
```diff
@@ -21,77 +21,28 @@ export type Provider = Map<
...
+  message: {
+    action: string;
+    data?: {
+      address: string | false;
     };
   };
 };
 
-export function isOkxTronStreamEvent(
+export function isOkxTronMessageEvent(
```
RyukTheCoder: This is a util not a type

wallets/provider-okx/src/actions/tron.ts
```diff
@@ -0,0 +1,37 @@
+import { utils } from '@rango-dev/wallets-core/namespaces/tron';
+
+import {
+  TRON_OK_REQUEST_CODE,
+  TRON_USER_REJECTION_CODE,
+} from '../constants.js';
+import { tronOKX } from '../utils.js';
+
+const connectAction = async () => {
```
RyukTheCoder:
```suggestion
const connect = async () => {
```

RyukTheCoder (changes requested):

wallets/provider-okx/src/builders/tron.ts
RyukTheCoder: Also check if we can have a validation to make sure the event is really account changed event from tron on okx wallet.

wallets/provider-okx/src/actions/tron.ts
```diff
@@ -0,0 +1,37 @@
...
+
+  if (accountsResult?.code && accountsResult.code !== TRON_OK_REQUEST_CODE) {
+    if (accountsResult.code === TRON_USER_REJECTION_CODE) {
+      throw new Error('User rejected the request.');
+    }
+    throw new Error(
+      accountsResult.message ?? 'Failed to connect to OKX Wallet Tron.'
+    );
+  }
+  return utils.formatAccountsToCAIP([instance.tronWeb.defaultAddress.base58]);
```
RyukTheCoder, line 23: NIT: in other providers format is happening in an after hook.

RyukTheCoder (changes requested):

wallets/provider-okx/src/builders/tron.ts
llawliet-l-l: I've handled wallet disconnection via the `disconnect` event returned by the wallet. We could also call the `disconnect` action as you said, but it would be redundant and call it twice.

I couldn't find a way to distinguish OKX `accountsChanged` events from those of other wallets.

wallets/provider-okx/src/builders/tron.ts
llawliet-l-l: I think it's better to be here for two reasons:
1. Type checking it without needing to assert any types.
2. `onSwitchAccount` is mostly in charge of detecting which events it can let through to reach the `format` function. How we detect the correct address in `format` should be separate and shouldn't rely on `onSwitchAccount` validation.

wallets/provider-okx/src/actions/tron.ts
llawliet-l-l: I've checked it, and we're formatting the address in the action in them. I couldn't find what you mentioned.

wallets/provider-okx/src/builders/tron.ts
llawliet-l-l: Fixed in https://github.com/rango-exchange/rango-client/pull/1439/commits/88adc44763acd699e6d09cbc32cdfa74ed2eb778

wallets/provider-okx/src/types.ts
llawliet-l-l: Fixed in https://github.com/rango-exchange/rango-client/pull/1439/commits/8f28be1e07bb96194b22ad39165f2ca81a2cf28f

wallets/provider-okx/src/actions/tron.ts
llawliet-l-l: Fixed in https://github.com/rango-exchange/rango-client/pull/1439/commits/058e0635ff4938f0225be058413ab896b17fc002

RyukTheCoder (approved):
LGTM

PR #1440, opened 2026-07-12
feat: add support for ledger wallet

RyukTheCoder:
# Summary

Adds a new wallet provider, @rango-dev/provider-ledger-wallet, integrating the [Ledger Wallet Provider](https://developers.ledger.com/docs/ledger-wallet-provider/overview) ("Ledger Button") into the widget. The provider is built on hub3 and connects over EIP-6963 / EIP-1193, letting users connect a Ledger device and sign/broadcast EVM transactions directly through Rango.

This is distinct from the existing provider-ledger package — it wraps Ledger's newer web provider rather than the low-level transport.

- New package wallets/provider-ledger-wallet (hub3 ProviderBuilder):
provider.ts — initializes @ledgerhq/ledger-wallet-provider, listens for the EIP-6963 announceProvider event, and stashes the announced EIP-1193 provider for the namespace/signer to use.
namespaces/evm.ts — EVM namespace (connect, disconnect, canEagerConnect, canSwitchNetwork, getChainId). connect requests accounts first because the Ledger provider rejects all other methods with "Unauthorized" until a session exists.
signers/evm.ts — CustomEvmSigner extends DefaultEvmSigner; overrides signAndSendTx (Ledger's eth_sendTransaction signs + broadcasts and returns only a hash) and wait (Ledger can't serve getTransaction/receipts, so status continues via the API watcher). Maps Rango's EvmTransaction to Ledger's JSON-RPC shape (gasLimit → gas, hex quantities, EIP-1559 preferred) and translates EIP-1193 user-rejection (4001) into a SignerError.
- Registration in provider-all (allProviders + dependency).
- New wallet type WalletTypes.LEDGER_WALLET = 'ledger-wallet' in wallets-shared.
- Widget config — new LedgerWalletConfig type (dAppIdentifier, apiKey, optional loggerLevel, hideButton) and wiring in the Wallets container.

Tested by connecting to ledger wallet and executing swaps on Ethereum, Polygon, BSC and bridges between them.

# Notes
Please notice that you should add `dAppIdentifier` and `apiKey` to the widget config to be able to test this.

# Checklist:


wallets/provider-ledger-wallet/src/signers/evm.ts
```diff
@@ -0,0 +1,142 @@
...
+/*
+ * DefaultEvmSigner.signAndSendTx requires a TransactionResponse, but Ledger's
+ * eth_sendTransaction only returns a hash string. Keep the result type
+ * assignable and leave response empty — wait doesn't use it, and
+ * updateStorageOnSuccessfulSign only stores response when truthy.
+ */
+type EvmSignAndSendResult = Awaited<
+  ReturnType<DefaultEvmSigner['signAndSendTx']>
+>;
+const NO_RESPONSE = undefined as unknown as EvmSignAndSendResult['response'];
```
arlert-armin: `NO_RESPONSE` casts `undefined` into a `TransactionResponse` — it lies to the type system. We don't need to: `GenericSigner.signAndSendTx` already declares `response?` as optional. The cast only exists because we `extends DefaultEvmSigner`, whose concrete return type is stricter (`response: TransactionResponse`, non-optional) than the interface it implements, and overrides are covariant so we inherit the stricter shape.

Two clean options:

1. **Use `GenericSigner`** — switch this class from `extends DefaultEvmSigner` to `implements GenericSigner<EvmTransaction>` and delegate `signMessage`/`signTypedData` to an internal `DefaultEvmSigner`. Then `return { hash }` with no cast.
2. **Or make the base optional** — change `DefaultEvmSigner.signAndSendTx` to return `response?: TransactionResponse` (aligning it with the interface), then drop the cast here.

Either is safe: consumers already treat `response` as optional — `updateStorageOnSuccessfulSign` types it `response?: any` and guards `if (signResult.response)`, and several sibling signers already return `{ hash }` only.

arlert-armin:
**Bluetooth mode doesn't work — needs to be investigated and fixed before merge.**

One confirmed cause for the **iframe SDK integration** path: the widget `<iframe>` is created without an `allow` attribute (`widget/iframe/src/main.ts`). Web Bluetooth / WebHID / WebUSB are Permissions-Policy-gated and disabled by default in cross-origin iframes, so Ledger's transport throws (`SecurityError … disallowed by permissions policy`). Fix is to delegate the features: `widget.setAttribute('allow', 'usb; bluetooth; hid')`.

Note: that gate does **not** apply to the top-level `widget:dev` app (no cross-origin iframe, secure localhost context). So if Bluetooth also fails there, it's a separate issue — please capture the exact console error so we can pin the root cause rather than guess.

arlert-armin:
Should we also remove the existing `provider-ledger` now that we're adding `provider-ledger-wallet`? Maintaining two Ledger providers isn't ideal.

Caveat before we do: they're **not** equivalent yet.
- `provider-ledger` (old): EVM **+ Solana**, direct device transport, with manual derivation-path selection (MetaMask / LedgerLive / Legacy / Solana paths).
- `provider-ledger-wallet` (new): **EVM-only**, via Ledger's Button (EIP-6963/1193), account selection delegated to Ledger's UI.

So removing the old one would drop Solana Ledger support and path selection. Suggest we treat consolidation as a separate task once the new provider reaches parity (or we consciously decide to drop Solana-on-Ledger), rather than removing it in this PR. Flagging for a decision.

wallets/provider-ledger-wallet/src/signers/evm.ts
RyukTheCoder: Fixed using GenericSigner

RyukTheCoder:
> **Bluetooth mode doesn't work — needs to be investigated and fixed before merge.**
> 
> One confirmed cause for the **iframe SDK integration** path: the widget `<iframe>` is created without an `allow` attribute (`widget/iframe/src/main.ts`). Web Bluetooth / WebHID / WebUSB are Permissions-Policy-gated and disabled by default in cross-origin iframes, so Ledger's transport throws (`SecurityError … disallowed by permissions policy`). Fix is to delegate the features: `widget.setAttribute('allow', 'usb; bluetooth; hid')`.
> 
> Note: that gate does **not** apply to the top-level `widget:dev` app (no cross-origin iframe, secure localhost context). So if Bluetooth also fails there, it's a separate issue — please capture the exact console error so we can pin the root cause rather than guess.

Bluetooth mode not working is not related to our code. The case you reported on iframe is not valid as we are testing on app not iframe. The root cause is that discovery drops the GATT link and connect() doesn't re-establish it, so service
   discovery runs against a dead link. This is already reported to Ledger team.
   
Update: It seems it works on mac

arlert-armin (approved):
LGTM

PR #1442, opened 2026-07-14
feat(approval): migrate EVM and Tron approval to the prerequisites flow

arlert-armin:
# Summary

Migrates EVM and Tron token approvals from the legacy flow (`isApprovalTx` on the create-transaction response + polling the server `check-approval` endpoint) to the unified **prerequisites** pattern already used for XRPL/Stellar trustlines. When the client opts in (`validations.approve: false`), the backend returns the approval requirement in the transaction's `prerequisites` list, and the client checks the on-chain allowance, builds/signs the approve itself, and waits for confirmation through the wallet's namespace — no server round-trips for approval.

### Key decisions

- **Per-wallet opt-in, fully backward compatible.** A wallet uses the new flow only if its hub namespace exposes the approval read action. Everything else — WalletConnect (no hub namespace) and any not-yet-migrated wallet — stays on the untouched legacy flow. No wallet is forced onto the new path.
- **One generic action, per-chain adapters.** EVM and Tron share a single `checkApprove` / `checkApproveTransactionStatus` pair; each chain supplies a small adapter (build-tx, status-check, namespace key). Avoids duplicating ~85% of the logic and lets other chains slot in later.
- **EVM:** injected wallets read via their EIP-1193 provider; **Ledger/Trezor** (no injected provider) read via a JSON-RPC instance shared with their existing signer.
- **Tron:** the approve tx is built client-side via the wallet's TronWeb (the Tron signer needs a fully-built tx); supported on bitget and tron-link.

### Dependent PRs

This branch pins **experimental** builds of two dependencies while their PRs are in review. Both must merge and publish stable versions before this can:

- rango-exchange/rango-sdk#84 — `rango-types`: new `EVM_APPROVE` / `TRON_APPROVE` prerequisite types
- rango-exchange/hub3js#21 — `@hub3js/evm`: new `allowance` / `getTransactionReceipt` actions

> Note: the `rango-types` / `@hub3js/evm` pins in this diff are the experimental versions; they will be swapped for the stable releases once the two PRs above land.

- [ ] EVM swap needing approval on an injected wallet (Phantom) — allowance-skip + full approve path
- [ ] Tron swap needing approval on bitget / tron-link
- [ ] WalletConnect / not-yet-migrated wallet — still uses the legacy approval flow
- [ ] `build:widget` green

# Checklist:


queue-manager/rango-preset/src/actions/checkPrerequisites/utils.ts
```diff
@@ -86,6 +86,44 @@ export function handleUnmetXrplChangeTrustLinePrerequisite(
...
+  });
+
+  if (!prerequisiteResult) {
+    return new Ok(prerequisiteInfo.checkAction);
+  }
+
+  switch (prerequisiteResult.status) {
+    case 'pending':
+      return new Ok(prerequisiteInfo.checkStatusAction);
+    case 'success': // Unreachable code
```
RyukTheCoder, line 115: throw error

queue-manager/rango-preset/src/actions/checkApprove/checkApprove.ts
```diff
@@ -0,0 +1,248 @@
...
+  } | null = null;
+
+  for (
+    let prerequisiteIndex = 0;
+    prerequisiteIndex < currentTransactionFromStorage.prerequisites.length;
+    prerequisiteIndex++
+  ) {
+    const prerequisite =
+      currentTransactionFromStorage.prerequisites[prerequisiteIndex];
+    if (adapter.isPrerequisite(prerequisite)) {
```
RyukTheCoder: I find this vague. Maybe renaming to `approveAdapter` and `isApprovePrerequisite` help with that.

queue-manager/rango-preset/src/actions/checkApprove/checkApprove.ts
```diff
@@ -0,0 +1,248 @@
...
+   * 3. Ensure wallet is connected and is on the correct network.
+   *
+   */
+  const checkResult = await checkEnvironmentBeforeExecuteTransaction(actions);
+  if (checkResult.err) {
+    requestBlockQueue(actions, checkResult.val);
+    return;
+  }
+
+  resetNetworkStatus(actions);
```
RyukTheCoder, line 146: Why is this needed here after environment is checked and it is not blocked?

queue-manager/rango-preset/src/actions/checkApprove/utils.ts
```diff
@@ -0,0 +1,60 @@
...
+>(
+  context: SwapQueueContext,
+  walletType: WalletType,
+  namespaceKey: keyof DefaultNamespaces
+): Result<TNamespace, null> {
+  try {
+    // `hubProvider` throws for wallets that are not migrated to the hub.
+    const namespace = context.hubProvider(walletType)?.get(namespaceKey);
+    if (namespace && 'allowance' in namespace) {
+      return new Ok(namespace as unknown as TNamespace);
```
RyukTheCoder: can you remove this cast?

queue-manager/rango-preset/src/actions/checkApprove/checkApprove.ts
RyukTheCoder: Also apply this naming improvements in `checkApproveTransactionStatus`

queue-manager/rango-preset/src/actions/checkApprove/types.ts
```diff
@@ -0,0 +1,75 @@
...
+  TTransaction extends Transaction
+> {
+  /** The prerequisite type this adapter handles, e.g. `EVM_APPROVE`. */
+  prerequisiteType: TransactionPrerequisiteType;
+  /** The hub namespace key backing this chain, e.g. `'evm'` | `'tron'`. */
+  namespaceKey: keyof DefaultNamespaces;
+  /** The signer transaction type used to sign the approve tx. */
+  signerTxType: TransactionType;
+  /** Type guard for this chain's approve prerequisite. */
+  isPrerequisite: (
```
RyukTheCoder:
```suggestion
  isApprovePrerequisite: (
```

queue-manager/rango-preset/src/actions/checkApprove/adapters/evm.ts
```diff
@@ -0,0 +1,70 @@
...
+  namespaceKey: 'evm',
+  signerTxType: TransactionType.EVM,
+
+  isPrerequisite: isEvmApprovePrerequisite,
+
+  isPrerequisiteResult: (
+    prerequisiteResult
+  ): prerequisiteResult is ExecutedApproveResult =>
+    isEvmApprovePrerequisiteResult(prerequisiteResult) &&
+    prerequisiteResult.status !== 'skipped',
```
RyukTheCoder: isPrerequisiteResult is skipping skipped prerequisiteResults, it is not what this method is intended to do

queue-manager/rango-preset/src/actions/checkApprove/types.ts
```diff
@@ -0,0 +1,75 @@
...
+import type {
+  TransactionPrerequisite,
+  TransactionPrerequisiteResult,
+  TransactionPrerequisiteType,
+  TransactionType,
+} from 'rango-types';
+import type { Result } from 'ts-results';
+
+/** The allowance-reading capability shared by every approve-capable namespace. */
+export interface ApproveCapableNamespace {
```
RyukTheCoder: Shouldn't this extend Namespace? Right now its only an object with allowance method not a namespace

queue-manager/rango-preset/src/actions/checkApprove/types.ts
```diff
@@ -0,0 +1,75 @@
...
+}
+
+/**
+ * The approve prerequisites (`EvmApprovePrerequisite | TronApprovePrerequisite`)
+ * — the members of the union that carry a `spender`.
+ */
+export type ApprovePrerequisite = Extract<
+  TransactionPrerequisite,
+  { spender: string }
+>;
```
RyukTheCoder:
```suggestion
export type ApprovePrerequisite = EvmApprovePrerequisite | TronApprovePrerequisite
```

queue-manager/rango-preset/src/actions/checkApprove/utils.ts
```diff
@@ -0,0 +1,60 @@
...
+    }
+    return new Err(null);
+  } catch {
+    return new Err(null);
+  }
+}
+
+export async function isAllowanceSufficient(
+  prerequisite: ApprovePrerequisite,
+  options: { namespace: ApproveCapableNamespace }
```
RyukTheCoder: namespace is not an option. it is required for this function.

wallets/core/src/namespaces/tron/mod.ts
```diff
@@ -4,5 +4,14 @@ export * as and from './and.js';
 export * as before from './before.js';
 export * as builders from './builders.js';
 export * as utils from './utils.js';
-export type { ProviderAPI, TronActions } from './types.js';
-export { CAIP_NAMESPACE, CAIP_TRON_CHAIN_ID } from './constants.js';
+export type {
+  ProviderAPI,
+  TronActions,
+  TronAllowanceParams,
```
RyukTheCoder: Why do you need to export theses types and constant?

widget/app/src/App.tsx
```diff
@@ -41,6 +41,10 @@ export function App() {
         walletConnectProjectId: WC_PROJECT_ID,
         trezorManifest: TREZOR_MANIFEST,
         tonConnect: { manifestUrl: TON_CONNECT_MANIFEST_URL },
+        // TODO: remove once the approval prerequisites routing is no longer experimental on the backend.
+        routing: {
+          experimental: 'enabled',
```
RyukTheCoder, line 46: Can you explain what effect it has?

wallets/provider-ledger/src/signers/ethereum.ts
```diff
@@ -48,7 +46,19 @@ export class EthereumSigner implements GenericSigner<EvmTransaction> {
     chainId: string | null
   ): Promise<{ hash: string }> {
     try {
-      const provider = new JsonRpcProvider(DEFAULT_ETHEREUM_RPC_URL); // Provider to broadcast transaction
+      /*
+       * Ledger signs a raw transaction on-device and does not estimate gas, so
+       * a gas limit must already be present. Client-built transactions that
+       * leave it null (e.g. approve prerequisites) are not supported yet.
+       */
+      if (!tx.gasLimit) {
```
RyukTheCoder, line 49: As you have access to `JsonRpcProvider` you can call `eth_estimateGas` and fetch `gasLimit`.

wallets/provider-ledger/src/utils.ts
```diff
@@ -1,20 +1,57 @@
...
+    evmRpcProvider = new JsonRpcProvider(DEFAULT_ETHEREUM_RPC_URL);
+  }
+  return evmRpcProvider;
+}
+
+/**
+ * Wraps the shared JSON-RPC provider into the `request({ method, params })`
+ * shape the EVM namespace read actions expect.
+ */
+export function evmRpcInstance(): ProviderAPI {
```
RyukTheCoder: I do not like this approach, you are converting a JsonRpcProvider to an EIP1193Provider shape just to use the default action in hub3js/evm. I prefer writing a custom action here which fetches allowance using the available JsonRpcProvider.

wallets/core/src/namespaces/tron/types.ts
```diff
@@ -9,7 +9,107 @@ export interface TronActions
...
+  owner: string;
+  spender: string;
+  /** Amount to approve, as an integer string in the token's smallest unit. */
+  amount: string;
+  /** Fee limit in SUN; defaults to DEFAULT_APPROVE_FEE_LIMIT when omitted. */
+  feeLimit?: number;
+};
+
+/**
+ * The focused TronWeb surface these actions call. `tronweb` is not a repo
```
RyukTheCoder: Add reference to the definitions in tronweb package or docs

wallets/core/src/namespaces/tron/utils.ts
```diff
@@ -1,9 +1,26 @@
...
+/**
+ * Prerequisite addresses arrive in 0x-hex (EVM-style) form; TronWeb expects a
+ * Tron address. Convert `0x<20-bytes>` → `41<20-bytes>` → Base58.
+ */
+export function hexAddressToTronBase58(
+  tronWeb: TronWebApi,
+  address: string
+): string {
+  const tronHex =
+    TRON_HEX_ADDRESS_PREFIX + address.replace(/^0x/, '').toLowerCase();
```
RyukTheCoder: unnecessary toLowerCase

RyukTheCoder (changes requested):

queue-manager/rango-preset/src/actions/checkApprove/utils.ts
arlert-armin: This one I couldn't fully drop. `hubProvider().get(namespaceKey)` is keyed by `keyof DefaultNamespaces`, so it returns a union of every namespace; narrowing that union to the adapter's concrete `TNamespace` needs an assertion (TypeScript can't evaluate the keyed-namespace conditional over the generic). The runtime `'allowance' in namespace` guard right above is what makes it safe — I added a comment explaining that. Happy to explore a key-parameterized redesign if you'd prefer removing it entirely.

queue-manager/rango-preset/src/actions/checkApprove/checkApprove.ts
arlert-armin: It clears a leftover `networkStatus` after the env check passes — `checkEnvironmentBeforeExecuteTransaction` can set `NetworkChanged` but never clears it. This mirrors `executeTransaction`, which calls `resetNetworkStatus` in the same spot right after the env check.

queue-manager/rango-preset/src/actions/checkPrerequisites/utils.ts
arlert-armin: Left these returning `Err` to match the sibling `handleUnmetStellarChangeTrustLinePrerequisite`, which returns `Err` for the same already-met/unreachable states. Happy to switch both to `throw` if you'd prefer that across the board.

wallets/provider-ledger/src/signers/ethereum.ts
arlert-armin: The missing `gasLimit` here is a backend gap — the approve prerequisite transaction should arrive with gas populated, and I've reported it to the backend team. I'd rather guard it with an explicit error than estimate on the client for now; client-side estimation can be a follow-up once the backend side is settled.

widget/app/src/App.tsx
arlert-armin: It opts the playground into the backend's experimental routing, which is what makes the API return the approval `prerequisites` this PR consumes (instead of the legacy auto-generated approval). It's temporary — the `// TODO` marks it for removal once that routing is no longer experimental on the backend.

queue-manager/rango-preset/src/actions/checkApprove/checkApprove.ts
arlert-armin: Done — renamed to `approveAdapter` / `isApprovePrerequisite`.

arlert-armin: Done — applied the same renames in `checkApproveTransactionStatus`.

queue-manager/rango-preset/src/actions/checkApprove/types.ts
arlert-armin: Done.

queue-manager/rango-preset/src/actions/checkApprove/adapters/evm.ts
arlert-armin: Good catch — made it a pure type guard; the `status === 'pending'` check at the call site does the narrowing now.

queue-manager/rango-preset/src/actions/checkApprove/types.ts
arlert-armin: Made it `{ namespaceId; providerId; allowance }` so it carries the namespace identity — `ProxiedNamespace<{ allowance }>` can't be the constraint (it's invariant in its actions type).

queue-manager/rango-preset/src/actions/checkApprove/types.ts
arlert-armin: Done — switched to the explicit union.

queue-manager/rango-preset/src/actions/checkApprove/utils.ts
arlert-armin: Done — `namespace` is now a required positional param.

wallets/core/src/namespaces/tron/mod.ts
arlert-armin: Removed — they weren't used outside the namespace.

wallets/provider-ledger/src/utils.ts
arlert-armin: Done — Ledger/Trezor now read allowance + receipt via the JsonRpcProvider directly; dropped the EIP-1193 shim.

wallets/core/src/namespaces/tron/types.ts
arlert-armin: Done — added a `@see` to the TronWeb docs.

wallets/core/src/namespaces/tron/utils.ts
arlert-armin: Done — removed.

queue-manager/rango-preset/src/actions/checkPrerequisites/utils.ts
RyukTheCoder: My bad. I meant "success" case should return error as well

wallets/provider-trezor/src/utils.ts
```diff
@@ -0,0 +1,38 @@
...
+    evmRpcProvider = new JsonRpcProvider(DEFAULT_ETHEREUM_RPC_URL);
+  }
+  return evmRpcProvider;
+}
+
+/**
+ * Wraps the shared JSON-RPC provider into the `request({ method, params })`
+ * shape the EVM namespace read actions expect.
+ */
+export function evmRpcInstance(): ProviderAPI {
```
RyukTheCoder: Dead code

queue-manager/rango-preset/src/actions/checkApprove/utils.ts
RyukTheCoder: I think key-parameterized redesign can improve it. By defining these types you can remove the cast entirely:
```
export type ApproveCapableNamespaceId = {
  [K in keyof DefaultNamespaces]: DefaultNamespaces[K] extends {
    getAllowance: unknown;
  }
    ? K
    : never;
}[keyof DefaultNamespaces];
// type ApproveCapableNamespaceId = "evm" | "tron"

export type ApproveCapableNamespace<K extends ApproveCapableNamespaceId> =
  FindProxiedNamespace<K, DefaultNamespaces>;
```
Now your `ApproveAdapter` can use that:
```
export interface ApproveAdapter<
  K extends ApproveCapableNamespaceId,
  TTransaction extends Transaction
> {
  /** The prerequisite type this adapter handles, e.g. `EVM_APPROVE`. */
  prerequisiteType: TransactionPrerequisiteType;
  /** The hub namespace key backing this chain, e.g. `'evm'` | `'tron'`. */
  namespaceKey: K;
  /** The signer transaction type used to sign the approve tx. */
  signerTxType: TransactionType;
  /** Type guard for this chain's approve prerequisite. */
  isApprovePrerequisite: (
    prerequisite: TransactionPrerequisite
  ) => prerequisite is ApprovePrerequisite;
  /** Type guard for this chain's approve prerequisite result (any status). */
  isApprovePrerequisiteResult: (
    prerequisiteResult: TransactionPrerequisiteResult
  ) => prerequisiteResult is ApprovePrerequisiteResult;
  /** Builds a signable approve transaction for the given prerequisite. */
  buildApproveTransaction: (
    prerequisite: ApprovePrerequisite,
    namespace: ApproveCapableNamespace<K>
  ) => Promise<Result<TTransaction, NextTransactionStateError>>;
  /** Reads the on-chain status of a submitted approve transaction. */
  getTransactionStatus: (
    namespace: ApproveCapableNamespace<K>,
    executedTransactionHash: string
  ) => Promise<ApproveTransactionStatus>;
}
export const tronApproveAdapter: ApproveAdapter<'tron', TronTransaction>
export const evmApproveAdapter: ApproveAdapter<'evm', EvmTransaction>
```
And this cast gets removed entirely:
```
export function resolveApproveNamespace<K extends ApproveCapableNamespaceId>(
  context: SwapQueueContext,
  walletType: WalletType,
  namespaceId: K
): Result<ApproveCapableNamespace<K>, null> {
  try {
    // `hubProvider` throws for wallets that are not migrated to the hub.
    const namespace = context.hubProvider(walletType)?.get(namespaceId);
    if (namespace && 'getAllowance' in namespace) {
      return new Ok(namespace);
    }
    return new Err(null);
  } catch {
    return new Err(null);
  }
}
```
Now it is type safe and `resolveApproveNamespace` can only get called with "evm" and "tron" namespace ids.

queue-manager/rango-preset/src/actions/checkApprove/types.ts
RyukTheCoder: No need for `namespaceId; providerId;`, you can also use the below type as I mentioned in the key-parameterezied comment:
```
{
    getAllowance: unknown;
  }
```
Type of function will be inferred from namespace actions.

queue-manager/rango-preset/src/actions/createTransaction.ts
```diff
@@ -21,6 +24,59 @@ import {
...
+/**
+ * The prerequisites-based approval flow needs the wallet's hub namespace to
+ * expose the `getAllowance` action (for the allowance pre-check and the post-mining
+ * re-check). For EVM, injected wallets back it with their EIP-1193 provider and
+ * Ledger/Trezor with a JSON-RPC instance; for Tron, bitget/tron-link back it with
+ * the injected TronWeb. Only wallets whose namespace does not register the action
+ * stay on the legacy server-generated approval flow — e.g. WalletConnect, which
+ * has no hub namespace at all.
+ */
+function shouldUseApprovePrerequisite(
```
RyukTheCoder: NIT: create a `createTransaction` directory and move this file to it and make this function a util in "utils" file inside the new directory.

queue-manager/rango-preset/src/actions/createTransaction.ts
```diff
@@ -21,6 +24,59 @@ import {
...
+  [TransactionType.TRON]: 'tron',
+};
+
+/**
+ * The prerequisites-based approval flow needs the wallet's hub namespace to
+ * expose the `getAllowance` action (for the allowance pre-check and the post-mining
+ * re-check). For EVM, injected wallets back it with their EIP-1193 provider and
+ * Ledger/Trezor with a JSON-RPC instance; for Tron, bitget/tron-link back it with
+ * the injected TronWeb. Only wallets whose namespace does not register the action
+ * stay on the legacy server-generated approval flow — e.g. WalletConnect, which
```
RyukTheCoder: Mentioning wallet name is not a good approach. Having this, we should update this comment after WalletConnect gets migrated to  hub.

queue-manager/rango-preset/src/actions/createTransaction/createTransaction.ts
```diff
@@ -56,7 +112,11 @@ export async function createTransaction(
       validations: {
         balance: swap.validateBalanceOrFee,
         fee: swap.validateBalanceOrFee,
-        approve: true,
+        approve: !shouldUseApprovePrerequisite(
```
RyukTheCoder, line 61: To make sure that we do not rely on rango service to not returning approve prerequisites when we send this, I think we should add guards on checking approve prerequisites and see if we should really have received approve prerequisites and if we shouldn't, just skip those prerequisites or apply any other proper actions.

RyukTheCoder (changes requested):

wallets/provider-trezor/src/utils.ts
arlert-armin: Good catch — removed. I'd deleted it from Ledger but missed the Trezor copy.

queue-manager/rango-preset/src/actions/checkApprove/utils.ts
arlert-armin: Applied your key-parameterized design — `ApproveAdapter` is now keyed by `K extends ApproveCapableNamespaceId`, so `get(namespaceId)` returns the concrete `ApproveCapableNamespace<K>` and the cast is gone entirely. Thanks for working it out.

queue-manager/rango-preset/src/actions/checkApprove/types.ts
arlert-armin: Done as part of the key-parameterized redesign — `ApproveCapableNamespace<K> = FindProxiedNamespace<K, DefaultNamespaces>`, so no more `namespaceId; providerId;`.

queue-manager/rango-preset/src/actions/createTransaction.ts
arlert-armin: Reworded — the comment no longer names WalletConnect; it now just says wallets whose namespace doesn't register the action (or have no hub namespace) stay on the legacy flow.

queue-manager/rango-preset/src/actions/createTransaction.ts
arlert-armin: Done — moved into a `createTransaction/` directory with the action in `createTransaction.ts` and `shouldUseApprovePrerequisite` extracted into `utils.ts`.

queue-manager/rango-preset/src/actions/createTransaction/createTransaction.ts
arlert-armin: `checkApprove` already guards this: it resolves the wallet's namespace and, if the wallet isn't approve-capable (no `getAllowance` — same condition as the `shouldUseApprovePrerequisite` opt-in), it marks the prerequisite `skipped` and moves on instead of executing it. So an approve prerequisite the service shouldn't have returned is skipped rather than trusted. Happy to lift the check earlier (into checkPrerequisites) if you'd prefer it more explicit.

queue-manager/rango-preset/src/actions/checkPrerequisites/utils.ts
arlert-armin: Thanks for clarifying — the `success` (and `skipped`) case already returns an `Err` here, so this is covered.

RyukTheCoder (approved):
LGTM

PR #1444, opened 2026-07-20
feat(provider-okx): add sui namespace

llawliet-l-l:
## Summary

Adds Sui support to the OKX wallet provider. Sui is discovered through the wallet-standard registry (independent of OKX's injected instance map used by EVM/Solana/BTC), so it follows the wallet-standard pattern like the Slush/Phantom providers. Connect, account switch, wallet-side disconnect, eager connect, and transaction signing are all wired and verified on-device.

## Changes

- Wire the `Sui` namespace into the provider and register `DefaultSuiSigner` (conditionally, since Sui comes from the registry rather than the injected map)
- Add a `Sui` metadata entry and the wallet-standard name constant
- Account switches handled by the core `standard:events` `change` subscriber; add a dedicated subscriber for wallet-side disconnect, which OKX emits as an (undocumented) `disconnect` event on the same `standard:events` feature
- Add the `@rango-dev/signer-sui` dependency (Sui signer)
- Docs: mark OKX Sui ✅ in the status matrix and note the auto-connect limitation in the provider readme

## Notes

- OKX fires a `disconnect` event on `standard:events` that isn't in the `@mysten/wallet-standard` type (which only declares `change`), so the disconnect subscriber casts `.on` to accept the extra event name (`TS2345` otherwise). The cast can be dropped if the event is ever added to the official types.

## How tested

`eslint` + `yarn build` pass. Verified on-device with the OKX extension: connect, account switch, disconnect from the wallet UI, eager connect on reload, and a successful real Sui transaction signature.

## Known limitations

- On auto connect, OKX reconnects to the **first-connected** account rather than the last-switched one — no client-side workaround, as the wallet exposes no way to read the active account at reconnect time.


wallets/provider-okx/src/builders/sui.ts
```diff
@@ -0,0 +1,35 @@
...
+      (
+        instance.features['standard:events'].on as (
+          event: string,
+          listener: (event: unknown) => void
+        ) => () => void
+      )('disconnect', callback)
+    )
+    .removeEventListener(() => {
+      // `standard:events.on` returns an unsubscribe function, invoked by the builder cleanup.
+    });
```
arlert-armin: don't need to use changeAccountSubscriberBuilder for disconnectSubscriber

wallets/provider-okx/src/namespaces/sui.ts
```diff
@@ -0,0 +1,57 @@
...
+    actions.canEagerConnect({
+      name: WALLET_NAME_IN_WALLET_STANDARD,
+    })
+  )
+  .build();
+
+const connect = builders
+  .connect({
+    name: WALLET_NAME_IN_WALLET_STANDARD,
+  })
```
arlert-armin, line 33: how about picking first returned account of the connect to fix auto connect issue.

arlert-armin (changes requested):

wallets/provider-okx/src/namespaces/sui.ts
llawliet-l-l: I only see one account in the list even though I've already connected multiple accounts, and that account is the first one that was connected.

wallets/provider-okx/src/builders/sui.ts
llawliet-l-l: I used it because it was a more familiar and easy to understand code
fixed in https://github.com/rango-exchange/rango-client/pull/1444/commits/1ef3cf1110542618ea0a1a3a69e6b44e8bfda270

wallets/provider-okx/src/hooks/sui.ts
arlert-armin, line 1: now it's not a builder it's just a hook, put it in hooks please

llawliet-l-l: Fixed in https://github.com/rango-exchange/rango-client/pull/1444/commits/8f0cf35e0fb1a812d75a2d473e32cf76c92a1af6

arlert-armin (approved):
LGTM

PR #1448, opened 2026-07-26
build: use rangutopia CLI to build library packages

llawliet-l-l:
## Summary

Replaces the in-repo `scripts/build/command.mjs` with `rangutopia library build`, so library builds are maintained in one place across our repos instead of being copy-pasted per monorepo. Output is byte-identical apart from one package (see caveat), and the removals below are dead code that the migration leaves behind.

## Changes

- Migrate 55 library packages' `build` script to `yarn run rangutopia library build`, and pin `rangutopia` in root devDependencies
- Drop the now-redundant `--inputs src/mod.ts` from 33 packages — rangutopia defaults to `src/index.ts` then `src/mod.ts`, and none of the 33 has an `index.ts`, so the resolved entry point is unchanged. `wallets/core` (14 entries) and `wallets/react` keep their explicit `--inputs`
- Delete `scripts/build/command.mjs`
- Remove the `analyze-bundle` CI job, `scripts/analyze-bundle/`, and `BUILD_META_FILE_SUFFIX`. `rangutopia library build` doesn't write the `dist/*.build.json` metafiles the job reads. It was already inert before this PR — on recent runs it only ever printed:

      Target Branch Artifact not found
      Directories for the target or current branch do not exist, the bundle size comparison will be skipped.

## Follow-up (not in this PR)

`.github/workflows/upload-build-artifacts.yml` can probably go too. It's `workflow_dispatch`-only and exists to seed the artifacts the bundle comparison downloaded — and checking GitHub, it has never been invoked once. Left in place here to keep this PR focused; happy to drop it separately. Its composite action `.github/actions/upload-build-artifacts` must stay regardless, since the `build-libs` job uses it.

## Note for reviewers

`esbuild` and `esbuild-plugins-node-modules-polyfill` must stay in root devDependencies even though the script that imported them is gone. rangutopia declares esbuild as a *devDependency* and ships no nested copy, so it resolves both from this repo at build time — that's also why output still matches (esbuild 0.20.1, not rangutopia's 0.25.4). Removing them fails at `library check`.

## How tested

`yarn clean` → `yarn install` → `yarn build:all`, 59 projects, exit 0. Against the pre-migration baseline: **3039 files identical, 0 declaration differences, +0 bytes**.

The only exception is `wallets/provider-ledger`. rangutopia passes `modules: { fs: true }` to the polyfill plugin where this repo passed `globals: { fs: true }`; the latter is an inert key that left the plugin on its default of polyfilling every builtin, while the former *narrows* it to `fs` alone. `fs` isn't imported by provider-ledger at all — what actually changes is `events` (jspm shim → npm `events`), worth 117 bytes. `widget/app`, `widget/playground` and `queue-manager-demo` shift only because they bundle ledger code, so their content hash and `index.html` script tag move.

# Checklist:


arlert-armin (approved):
LGTM

PR #1449, opened 2026-07-26
refactor: declare chain support by CAIP-2 chain id

RyukTheCoder:
# Summary

Providers used to declare their supported chains by filtering a list of
`BlockchainMeta` keyed on Rango blockchain names:

```ts
getSupportedChains: (chains: BlockchainMeta[]) => BlockchainMeta[]
```

This conflated two responsibilities — deciding whether a chain is supported, and
building the list of supported chains — and made every provider reason in Rango
names while the hub keys everything by CAIP-2. Providers pulled in `rango-types`
helpers (`evmBlockchains`, `solanaBlockchain`, `tronBlockchain`, …) purely to
answer a yes/no question, and name-keyed lookups are what made Ctrl's per-chain
provider-instance lookup miss.

This PR replaces the hook with a predicate over a CAIP-2 chain id:

```ts
isChainSupported: (chainId: string) => boolean
```

**What changed**

- `NamespaceMeta.getSupportedChains` → `isChainSupported` in
  `wallets-core/legacy`, and across all 33 provider `constants.ts` files.
- New `wallets/shared/src/blockchains.ts` — the single bridge between Rango meta
  and CAIP-2:
  - `convertBlockchainMetaToCaip`, resolving per `TransactionType`. The resolver
    map is a total `Record`, so adding a `TransactionType` in `rango-types`
    fails to compile here instead of silently making that family invisible to
    every wallet.
  - `CAIP_NAMESPACES` / `CAIP_CHAINS`, pairing each namespace with its reference
    in one place rather than at call sites.
  - `parseCaip2ChainId` + namespace/reference accessors, built on `ChainId.parse`
    so a malformed id matches nothing. Rejecting matters more than parsing here:
    a lenient split would let a bare `'eip155'` match every EVM wallet.
  - `isBitcoinBlockchain` / `isZcashBlockchain`, for the 7 signers that receive
    `asset.blockchain` (`'BTC'`) rather than a chain id.
- Building the `BlockchainMeta` list moves to `wallets-react` as
  `getSupportedChainsFromNamespace` / `getSupportedChainsFromProvider`, now
  exported from the package root. `NamespaceItem`, `Namespaces`, and the
  playground's `getWalletsList` use those instead of reaching into the metadata
  property themselves.
- Ctrl's UTXO instance map is keyed by fully-qualified CAIP-2 id
  (`bip122:0000…e93`), with the bare bip122 reference kept alongside for
  `formatAccountsToCAIP`, which prepends the namespace itself. Collapsing the two
  representations is what broke the lookup previously.

**Notes for reviewers**

- Several families report a meta `chainId` that is *not* their CAIP reference
  (Solana `mainnet-beta`, Sui `sui-mainnet`, Starknet `SN_MAIN`), so those map to
  fixed ids and ignore `chainId`. EVM is the opposite: meta reports hex
  (`0xa86a`) and `eip155` references are decimal, so the prefix picks the radix.
- `HYPERLIQUID` deliberately lands in `eip155`, using meta's `1337` rather than
  HyperEVM's real `999`. Safe only because an `eip155` reference is never turned
  back into a chain id — see the comment before deriving a network switch or RPC
  target from it.
- `COSMOS` maps to `null` on purpose: we don't connect to it through the hub, so
  no namespace claims it. `DASH` likewise has no bip122 constant and no provider
  support, so it drops out.

**Behaviour change:** `provider-ctrl`, `provider-default` and `provider-safe`
previously used `evmBlockchains()` *without* `hyperliquidBlockchain()`. Their
predicate is now a bare `eip155` namespace check, so they report Hyperliquid
support. The other twelve EVM providers already included it and are unchanged.
Flagging in case any of the three should keep excluding it.

Fixes # (issue)

- [x] `npx vitest run wallets/shared/tests/blockchains.test.ts` — 29 new tests
      covering hex/decimal EVM conversion, malformed-`chainId` rejection, the
      fixed-id single-chain families, `COSMOS`/`DASH` returning `null`, the
      Hyperliquid namespace guarantee, `TRANSFER` name resolution, and CAIP-2
      parse rejection (bare namespace, empty reference, extra segments).
- [x] `nx run-many --target=build` for `wallets-core`, `wallets-shared`,
      `wallets-react`, `provider-all`, `widget-embedded` — passes. The build runs
      `tsc --declaration --emitDeclarationOnly`, so this typechecks all 33
      rewritten providers and both widget call sites.
- [x] `eslint` over all 32 changed provider packages — clean. Remaining lint
      failures in `wallets-core` / `wallets-react` / `wallets-shared` /
      `widget-embedded` / `widget-playground` are pre-existing, in files this PR
      doesn't touch.
- [x] Grep audit: no `getSupportedChains`, `OKX_WALLET_SUPPORTED_CHAINS`, or
      `SUPPORTED_UTXO_CHAINS` references remain; every provider that newly
      imports `@rango-dev/wallets-shared` declares it as a dependency.
- [x] Manual: connect Ctrl and confirm BTC/LTC/DOGE/BCH each resolve their own
      instance and sign; connect Phantom/Unisat/Xverse/Trezor/Bitget/OKX and
      confirm the BTC signer still accepts BTC and rejects other chains.
- [x] Manual: confirm the namespace chain lists in the wallet-connect modal and
      the playground wallet list render the same chains as before, per wallet.

# Checklist:


wallets/provider-binance/src/constants.ts
```diff
@@ -1,29 +1,50 @@
...
+  CAIP_OPTIMISM_CHAIN_ID,
+  CAIP_SCROLL_CHAIN_ID,
+  CAIP_SONIC_CHAIN_ID,
+  CAIP_ZETA_CHAIN_ID,
+  CAIP_NAMESPACE as EVM_CAIP_NAMESPACE,
+} from '@hub3js/evm';
+import {
+  getChainIdFromCaip2ChainId,
+  getNamespaceFromCaip2ChainId,
+} from '@rango-dev/wallets-shared';
```
arlert-armin: I'm not sure if the `wallets-shared` is the best place for these two since we are deciding to retire it.

wallets/provider-bitget/src/constants.ts
```diff
@@ -34,27 +39,24 @@ export const metadata: ProviderMetadata = {
...
           },
           {
             label: 'Tron',
             value: 'Tron',
             id: 'TRON',
-            getSupportedChains: (allBlockchains: BlockchainMeta[]) =>
-              tronBlockchain(allBlockchains),
+            isChainSupported: (chainId: string) =>
+              getNamespaceFromCaip2ChainId(chainId) === TRON_CAIP_NAMESPACE &&
+              getChainIdFromCaip2ChainId(chainId) === CAIP_TRON_CHAIN_ID,
```
arlert-armin: Why do we need this extra check?

wallets/provider-bitget/src/constants.ts
```diff
@@ -34,27 +39,24 @@ export const metadata: ProviderMetadata = {
...
             value: 'UTXO',
             id: 'BTC',
-            getSupportedChains: (allBlockchains: BlockchainMeta[]) =>
-              allBlockchains.filter(
-                (chain): chain is TransferBlockchainMeta =>
-                  chain.name === LegacyNetworks.BTC
-              ),
+            isChainSupported: (chainId: string) =>
+              getNamespaceFromCaip2ChainId(chainId) === UTXO_CAIP_NAMESPACE &&
+              getChainIdFromCaip2ChainId(chainId) === CAIP_BITCOIN_CHAIN_ID,
```
arlert-armin: Can't we introduce a util in their namespace to avoid doing the same thing for all wallets? Just checking expectional ones for exceptional wallets.

wallets/shared/src/blockchains.ts
arlert-armin, line 1: We want to retire this file let's have a solution for each namespace on its util maybe separately for each namespace.

arlert-armin (changes requested):

wallets/provider-binance/src/constants.ts
RyukTheCoder: moved to `@hub3js/std`

wallets/provider-bitget/src/constants.ts
RyukTheCoder: removed chain check for single chain namespaces

wallets/provider-bitget/src/constants.ts
RyukTheCoder: utils added to hub3js and wallets-core namespace packages

wallets/shared/src/blockchains.ts
RyukTheCoder: This file will remain here as a single point of converting `BlockchainMeta` to `CaipChainId`.

wallets/core/src/namespaces/solana/utils.ts
```diff
@@ -18,7 +16,7 @@ export async function getAccounts(provider: ProviderAPI) {
     : provider.publicKey.toString();
   return {
     accounts: [account],
-    chainId: LegacyNetworks.SOLANA,
+    chainId: 'SOLANA',
```
arlert-armin, line 19: Do we need this change? We are putting this library away.

arlert-armin (approved):
Thanks for all you efforts on this. 
We need to decide a better place for the networks later on.
Right now it LGTM

wallets/core/src/namespaces/solana/utils.ts
RyukTheCoder: just removing LegacyNetworks usage. nothing important.

PR #1452, opened 2026-07-28
Calculate total payable fee by using feesGroup

schrodingers-fox:
## Summary

Fixed a bug where the **Total Payable Fee** row in the quote cost details rendered a generic fee prop instead of calculating the true sum of all payable fees for the route.

### Changes Included

* **Fee Calculation**: Computed `totalPayableFee` by reducing over `feesGroup.payable` and accumulating each group's value in USD via `getTotalFeesInUsd`.
* **Import Fix**: Updated the `BigNumber` import from a type-only import (`import type { BigNumber }`) to a value import since it is now instantiated (`new BigNumber(0)`) for the reducer's initial value.

---

## How Was This Tested?

Verified in the widget UI by opening the quote cost details and confirming that the rendered **Total Payable Fee** matches the manual sum of individual payable fee rows across multi-step routes.

* [x] Verified the Total Payable Fee equals the sum of all payable fee entries for a single-step route
* [x] Verified the Total Payable Fee equals the sum of all payable fee entries for a multi-step route
* [x] Verified non-payable fees are excluded from the total and still render separately

---

## Checklist

### DIFF
--- a/widget/embedded/src/components/Quote/QuoteCostDetails.tsx
+++ b/widget/embedded/src/components/Quote/QuoteCostDetails.tsx
@@ -1,6 +1,5 @@
 import type { QuoteCostDetailsProps } from './Quote.types';
 import type { NameOfFees } from '../../constants/quote';
-import type BigNumber from 'bignumber.js';
 
 import { i18n } from '@lingui/core';
 import {
@@ -12,6 +11,7 @@ import {
   QuoteCost,
   Typography,
 } from '@rango-dev/ui';
+import BigNumber from 'bignumber.js';
 import React, { useState } from 'react';
 
 import { getFeeLabel } from '../../constants/quote';
@@ -62,7 +62,6 @@ export function QuoteCostDetails(props: QuoteCostDetailsProps) {
   const {
     steps,
     quote,
-    fee,
     time,
     feeWarning,
     timeWarning,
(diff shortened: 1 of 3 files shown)


widget/embedded/src/components/Quote/QuoteCostDetails.tsx
```diff
@@ -79,6 +79,19 @@ export function QuoteCostDetails(props: QuoteCostDetailsProps) {
...
+   * (feesGroup.payable), not the generic `fee` prop. We sum the USD value of
+   * every payable fee group so the total always matches the itemized rows above.
+   */
+  const totalPayableFee = Object.entries(feesGroup.payable).reduce(
+    (acc: BigNumber, [_, fees]) => {
+      const totalFeeInUsd = getTotalFeesInUsd(fees);
+      return acc.plus(totalFeeInUsd);
+    },
+    new BigNumber(0)
+  );
```
Ikari-Shinji-re:
```suggestion
  const totalPayableFee = Object.values(feesGroup.payable).reduce(
    (acc: BigNumber, fees) => acc.plus(getTotalFeesInUsd(fees)),
    new BigNumber(0)
  );
```

widget/embedded/src/components/Quote/QuoteCostDetails.tsx
Ikari-Shinji-re, line 1: Currently, this component is still using the fee received through props. Please check its usage. I think we can remove that prop and use `totalPayableFee` instead to prevent inconsistencies between the fee values.

widget/embedded/src/components/Quote/QuoteCostDetails.tsx
schrodingers-fox: Done — switched to `Object.values(...)` and dropped the unused destructuring. Thanks!

widget/embedded/src/components/Quote/QuoteCostDetails.tsx
schrodingers-fox: Good catch. Removed the fee prop from `QuoteCostDetails` (and its type/call sites) and now derive the collapsed summary from `totalPayableFee` too, so the summary and modal always show the same value.

I updated the comment I previously added for the code snippet description.

Ikari-Shinji-re (approved):

PR #1459, opened 2026-08-09
Refactor/rc 361 migrate to hub3js bip122

llawliet-l-l:
## Summary

Moves every UTXO import off `@rango-dev/wallets-core/namespaces/utxo` and onto the extracted `@hub3js/bip122` package. The namespace now represents all of its chains instead of assuming Bitcoin. each wallet's quirks are preserved as explicit overrides.

## Changes

- Adopt the package's shared `connect` / `canEagerConnect` / account-change subscriber, deleting the hand-rolled per-wallet copies; each wallet keeps only its delta (Xverse's payment-purpose filter, OKX's null-payload quirk, Bitget's singular `accountChanged`, Ctrl's EVM-driven multi-chain subscriber, Trezor's derivation path)
- Register `getConnectableNetworks` on each UTXO namespace so a wallet can report the chains it connects
- Zcash's CAIP-2 reference becomes its genesis-hash prefix rather than the placeholder `zcash`; nothing persists it, so no migration is needed

# Checklist:


arlert-armin (approved):
LGTM

PR #1468, opened 2026-08-22
Refactor/rc 395 remove dynamic import from wallets shared

llawliet-l-l:
**You should only review the last commit. Other changes are related to the rebase with the Walletconnect branch.**

## Summary

`dynamicImportWithRefinedError` lived in `@rango-dev/wallets-shared` but has nothing wallet-specific about it — it just runs a lazy import and rethrows failures with a user-friendly message. This adds a domain-neutral `@rango-dev/common-core` package and moves it there.

## Changes

- Add `common/core` (`@rango-dev/common-core`) and register `common/*` as a workspace group
- Repoint 34 files across 28 provider packages to the new import
- Drop the now-unused `@rango-dev/wallets-shared` dependency from `provider-slush`

## How tested

`build:all` (54/54 buildable projects), `yarn test` (17 files, 151 passed), and lint across all touched packages.

# Checklist:

### DIFF
--- a/widget/embedded/src/utils/swap.ts
+++ b/widget/embedded/src/utils/swap.ts
@@ -49,6 +49,7 @@ import {
   TOKEN_AMOUNT_MIN_DECIMALS,
 } from '../constants/routing';
 
+import { tryRefineError } from './errors';
 import { getBlockchainShortNameFor, isValidTokenAddress } from './meta';
 import { numberToString } from './numbers';
 import { getRequiredBalanceOfWallet } from './quote';
@@ -737,6 +738,12 @@ export function getSwapMessages(
   }
   detailedMessage = detailedMessage || '';
   message = message || '';
+  const refined = tryRefineError(message) ?? tryRefineError(detailedMessage);
+  if (refined) {
+    detailedMessage = [message, detailedMessage].filter(Boolean).join('\n');
+    message = refined.message;
+  }
+
   const isRpc =
     message?.indexOf('code') !== -1 && message?.indexOf('reason') !== -1;
 
--- a/wallets/provider-binance/src/signer.ts
+++ b/wallets/provider-binance/src/signer.ts
@@ -2,10 +2,7 @@ import type { Provider } from './types.js';
 import type { SignerFactory } from 'rango-types';
 
 import { EVM_NAMESPACE } from '@hub3js/namespaces';
-import {
-  dynamicImportWithRefinedError,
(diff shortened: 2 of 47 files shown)


yarn.lock
```diff
@@ -4798,6 +4798,18 @@
   dependencies:
     "@babel/runtime" "^7.13.10"
 
+"@rango-dev/wallets-core@^0.60.1-next.3":
```
RyukTheCoder: Should be removed, probably exists because of a wrong version in a package.json file.

widget/embedded/src/utils/swap.ts
RyukTheCoder, line 703: Refinement here gets discarded for any step that has a network status. The networkWarningState block at line 708 unconditionally reassigns both message and detailedMessage from networkStatusExtraMessage / networkStatusExtraMessageDetail, so whatever refineError produced above is overwritten and the user sees the raw error.

Two options: move the refine call below that block, or also run it over the networkStatus* fields. Since network-status messages are exactly the ones carrying connect/switch-network failures, they're the case that most needs refining.

RyukTheCoder (changes requested):

RyukTheCoder:
These are missed missed:
https://github.com/rango-exchange/rango-client/blob/5346fa3888ac6068bd36a70ec721faaa08ddfbaf/widget/embedded/src/components/WalletStatefulConnect/NamespaceDetachedItem.tsx#L41

https://github.com/rango-exchange/rango-client/blob/5346fa3888ac6068bd36a70ec721faaa08ddfbaf/widget/embedded/src/components/SwapDetailsModal/SwapDetailsModal.ConnectWallet.tsx#L91

yarn.lock
llawliet-l-l: Fixed in https://github.com/rango-exchange/rango-client/pull/1468/commits/66dbd63926bc6d4e65bb586cd52491fe62cb398f

widget/embedded/src/utils/swap.ts
llawliet-l-l: I think `refineError` should only accept the error(or its message) as an input and try to refine it. Network-status refinement is out of the function scope; it is a specific situation that needs multiple inputs, so I went with the first option. fixed in https://github.com/rango-exchange/rango-client/pull/1468/commits/8645090ce840da9cefb3f58fd86c0d0c3567e81d

RyukTheCoder (approved):

PR #1471, opened 2026-08-22
refactor(wallets): extract the Rango-to-CAIP chain mapping from wallets-shared

llawliet-l-l:
## Summary

`wallets-shared` is being deleted, so the chain-vocabulary table needs a home — but it is Rango domain logic that providers depend on, and providers are moving to hub3js, which cannot hold Rango logic. Splitting it in two puts that constraint in the dependency graph: provider code that only declares chain support depends on the CAIP half and stays movable, while signers depend on the Rango half and stay behind.

## Changes

- Move the mapping (`convertBlockchainMetaToCaip`, the transfer-name guards, `Networks`) to a new `@rango-dev/wallets-blockchains`; `wallets-shared` re-exports `Networks` for external consumers
- Move the general CAIP-2 registry (`CaipChainId`, `CAIP_NAMESPACES`, `CAIP_CHAINS`) to `@hub3js/caip` — it is standard chain knowledge with no Rango vocabulary
- Declare the `wallets-shared` dependency `provider-noir-wallet` was importing without listing, and switch walletconnect-2's UTXO signer to the shared `isBitcoinBlockchain` guard

## Notes

Signers keep reading Rango chain names, as a documented exception: `GenericSigner<Transfer>` comes from `rango-types`, so `tx.asset.blockchain` is a Rango name by construction and no signer can move to hub3js.

## How tested

`yarn build:all` 62/62 and root `vitest run` (17 files, 152 tests) pass. The mapping's 20 tests moved to the new package.

# Checklist:


wallets/blockchains/package.json
```diff
@@ -0,0 +1,36 @@
+{
+  "name": "@rango-dev/wallets-blockchains",
```
RyukTheCoder: why naming it wallets-blockchains when it does not have anything related to wallets?

RyukTheCoder (changes requested):
Please add an experimental hub3js version.

wallets/blockchains/package.json
llawliet-l-l: Because it depends on both the wallet's info (driven by hub3 standards) and Rango's blockchains, it was one of the short and precise names. What do you suggest?

RyukTheCoder: Didn't find any wallet info in this package please add a reference about what you mean. I find the purpose of this package as the translation layer for blockchains of Rango services to caip. So i suggest names like this:
"@rango-dev/blockchains", "@rango-dev/chains-caip"

llawliet-l-l: You're right. I meant the data we use in this package is from hub3. Fixed in https://github.com/rango-exchange/rango-client/pull/1471/commits/d0e5f61e7a7bed0a1b060771588c58405de05b02

RyukTheCoder (approved):

PR #1472, opened 2026-08-22
refactor(wallets): move widget-only helpers out of wallets-shared

llawliet-l-l:
## Summary

Audited every surviving `wallets-shared` export against its importers in widget packages.  Each moves to the lowest package that actually needs it.

## Changes

- `detectInstallLink` (with its private `isBrave`) to `@rango-dev/ui` — used by widget/ui and widget/embedded, and embedded depends on ui
- `detectMobileScreens` to `@rango-dev/common-core` — a plain user-agent test with no wallet or Rango vocabulary
- The derivation path table to `@rango-dev/wallets-blockchains` as `DERIVATION_PATHS_BY_NAMESPACE`, keeping only path data; nothing read `mainBlockchain` or `title`, so those and the six namespaces carrying nothing else are gone
- `InstallObjects` to `@rango-dev/ui`, and `NamespaceData` replaced by a widget/embedded type built on `Namespace` from `@hub3js/namespaces`

## Notes

`InstallObjects` could not be replaced by hub3's `extensions`: that type is lower-cased, fully optional and carries `homepage` rather than a required `DEFAULT`, while `detectInstallLink` reads the upper-cased keys `useHubAdapter` produces. Swapping it would have compiled and returned the wrong link.

Everything else widget imports — `WalletType`, `WalletTypes`, `Network`, `WalletState`, `WalletInfo` — has provider or queue-manager consumers too, will be fixed in the next tasks.

## How tested

`yarn build:all` 62/62 and root `vitest run` (17 files, 152 tests) pass. widget/embedded's public export list is byte-identical.

# Checklist:

### DIFF
--- a/widget/embedded/src/components/WalletStatefulConnect/DerivationPath.tsx
+++ b/widget/embedded/src/components/WalletStatefulConnect/DerivationPath.tsx
@@ -1,9 +1,9 @@
 import type { PropTypes } from './DerivationPath.types';
-import type { DerivationPath } from '@rango-dev/wallets-shared';
 
 import { i18n } from '@lingui/core';
 import { Divider, Image, MessageBox, Select, TextField } from '@rango-dev/ui';
-import React, { useEffect, useState } from 'react';
+import { useWallets } from '@rango-dev/wallets-react';
+import React, { useEffect, useMemo, useState } from 'react';
 
 import {
   CUSTOM_DERIVATION_PATH,
@@ -26,6 +26,8 @@ export function DerivationPath(props: PropTypes) {
     providerType: type,
   } = props.value;
 
+  const { getWalletInfo } = useWallets();
+
   const [selectedDerivationPathId, setSelectedDerivationPathId] = useState<
     string | null
   >(null);
@@ -36,7 +38,14 @@ export function DerivationPath(props: PropTypes) {
   const isCustomOptionSelected =
     selectedDerivationPathId === CUSTOM_DERIVATION_PATH.id;
 
-  const derivationPaths = getDerivationPaths(selectedNamespace);
+  const derivationPaths = useMemo(
+    () =>
+      getDerivationPaths(
+        getWalletInfo(type).needsDerivationPath?.data ?? [],
+        selectedNamespace
+      ),
+    [type, selectedNamespace]
+  );
 
   const handleDerivationPathItemClick = ({ value }: { value: string }) => {
     const selectedDerivationPath = derivationPaths?.find(
@@ -75,10 +84,8 @@ export function DerivationPath(props: PropTypes) {
   };
 
   useEffect(() => {
-    setSelectedDerivationPathId(
-      getDerivationPaths(selectedNamespace)[0]?.id || null
-    );
-  }, [selectedNamespace]);
+    setSelectedDerivationPathId(derivationPaths[0]?.id || null);
+  }, [derivationPaths]);
 
   return (
     <>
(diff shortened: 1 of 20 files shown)


common/core/src/device.ts
```diff
@@ -0,0 +1,5 @@
+export function detectMobileScreens(): boolean {
```
RyukTheCoder: I prefer moving this to "@rango-dev/widget-embedded" utils. I do not agree with the whole having "@rango-dev/common-core" package

wallets/blockchains/src/derivationPaths.ts
```diff
@@ -0,0 +1,63 @@
+import type { Namespace } from '@hub3js/namespaces';
+
+export type DerivationPath = {
+  id: string;
+  label: string;
+  generateDerivationPath: (index: string) => string;
+};
+
+export const DERIVATION_PATHS_BY_NAMESPACE: Partial<
```
RyukTheCoder: Please try to remove this and get derivation paths from "derivationPath" property of provider in DerivationPath component

RyukTheCoder (changes requested):

common/core/src/device.ts
llawliet-l-l: Fixed in https://github.com/rango-exchange/rango-client/pull/1472/commits/7d45d69d74ed8dde26b99ca2358b590b47017af2

wallets/blockchains/src/derivationPaths.ts
llawliet-l-l: Fixed in https://github.com/rango-exchange/rango-client/pull/1472/commits/540e91618128904bdb39401ad80945979ad8d7ed

widget/ui/src/components/Wallet/ClickableWallet.tsx
```diff
@@ -59,7 +60,8 @@ function Wallet(props: WalletPropTypes) {
         } else {
           onClick(type);
         }
-      }}>
+      }}
```
RyukTheCoder: Make sure your linter is working properly as you added a line here.

RyukTheCoder (changes requested):

widget/embedded/src/components/WalletStatefulConnect/DerivationPath.tsx
```diff
@@ -75,10 +84,8 @@ export function DerivationPath(props: PropTypes) {
   };
 
   useEffect(() => {
-    setSelectedDerivationPathId(
-      getDerivationPaths(selectedNamespace)[0]?.id || null
-    );
-  }, [selectedNamespace]);
+    setSelectedDerivationPathId(derivationPaths[0]?.id || null);
```
RyukTheCoder, line 87: Runs on every render.

RyukTheCoder (changes requested):

widget/ui/src/components/Wallet/ClickableWallet.tsx
llawliet-l-l: fixed

widget/embedded/src/components/WalletStatefulConnect/DerivationPath.tsx
llawliet-l-l: i'm using `useMemo` to calculate `derivationPaths` so it won't change on every render. Could revert it back to only depends on the `selectedNamespace` though

RyukTheCoder (approved):

PR #1474, opened 2026-08-22
refactor(queue-manager): move swap-execution codes out of wallets-shared

llawliet-l-l:
## Summary

The queue manager reached into `@rango-dev/wallets-shared` for swap-execution logic that only lived there because nobody had picked an owner. Each piece now sits with whoever owns it, and neither the queue manager nor its preset imports `wallets-shared` any more. Most of what it pulled turned out to be a re-declaration of types that already exist in `wallets-core/legacy`, so much of this is a repoint rather than a move.

## Changes

- `WalletType`, `Network`, `WalletState`, `Providers` → repointed to `@rango-dev/wallets-core/legacy`, which already declared them (no new definitions; `LegacyProviders` newly exported from its barrel)
- `HYPERLIQUID_SIGN_NETWORK` → `@rango-dev/wallets-blockchains`, next to the `Networks` enum it derives from. It is Rango routing knowledge, not wallet code, and both the queue manager and the widget read it
- `Meta` / `AllBlockchains` → `queue-manager-rango-preset`, the lower of the two packages using them
- `getEvmProvider` **deleted**, not moved: all 35 providers register only the hub version (`1.0.0`), so `providers()` is empty for every real wallet and the legacy branch it guarded was unreachable. The now-unused argument stays as `_` behind a TODO so retiring `SwapQueueContext.providers` is its own change
- `wallets-shared` no longer stands between owners and consumers — its `Networks` re-export and duplicate `Network = string` alias are gone, so the widget now reads both from the same place the queue manager does

## How tested

`yarn build:all` (62 projects) and `vitest` (152 tests) pass.

# Checklist:


queue-manager/rango-preset/src/helpers.ts
```diff
@@ -679,33 +676,12 @@ export function getRequiredWallet(swap: PendingSwap): {
...
-      (await provider.request({ method: 'eth_chainId' })) || provider?.chainId;
-    return chainId;
-  } catch {
-    return provider?.chainId;
-  }
-}
-
 export async function getProviderChainId(
-  providers: Providers,
+  // TODO: unused since the legacy provider lookup was dropped — remove this parameter.
```
RyukTheCoder: You can remove it.

RyukTheCoder (changes requested):

RyukTheCoder (approved):

PR #1476, opened 2026-08-23
ci: use rangutopia for publishing

llawliet-l-l:
Replaces the repo's homegrown publish pipeline (`scripts/publish/**`, ~1,900 lines) with the [rangutopia](https://github.com/rango-exchange/rangutopia) CLI, which is a productized port of it — same channel model (`prod`/`next`/`experimental` from the branch), same `chore(release): publish` commit detection, same `name@version` tag scheme, so git history stays compatible.

### The new flow (`publish.yml`)

1. **`publish:version`** — `rangutopia library version --deferred`: detects changed public packages, computes next versions from conventional commits, saves them to a state file. Emits a `count` step output.
2. **`publish:version:check`** — refuses to continue if any version is already on npm, tagged, or GitHub-released.
3. **`release-root`** (new composite action) — bumps the repo root + `widget/app` + `widget/playground` via `rangutopia client version`, regenerates the root `CHANGELOG.md`, and commits **without pushing**. Gated on `count != '0'`, so a private-package-only change no longer triggers a root bump (parity with the old script's early exit). Fail-open: an older rangutopia without the output still runs the step.
4. **`publish`** — `rangutopia library publish`: builds, applies each version, publishes to npm, writes changelogs, then commits/tags/pushes everything — including the unpushed release-root commit, so a failed publish leaves nothing on the remote. Channel is restored from the saved state; no flag needed.

### Also in this PR

- `permissions: contents: write` on the publish job **and** its caller in `release-prod.yml` (effective reusable-workflow permissions are the intersection of both) — fixes the silent `gh release create` 403s.
- commitlint now ignores `chore(release):` commits (their package-list bodies exceed `body-max-line-length`; the CLI commits with hooks enabled).
- All `actions/*` bumped v4 → v6; `detect-publish-flag` composite action maps branch → channel flag.
- Dead code removed: `scripts/publish/**` and the `scripts/common` modules only it used (graph, repository, features, version). `docs/release.md` rewritten for the new flow.

# Checklist:


arlert-armin (approved):
LGTM

PR #1478, opened 2026-08-24
fix: use wallet type instead of wallet name for the wallet detected event

Ikari-Shinji-re:
# Summary

We were passing the wallet type in the `walletName` field for widget events. However, the `walletDetected` event was not following this convention and was using the actual wallet name, which caused inconsistency. I updated it so that it now also passes the wallet type.

Fixes # (issue)

- [ ] Test A
- [ ] Test B

# Checklist:


nikaaru (approved):
LGTM

PR #1480, opened 2026-08-26
fix(signers): remove extra ton transaction fields

llawliet-l-l:
## Summary

The three TON signers each spread the `tx` object into the payload they sent for signing, so unrelated fields leaked through and each one built a slightly different shape. Adds a static `DefaultTonSigner.buildTx` — mirroring `DefaultEvmSigner.buildTx` — that explicitly extracts `validUntil`, `network`, `from` and `messages` and returns the standard TON Connect bridge payload, and routes all three signers through it.

# Checklist:

### DIFF
--- a/wallets/provider-tonconnect/src/signers/ton.ts
+++ b/wallets/provider-tonconnect/src/signers/ton.ts
@@ -1,9 +1,15 @@
 import type { TonConnectUI } from '@tonconnect/ui';
 import type { GenericSigner, TonTransaction } from 'rango-types';
 
-import { Address, Cell } from '@ton/core';
+import { DefaultTonSigner } from '@rango-dev/signer-ton';
+import { Cell } from '@ton/core';
 import { CHAIN } from '@tonconnect/ui';
-import { SignerError } from 'rango-types';
+import { SignerError, TonChainID } from 'rango-types';
+
+const NETWORKS: Record<TonChainID, CHAIN> = {
+  [TonChainID.MAINNET]: CHAIN.MAINNET,
+  [TonChainID.TESTNET]: CHAIN.TESTNET,
+};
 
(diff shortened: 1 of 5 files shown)


wallets/provider-tonconnect/src/signers/ton.ts
```diff
@@ -17,16 +23,13 @@ export class CustomTonSigner implements GenericSigner<TonTransaction> {
...
-      messages: messages.map(({ stateInit, payload, ...msg }) => ({
-        ...msg,
-        ...(stateInit != null && { stateInit }),
-        ...(payload != null && { payload }),
-      })),
-      network: CHAIN.MAINNET,
+      validUntil: transaction.valid_until,
+      network: NETWORKS[transaction.network],
+      ...(transaction.from && { from: transaction.from }),
+      messages: transaction.messages,
```
arlert-armin: If the difference is only on network, why can't we just spread the whole transaction and just replace the network?

llawliet-l-l: The difference is in `valid_until` field too. If we spread it we'd have a unvalid `valid_until` field in the transaction

arlert-armin: just destructure it const  {valid_unil,..restTransactionFields} = transaction; then spread the restTransaction and use new key for valid_until this way we will make sure that we are covering all sent values.

llawliet-l-l: Fixed in https://github.com/rango-exchange/rango-client/pull/1480/commits/2d88247892d8c9d374caa67144676c901da660b8

arlert-armin (approved):
LGTM

PR #1496, opened 2026-09-16
fix: ship dist and changelog when publishing with npm

llawliet-l-l:
## Summary

Publishing moved from `yarn publish` to `npm publish`, and npm packs workspaces differently:

- It applies the repo root `.gitignore` (`dist/`), so `@rango-dev/provider-ready@0.63.0` was published **without `dist`**, because it had no `files` field.
- It doesn't always include `CHANGELOG.md` the way Yarn v1 does, so every other package stopped shipping its changelog.

This PR sets `"files": ["dist", "src", "CHANGELOG.md"]` on every public package.

## Verification

Packed all 56 public packages with `yarn pack` and with `npm pack --dry-run -w <pkg>`: the file lists now match for every package. The only change from the old Yarn output is that provider-ready no longer ships its two `tsconfig` files, like the other packages.

# Checklist:


arlert-armin (approved):
LGTM

PR #1502, opened 2026-09-21
fix: stop walletconnect from logging uncaught errors on connect

RyukTheCoder:
# Summary

Connecting through WalletConnect logs this error in the console, even though the connection succeeds:

```
Uncaught (in promise) Error: No matching key. session topic doesn't exist: <topic>
    at isValidSessionTopic
    at isValidEmit
    at onSessionEventRequest
    at processRequest
    ...
```

**Cause.** The bug is in `@walletconnect/sign-client`, not in our code. Wallets often emit `chainChanged` or `accountsChanged` right after approving. That event can be processed before `wc_sessionSettle` has stored the session, so validation fails. `onSessionEventRequest` calls the async `isValidEmit` without `await`, so its `try/catch` never sees the failure and the browser reports it as unhandled. The event is emitted anyway, and our subscribers ignore it because the session isn't cached until connect resolves. So the error is only noise. The call is still unawaited in the latest release (2.25.0). The same stack trace was reported upstream in [WalletConnect/walletconnect-monorepo#4240](https://github.com/WalletConnect/walletconnect-monorepo/issues/4240) and closed as not reproducible.

**Fix.** `createUniversalProvider` now wraps the client's `isValidEmit`. The wrapper attaches a handler to the promise the SDK returns, logs the failure through `debug`, and returns that same promise, so SDK code that awaits it still gets the error. Nothing else changes.

**Notes for review**
- **Why not `patch-package`:** host apps such as app-v2 install the SDK themselves and never get this repo's patches, so a runtime fix has to ship in our `dist`.
- **Private SDK API:** `isValidEmit` isn't public, so the wrapper is guarded with `typeof`. If it disappears, the wrapper does nothing and the old log comes back. `@walletconnect/*` is pinned to `2.23.7`.
- **`eslint-disable` for `promise-function-async`:** an `async` wrapper would return a new promise that the SDK doesn't await, bringing back the same unhandled failure.

**Follow-up:** report the missing `await` upstream, and remove the wrapper once a fixed release is out and we've bumped to it.

- [x] `yarn ts-check` in `wallets/provider-walletconnect-2` and ESLint on the changed file both pass.
- [x] Ran the real sign-client 2.23.7 engine in Node. I bundled `provider-client.ts` with esbuild and stubbed two imports: `UniversalProvider.init` returns a `SignClient` that is constructed but not initialized, so no relay is involved, and `logging-core` records `debug` calls. I then called `client.engine.onSessionEventRequest(<unknown topic>, chainChanged)` before and after `createUniversalProvider`:

  | | Unhandled rejections | `session_event` emitted | `debug` called |
  |---|---|---|---|
  | Before | 1 (same "No matching key" message) | yes | no |
  | After | 0 new | yes | yes |

  Awaiting `client.engine.isValidEmit(...)` directly still rejects.
- [ ] Manual: connect WalletConnect in the widget with a mobile wallet and approve. There's no `Uncaught (in promise)` in the console, and connecting, switching chains and disconnecting still work.

# Checklist:

- [ ] I have made corresponding changes to the documentation (not needed)
- [ ] I have added tests that prove my fix is effective or that my feature works (this package has no test setup; tested with a Node script as described above)
- [ ] Implemented a user interface (UI) change, referencing our Figma design to ensure pixel-perfect precision. (not applicable)


wallets/provider-walletconnect-2/src/adapter/provider-client.ts
```diff
@@ -1,13 +1,57 @@
...
+ * This has to live in our code rather than in a `patch-package` patch: host apps
+ * install the SDK themselves and never get this repo's patches.
+ *
+ * The wrapper only attaches a handler and returns the original promise, so the
+ * SDK's awaiting caller (`emit()`) still receives the rejection. `isValidEmit` is
+ * private SDK API, hence the guard. Still unawaited as of sign-client 2.25.0;
+ * drop this once upstream awaits it.
+ */
+function handleSessionEventValidation(client: UniversalProvider['client']) {
+  const engine = client.engine as unknown as { isValidEmit?: EmitValidator };
```
arlert-armin, line 43: Can we avoid the `as unknown as` here? `@walletconnect/types` (already a devDependency) exports `EnginePrivate`, which declares `isValidEmit(params: EngineTypes.EmitParams): Promise<void>`. A type guard would give us the real signature without an assertion, and `EmitValidator` could go:

```ts
import type { EnginePrivate, IEngine } from '@walletconnect/types';

type EmitValidatingEngine = IEngine & Pick<EnginePrivate, 'isValidEmit'>;

function hasEmitValidator(engine: IEngine): engine is EmitValidatingEngine {
  return 'isValidEmit' in engine && typeof engine.isValidEmit === 'function';
}
```

wallets/provider-walletconnect-2/src/adapter/provider-client.ts
```diff
@@ -1,13 +1,57 @@
...
+  const isValidEmit = engine.isValidEmit;
+  if (typeof isValidEmit !== 'function') {
+    return;
+  }
```
arlert-armin, line 47: Keeping the original in a variable makes sense (the wrapper needs it once `engine.isValidEmit` is replaced), but naming it `isValidEmit` makes it read like the same thing as `engine.isValidEmit` below. Could you rename it to `originalIsValidEmit` so it's clear which is the original and which is the wrapper?

wallets/provider-walletconnect-2/src/adapter/provider-client.ts
```diff
@@ -1,13 +1,57 @@
...
+
+  // eslint-disable-next-line @typescript-eslint/promise-function-async -- must return the SDK's own promise; an async wrapper would re-wrap it and leave that promise unhandled
+  engine.isValidEmit = (params) => {
```
arlert-armin, line 50: I first wondered why this isn't just an `async` function with `try/catch`, so the reason could be spelled out more. As I understand it, `onSessionEventRequest` doesn't await `isValidEmit`, so any new promise that rejects (an `async` wrapper that rethrows, or a `new Promise` that rejects) would be unhandled again. Swallowing the error instead would break `emit()`, which awaits it to reject invalid events. Returning the SDK's own promise with a handler attached is the only way to keep both. Could you put that in the comment here?

wallets/provider-walletconnect-2/src/adapter/provider-client.ts
```diff
@@ -1,13 +1,57 @@
...
+    const validation = isValidEmit.call(engine, params);
```
arlert-armin: Couldn't you await it with a try and catch?

wallets/provider-walletconnect-2/src/adapter/provider-client.ts
arlert-armin: Why in here you are using engine.isValidEmit and in the body isValidEmit? aren't they all the same?

wallets/provider-walletconnect-2/src/adapter/provider-client.ts
arlert-armin, line 51: `isValidEmit` is defined as an arrow class field in sign-client, so `.call(engine, ...)` doesn't change anything today. How about binding once when you save it, e.g. `const originalIsValidEmit = engine.isValidEmit.bind(engine);`, and calling `originalIsValidEmit(params)` here? That's clearer and still safe if the SDK ever turns it into a prototype method.

RyukTheCoder:
As this change does not fix the problem, I close it.

PR #1504, opened 2026-09-21
fix: isolate walletconnect client from the host app's clients

RyukTheCoder:
# Summary

In app-v2, connecting or signing through WalletConnect logs this error, even though the flow succeeds:

```
Uncaught (in promise) Error: No matching key. session topic doesn't exist: <topic>
    at isValidSessionTopic
    at isValidEmit
    at onSessionEventRequest
    ...
```

**Cause.** app-v2 also runs Privy, which creates its own WalletConnect client (`@walletconnect/ethereum-provider` 2.21.7). WalletConnect keeps one Core per storage prefix on `globalThis`, and every client created with that prefix reuses it. So our client and Privy's shared one relay connection and one storage namespace, and every client handles every incoming message.

When the wallet sent `chainChanged` (after approval, and after our chain switch before signing), Privy's client received it too. It didn't know our session topic, so the check failed. sign-client doesn't `await` that check, so the failure surfaced as an unhandled rejection.

Evidence: the error only appears in the app that includes Privy, and `window._walletConnectCore__count` was `2` there.

Sharing a Core also had two quieter effects:
- Each client saves its whole session list under the same storage key, overwriting the other's.
- `purgeOrphanedPairings` sees the host's pairings in the shared pairing store and disconnects them.

**Fix.** `createUniversalProvider` now passes `customStoragePrefix: 'rango'` (`WC_STORAGE_PREFIX`). Our client gets its own Core, relay connection and storage keys, so it no longer shares anything with WalletConnect clients the host app runs.

**User impact.** Sessions saved under the old keys aren't restored, so users connected through WalletConnect have to connect again once after updating. The old entries stay in storage on purpose: the default keys may belong to the host app's own client.

**Alternatives considered**
- **Wrapping the SDK's unawaited check:** it only runs on our client, but the error came from Privy's client.
- **`patch-package`:** host apps install the SDK themselves and never get this repo's patches.

- [x] Type-check, ESLint on the changed files, and `yarn build` in `wallets/provider-walletconnect-2` all pass.
- [x] Checked against the real sign-client 2.23.7 in Node. I created a default-prefix client standing in for Privy's, then ours through `createUniversalProvider`, with `UniversalProvider.init` stubbed to build a `SignClient` from the options we pass (no relay connection). Ours gets a separate Core, relayer and pairing store, and its session key is `wc@2:client:0.3:rango//session` instead of `wc@2:client:0.3//session`.
- [ ] Manual, in app-v2 with Privy, using this package through yalc:
  1. Connect with WalletConnect, approve, and sign a transaction.
  2. There's no `No matching key` error in the console.
  3. `window._walletConnectCore__count` is `1` (Privy only) and `window._walletConnectCore_rango_count` is `1` (ours).

# Checklist:

- [ ] I have made corresponding changes to the documentation (not needed)
- [ ] I have added tests that prove my fix is effective or that my feature works (this package has no test setup; checked as described above)
- [ ] Implemented a user interface (UI) change, referencing our Figma design to ensure pixel-perfect precision. (not applicable)


arlert-armin (approved):
LGTM

PR #1507, opened 2026-09-22
fix: add jsonrpc field to btc broadcast requests

RyukTheCoder:
## Summary

Broadcasting signed BTC transactions failed with `Error Code -32603: Internal error` for UniSat, Phantom and Bitget.

GetBlock, our BTC RPC endpoint, now rejects JSON-RPC requests that lack the `jsonrpc` field, whatever the method. Our `sendrawtransaction` request didn't include that field, so the transaction was never broadcast, even though signing had succeeded.

## Changes

- Add `jsonrpc: '2.0'` and `id` to the broadcast request in the UniSat, Phantom and Bitget UTXO signers.
- Add the missing `Content-Type: application/json` header to the Phantom signer.

## Testing

I sent requests to the GetBlock endpoint directly:
- Without `jsonrpc`, every request, including `getblockcount`, returns `-32603 upstream request failed`.
- With `jsonrpc`, `getblockcount` returns the block height. An invalid transaction gets a normal node error (`-22 TX decode failed`), which shows the request is being processed.

A real swap through each wallet hasn't been tested yet.


arlert-armin (approved):
LGTM

